PR #2, opened 2022-10-23
hardcoded proxy works!

futurepaul:


benthecarman:
With this should we make the root directory a cargo workspace so we can call things like `cargo fmt` or `cargo clippy` and it applies to both modules

TonyGiorgio:
I think this is as far as I can get without having a `PeerManager`. There's some things I'm unsure about how to implement until we get one because there's some reimplementation of `lightning-tokio-net` that have to happen. But this does seem to do a good job of sending and receiving data from a websocket with the connection being initiated by clicking on a button in the browser after you create a node.

node-manager/src/node.rs
```diff
@@ -28,13 +29,135 @@ impl Node {
...
+    // TODO then give that connection to the peer manager
+
+    // TODO then schedule a reader on the connection
+
+    Ok(())
+}
+
+pub(crate) fn parse_peer_info(
+    peer_pubkey_and_ip_addr: String,
+) -> Result<(PublicKey, SocketAddr), std::io::Error> {
```
benthecarman, line 99: Is there no helper for this in ldk?

benthecarman: Also the Err type should be MutinyError

node-manager/src/utils.rs
```diff
@@ -8,3 +10,35 @@ pub fn set_panic_hook() {
...
+        Some(bytes) => bytes,
+        None => return None,
+    };
+    match PublicKey::from_slice(&data) {
+        Ok(pk) => Some(pk),
+        Err(_) => None,
+    }
+}
+
+pub fn to_vec(hex: &str) -> Option<Vec<u8>> {
```
benthecarman, line 25: rust-bitcoin has a helper for this I believe

node-manager/src/utils.rs
```diff
@@ -8,3 +10,35 @@ pub fn set_panic_hook() {
     #[cfg(feature = "console_error_panic_hook")]
     console_error_panic_hook::set_once();
 }
+
+pub fn to_compressed_pubkey(hex: &str) -> Option<PublicKey> {
+    let data = match to_vec(&hex[0..33 * 2]) {
+        Some(bytes) => bytes,
+        None => return None,
+    };
+    match PublicKey::from_slice(&data) {
```
benthecarman, line 19: Looks like there is `from_string` that does this

node-manager/src/node.rs
```diff
@@ -28,13 +29,135 @@ impl Node {
...
         })
     }
+
+    pub async fn connect_peer(
+        &self,
+        websocket_proxy_addr: String,
+        peer_pubkey_and_ip_addr: String,
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        if peer_pubkey_and_ip_addr.is_empty() {
+            info!(
```
benthecarman, line 42: this should be an error log

benthecarman (changes requested):
Overall looks good, few comments

many2many/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+    ws: WebSocketUpgrade,
+    user_agent: Option<TypedHeader<headers::UserAgent>>,
+) -> impl IntoResponse {
+    println!("ip: {}, port: {}", ip, port);
+    if let Some(TypedHeader(user_agent)) = user_agent {
+        println!("`{}` connected", user_agent.as_str());
+    }
+
+    // TODO: actually use the ip and port, instead of the hardcoded one
+    // format!("{}:{}", ip.replace("_", "."), port)
```
futurepaul, line 44: sry forgot to delete this comment, it's not todo anymore

node-manager/src/node.rs
```diff
@@ -29,4 +32,129 @@ impl Node {
             keys_manager: Arc::new(keys_manager),
         })
     }
+
+    pub async fn connect_peer(
+        &self,
+        websocket_proxy_addr: String,
+        peer_pubkey_and_ip_addr: String,
+    ) -> Result<(), Box<dyn std::error::Error>> {
```
futurepaul, line 40: this can be a MutinyError or could create a new sub-error like MutinyConnectionError or something for these

the shortcut is to just to return `anyhow::Result<()>` (I like to be explicit that I'm using the anyhow result, because it's a little magic and hides the error type, and will eventually want to swap out with a more concrete error type)

then when you want an error use anyhow's `bail!` macro to return early, or use `?` to return another lib's errors

in the long run it's nice to write all our own errors but also don't want to block on it so anyhow is a nice shortcut when you're not in the mood

futurepaul: if you want to return an error from another lib, but don't want to totally erase it by coercing it into some generic anyhow error, you can do `.with_context(...)`

node-manager/src/tcpproxy.rs
```diff
@@ -15,47 +15,85 @@ use wasm_bindgen_futures::spawn_local;
 
 #[wasm_bindgen]
 pub struct TcpProxy {
-    connections: Vec<WsSplit>,
+    connection: WsSplit,
 }
 
-type WsSplit = Arc<Mutex<(SplitSink<WebSocket, Message>, SplitStream<WebSocket>)>>;
+type WsSplit = Arc<Mutex<SplitSink<WebSocket, Message>>>;
```
futurepaul, line 12: this makes a lot more sense as a non-vec thanks

futurepaul (commented):
won't let me approve this because I helped write it. but I'm a fan!

TonyGiorgio:
> With this should we make the root directory a cargo workspace so we can call things like cargo fmt or cargo clippy and it applies to both modules

Just did this, and updated some of the `just` files and github actions to consider both workspaces.

node-manager/src/node.rs
TonyGiorgio: No, this function comes from LDK-sample in fact, since LDK doesn't have any helpers for it. Unless they added it and have not updated the ldk-sample either. 

Changed it to `MutinyError`

node-manager/src/utils.rs
TonyGiorgio: Thanks, was able to remove all that util logic.

node-manager/src/node.rs
TonyGiorgio: Moved all of these logs out of here and handling at the caller.

node-manager/src/node.rs
TonyGiorgio: Changed to `MutinyError`.

frontend/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "frontend",
   "version": "0.1.0",
   "private": true,
-  "homepage": "https://BitcoinDevShop.github.io/mutiny-web-poc",
+  "homepage": "https://BitcoinDevShop.github.io",
```
benthecarman, line 5: I think this might screw up the github pages deployment, probably fine to merge and see what happens

frontend/src/App.tsx
```diff
@@ -14,13 +14,17 @@ function App() {
...
+  const [newPubkey, setNewPubkey] = useState("")
 
   // Send state
   const [txid, setTxid] = useState("...")
   const [amount, setAmount] = useState("")
   const [destinationAddress, setDestinationAddress] = useState("")
 
+// TODO make these configurable via input box
+  const [proxyAddress, setProxyAddress] = useState("ws://127.0.0.1:3001")
+  const [connectPeer, setConnectPeer] = useState("0307a40e17f451e1030401ef8dd698dac0a4a37bae2e904a337aeeefd39a78b2b9@127.0.0.1:4000")
```
benthecarman, line 26: should we do this TODO?

websocket-tcp-proxy/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+
+#[tokio::main]
+async fn main() {
+    tracing_subscriber::fmt::init();
+
+    let app = Router::new().route("/v1/:ip/:port", get(ws_handler)).layer(
+        TraceLayer::new_for_http().make_span_with(DefaultMakeSpan::default().include_headers(true)),
+    );
+
+    let addr = SocketAddr::from(([127, 0, 0, 1], 3001));
```
benthecarman, line 25: eventually should pick a more random port that is specific to mutiny

websocket-tcp-proxy/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+async fn ws_handler(
+    Path((ip, port)): Path<(String, String)>,
+    ws: WebSocketUpgrade,
+    user_agent: Option<TypedHeader<headers::UserAgent>>,
+) -> impl IntoResponse {
+    println!("ip: {}, port: {}", ip, port);
+    if let Some(TypedHeader(user_agent)) = user_agent {
+        println!("`{}` connected", user_agent.as_str());
```
benthecarman, line 40: you're using printlns here but tracing::info below, should stick to one style

frontend/package.json
TonyGiorgio: I don't think so, it seems to be triggered by that `homepage` url.

TonyGiorgio: Agree just to see what happens

websocket-tcp-proxy/src/main.rs
TonyGiorgio: Fixed

websocket-tcp-proxy/src/main.rs
TonyGiorgio: Made a TODO to let that be configurable

benthecarman (approved):

.github/workflows/test.yml
```diff
@@ -36,10 +36,47 @@ jobs:
...
+      - name: Check formatting
+        working-directory: .
         run: cargo fmt --check
 
-      - name: Run tests
-        working-directory: ./node-manager
-        run: wasm-pack test --headless --firefox --chrome
+      - name: Check clippy
+        working-directory: .
+        run: cargo clippy -- -D warnings
```
benthecarman, line 78: This is always gonna fail for now from unused code because things aren't all hooked up yet, should comment out for now

frontend/src/App.tsx
TonyGiorgio: Added an input box for peer connection string

.github/workflows/test.yml
TonyGiorgio: It hasn't failed so far for unused code, maybe add a flag for your unused code

futurepaul (commented):
let's ship it

PR #3, opened 2022-10-23
Create new node key based on existing node keys

TonyGiorgio:


node-manager/src/nodemanager.rs
```diff
@@ -17,10 +22,23 @@ use crate::{
...
 
+#[derive(Serialize, Deserialize, Clone)]
+pub struct NodeStorage {
+    pub nodes: Vec<NodeIndex>,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct NodeIndex {
+    pub id: String,
+    pub pubkey: String,
```
benthecarman, line 38: can we make this `PublicKey` instead of `String`? Not sure if it has `Serialize, Deserialize`

node-manager/src/nodemanager.rs
```diff
@@ -17,10 +22,23 @@ use crate::{
...
+    pub child_index: i32,
```
benthecarman, line 39: I dont think this will ever be negative so we can use u32 here

node-manager/src/nodemanager.rs
```diff
@@ -98,6 +119,59 @@ impl NodeManager {
         self.wallet.sync().await.expect("Wallet failed to sync")
     }
 
+    #[wasm_bindgen]
+    pub async fn new_node(&self) -> String {
```
benthecarman, line 123: I think this function would make more sense to return `PublicKey` and have a separate function with `#[wasm_bindgen]` that returns calls that and converts it to a string. We will probably end up using this mostly internally so having the `PublicKey` type will be more useful.

node-manager/src/nodemanager.rs
```diff
@@ -98,6 +119,59 @@ impl NodeManager {
...
+        let next_node = NodeIndex {
+            id: Uuid::new_v4().to_string(),
+            pubkey: pubkey.clone(),
+            child_index: next_node_index,
+        };
+        existing_nodes.nodes.push(next_node.clone());
+        MutinyBrowserStorage::insert_nodes(existing_nodes.clone()).expect("could not insert nodes");
+        node_mutex.nodes = existing_nodes.nodes.clone();
+
+        return pubkey.clone();
```
benthecarman, line 172: nit: don't need `return or clone`

node-manager/src/nodemanager.rs
```diff
@@ -17,10 +22,23 @@ use crate::{
...
     mnemonic: Mnemonic,
     wallet: MutinyWallet,
+    node_storage: Mutex<NodeStorage>,
     ws_write: Arc<Mutex<SplitSink<WebSocket, Message>>>,
     counter: usize,
 }
 
+#[derive(Serialize, Deserialize, Clone)]
+pub struct NodeStorage {
+    pub nodes: Vec<NodeIndex>,
```
benthecarman, line 32: would this make more sense as a `Map` so we can quickly access specific nodes?

benthecarman (changes requested):

node-manager/src/nodemanager.rs
```diff
@@ -98,6 +119,59 @@ impl NodeManager {
...
+        let mut node_mutex = self.node_storage.lock().await;
+
+        // Get the current nodes and their bip32 indices
+        // so that we can create another node with the next.
+        // Always get it from our storage, the node_mutex is
+        // mostly for read only and locking.
+        let mut existing_nodes =
+            MutinyBrowserStorage::get_nodes().expect("could not retrieve nodes");
+        let next_node_index = match existing_nodes.nodes.iter().max_by_key(|n| n.child_index) {
+            None => 1,
```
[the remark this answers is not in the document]
TonyGiorgio: something you said about the private key of the parent exposing all children downstream of it, so i just started at one instead of 0 since most wallet will be something like `m/0/.....` right? I can start at zero either way is fine.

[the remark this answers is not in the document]
benthecarman: Ah I see, we should probably using a longer path, at least something like `m/0'/x'` because our wallet utxos are on `86` so it'll eventually have overlap. Normally for HD paths it is 5 items long, not sure if that is required here, but we shouldn't use the the top level key.

node-manager/src/nodemanager.rs
TonyGiorgio: Looks like it does serialize to string correctly, I'll do that

node-manager/src/nodemanager.rs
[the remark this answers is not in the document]
TonyGiorgio: Using `m/0'/x'` now, thanks

node-manager/src/nodemanager.rs
TonyGiorgio: done

node-manager/src/nodemanager.rs
TonyGiorgio: done

node-manager/src/nodemanager.rs
TonyGiorgio: made a function `create_new_node_from_node_manager` for this

node-manager/src/nodemanager.rs
TonyGiorgio: done

node-manager/src/nodemanager.rs
TonyGiorgio: done

TonyGiorgio:
I'll squash in the end

node-manager/src/seedgen.rs
```diff
@@ -1,7 +1,68 @@
...
+        .unwrap()
+        .derive_child(bip32::ChildNumber::new(0, true).unwrap())
+        .unwrap()
+        .derive_child(bip32::ChildNumber::new(child_index, true).unwrap())
+        .unwrap();
+    let current_time = instant::now();
+    let keys_manager = KeysManager::new(
+        &xpriv.to_bytes(),
+        (current_time / 1000.0).round() as u64,
+        (current_time * 1000.0).round() as u32,
```
benthecarman, line 25: this doesn't give us true nano seconds time, just milliseconds with extra 0s, we can just copy what sensei did here

https://github.com/L2-Technology/sensei/blob/3d76006d1a8509ac180886c057e281cc594fb545/senseicore/src/node.rs#L659

benthecarman: seems `instant::now()` is just a layer on top of this anyways

node-manager/Cargo.toml
```diff
@@ -10,12 +10,17 @@ crate-type = ["cdylib"]
...
+bip32 = "0.4.0"
+secp256k1 = "0.24.0"
 bitcoin = "0.29.1"
 bdk = { git = "https://github.com/afilini/bdk", branch = "upgrade/rust-bitcoin-29", default-features = false, features = ["keys-bip39", "esplora", "use-esplora-reqwest"] }
 getrandom = { version = "0.2", features = ["js"] }
 serde = { version = "^1.0", features = ["derive"] }
 serde_json = { version = "^1.0" }
 gloo-storage = "0.2.2"
+uuid = { version = "1.1.2", features = ["v4"] }
+lightning = { version = "0.0.111", features = ["max_level_trace"] }
```
benthecarman, line 22: should we use 112?

benthecarman (commented):
looks good to me otherwise!

node-manager/src/seedgen.rs
TonyGiorgio: can't use `SystemTime::now()` in wasm and [instant](https://docs.rs/instant/0.1.12/instant/index.html) doesn't provide anything but millisecounds.

node-manager/Cargo.toml
TonyGiorgio: ah good catch

benthecarman (approved):

PR #4, opened 2022-10-24
add local testing with console.log

futurepaul:


benthecarman:
I have the ci using Firefox, should maybe should update to use chrome since that's what is we are preferring here

futurepaul:
yeah I think you're using both, but currently our plan is to only support chrome latest and safari latest

benthecarman (approved):

PR #5, opened 2022-10-24
Implement BDK Database types for gloo::LocalStorage

benthecarman:
This has a lot of copy pasting from ldk lite and bdk. I copied ldk lite's Error enum, some of bdk's private helpers for key-value and memory database stuff, and bdk's helpers for testing the database.

One caveat to this is that the database does not actually batch (we just immediately write to LocalStorage, I'm not sure if that will break anything, but should be enough to get started on doing bdk stuff :rocket:


node-manager/Cargo.toml
```diff
@@ -10,8 +10,11 @@ crate-type = ["cdylib"]
 cfg-if = "1.0.0"
 wasm-bindgen = "0.2.83"
 bip39 = { version = "1.0.1" }
-bitcoin = "0.29.1"
+bitcoin = "0.28.1"
+bdk = { git = "https://www.nakamoto.codes/BitcoinDevShop/bdk", branch="feat/use-external-esplora-client", default-features = false}
```
benthecarman, line 14: once https://github.com/bitcoindevkit/bdk/pull/770 is merged, we should update to that

node-manager/src/error.rs
```diff
@@ -0,0 +1,86 @@
+use gloo_storage::errors::StorageError;
+use std::fmt;
+
+#[derive(Debug)]
```
futurepaul, line 4: #[allow(dead_code)] to get rid of warnings for unconstructed variants

node-manager/src/localstorage.rs
```diff
@@ -0,0 +1,959 @@
+use bdk::database::{BatchDatabase, BatchOperations, Database, SyncTime};
+use bdk::{KeychainKind, LocalUtxo, TransactionDetails};
+use bitcoin::consensus::deserialize;
+use bitcoin::consensus::encode::serialize;
+use bitcoin::hash_types::Txid;
+use bitcoin::{OutPoint, Script, Transaction};
+use gloo_storage::{LocalStorage, Storage};
+
+use bitcoin::hashes::hex::{FromHex, ToHex};
+use log::{error, info};
```
futurepaul, line 10: you aren't using these

node-manager/src/localstorage.rs
```diff
@@ -0,0 +1,959 @@
...
+use serde::{Deserialize, Serialize};
+use serde_json::{Map, Value};
+
+pub struct MutinyBrowserStorage {}
+
+macro_rules! log {
```
futurepaul, line 16: not using this either... wonder if we could keep it somewhere top level to have for next time tho, useful for debugging (the `log` crate stuff shows up in a context where we've run `wasm_logger::init` and has some weird formatting stuff that makes it annoying outside of a real browser console)

futurepaul (commented):
this is so epic it might be nice to do a call to review a bit / make sure I understand it. but overall it looks super clean.

if you run `cargo clippy` you get some suggestions about unnecessary clones and formats

also some unused imports and dead code

probably a good practice to handle most warnings before merging if possible?

node-manager/src/lib.rs
```diff
@@ -26,3 +28,13 @@ pub async fn main_js() -> Result<(), JsValue> {
     debug!("Main function ends");
     Ok(())
 }
+
+#[cfg(test)]
+mod test {
+    use crate::storage::delete_mnemonic;
```
futurepaul, line 34: not using?

node-manager/src/localstorage.rs
benthecarman: Yeah I can remove for now, I tried moving it to the general test mod but couldn't get it to work that way

node-manager/src/localstorage.rs
```diff
@@ -0,0 +1,952 @@
...
+                Ok(tx)
+            })
+            .collect()
+    }
+
+    fn iter_txs(&self, include_raw: bool) -> Result<Vec<TransactionDetails>, bdk::Error> {
+        let key = MapKey::Transaction(None).as_map_key();
+        self.scan_prefix(key)
+            .into_iter()
+            .map(|tuple| -> Result<_, bdk::Error> {
```
futurepaul, line 318: you can destructure this right inside the `||`:

|(key, value)| ->

no need for the intermediate `tuple`

node-manager/src/localstorage.rs
```diff
@@ -0,0 +1,946 @@
...
+
+    fn iter_txs(&self, include_raw: bool) -> Result<Vec<TransactionDetails>, bdk::Error> {
+        let key = MapKey::Transaction(None).as_map_key();
+        self.scan_prefix(key)
+            .into_iter()
+            .map(|(key, value)| -> Result<_, bdk::Error> {
+                let mut tx_details: TransactionDetails = Deserialize::deserialize(value)?;
+                if include_raw {
+                    // first byte is prefix for the map, need to drop it
+                    let rm_prefix: &str = &key[2..key.len()];
```
futurepaul, line 317: direct array access can panic so you gotta be certain you def have this... if you're comfy with it I am, I just usually get these wrong :)

can use `get` if you want a way to handle failure: https://doc.rust-lang.org/std/primitive.slice.html#method.get

futurepaul (approved):
okay I'm happy with this. I'd like to kick the tires from the application code but pretty stoked to have so much bdk stuff in here already

node-manager/src/localstorage.rs
benthecarman: Okay switched to use that, seems safer now

benthecarman:
Updated bdk version to https://github.com/bitcoindevkit/bdk/pull/770 as it seems it does not work with wasm without it

TonyGiorgio:
Now that we're using nightly, should we add that to the project configurations? https://stackoverflow.com/a/71978441

TonyGiorgio (approved):
Nicely done man, looking good.

notmandatory:
> Updated bdk version to [bitcoindevkit/bdk#770](https://github.com/bitcoindevkit/bdk/pull/770) as it seems it does not work with wasm without it

Yes you will need the next release of bdk (0.24.0) to work easily with Wasm. There is a way to work around it (see bitcoindevkit/bdk#646 for the whole story), but if you can wait a week we'll have the official release out.

notmandatory:
BTW it's very cool you got this working!! 🤩

PR #10, opened 2022-10-26
Encrypt data in LocalStorage

benthecarman:
The encryption code is mostly just copied from here, fixed some clippy stuff and updated the dep.

Did not do the front end code as we'll probably want a new flow for that, for now the password is just an empty string.

### DIFF
--- a/node-manager/src/encrypt.rs
+++ b/node-manager/src/encrypt.rs
@@ -45,7 +45,7 @@ pub fn decrypt(encrypted: &str, password: &str) -> String {
 
 fn derive_key(password: &str, salt: &[u8]) -> Output {
     let params = Params {
-        rounds: 12345,
+        rounds: 2048,
         output_length: 32,
     };
 
--- a/node-manager/src/encrypt.rs
+++ b/node-manager/src/encrypt.rs
@@ -1,0 +1,82 @@
+use aes_gcm::aead::Aead;
+use aes_gcm::{Aes256Gcm, KeyInit, Nonce};
+use pbkdf2::password_hash::Output;
+use pbkdf2::password_hash::{PasswordHasher, Salt, SaltString};
+use pbkdf2::{Params, Pbkdf2};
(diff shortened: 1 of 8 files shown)


TonyGiorgio (approved):
Awesome, looks great. Glad that was able to slip into the setting and getting.

node-manager/src/encrypt.rs
```diff
@@ -0,0 +1,82 @@
...
+    let nonce = Nonce::from_slice(iv);
+    let decrypted = cipher.decrypt(nonce, data).unwrap();
+    String::from_utf8(decrypted).unwrap()
+}
+
+fn derive_key(password: &str, salt: &[u8]) -> Output {
+    let params = Params {
+        rounds: 12345,
+        output_length: 32,
+    };
```
TonyGiorgio, line 50: It would be good to confirm these parameters, because these can't change later. I had to go through something similar before because we didn't have enough rounds. And too many rounds cause memory issues. Need to look at how many to do based on security but also practicality.

benthecarman: Yeah need to be conscious of mobile phones here too

benthecarman: tried this on my phone and definitely noticed a slow down. I think 2048 will be a good number to pick here, won't be _as secure_ but we are using BIP 39 which uses this anyways so we are falling back to that level of security anyway. Tested 2048 on my phone and didn't notice any real slow down

benthecarman: Something we could also add is a version number to our database so we can migrate if needed

TonyGiorgio: Here's an interesting article about how lastpass went about it: https://support.1password.com/pbkdf2/

TonyGiorgio: But yeah this is good to merge and we can think about it later too

PR #11, opened 2022-10-26
Bdk send

futurepaul:


node-manager/src/wallet.rs
```diff
@@ -39,22 +53,63 @@ impl MutinyWallet {
...
+        let send_to = Address::from_str(&destination_address)
+            .map_err(|e| bdk::Error::Generic(e.to_string()))?;
+
+        let (psbt, details) = {
+            let mut builder = wallet.build_tx();
+            builder
+                .add_recipient(send_to.script_pubkey(), amount)
+                .enable_rbf()
+                .do_not_spend_change()
+                .fee_rate(FeeRate::from_sat_per_vb(5.0));
```
benthecarman, line 82: should make fee rate a param here

node-manager/src/nodemanager.rs
```diff
@@ -114,6 +114,16 @@ impl NodeManager {
             .get_total()
     }
 
+    #[wasm_bindgen]
+    pub async fn send_to_address(&self, destination_address: String, amount: u64) -> String {
```
benthecarman, line 118: should make fee an optional param here as well, and if it is `None` we can get the free rate from esplora

node-manager/src/wallet.rs
```diff
@@ -39,22 +53,63 @@ impl MutinyWallet {
...
+
+        let send_to = Address::from_str(&destination_address)
+            .map_err(|e| bdk::Error::Generic(e.to_string()))?;
+
+        let (psbt, details) = {
+            let mut builder = wallet.build_tx();
+            builder
+                .add_recipient(send_to.script_pubkey(), amount)
+                .enable_rbf()
+                .do_not_spend_change()
```
benthecarman, line 81: i dont see the need for this for now

benthecarman (changes requested):

TonyGiorgio (approved):

benthecarman (approved):

PR #20, opened 2022-11-02
anyhow and thiserror

futurepaul:
pretty happy with this. next step will be to start using MutinyError in the browser and start trying to kill our use of `.unwrap()`


node-manager/src/wallet.rs
```diff
@@ -57,22 +59,34 @@ impl MutinyWallet {
...
+        // #[derive(Error, Debug)]
+        // pub enum MutinyBitcoinError {
+        //     #[error("Failed to use browser storage")]
+        //     AddressError {
+        //         #[from]
+        //         source: bitcoin::util::address::Error,
+        //     },
+        // }
+        let send_to =
+            Address::from_str(&destination_address).with_context(|| "Address parse error")?;
```
TonyGiorgio, line 89: what's `with_context`?

TonyGiorgio:
Looks like a strong refactor, nice one. So should we always be using `MutinyError` now for everything? Do you think it'll play well with LDK errors too and any other library we pull in? Does it pass through to the JS wasm side as well?

TonyGiorgio (approved):
I like the bdkstorage refactor, code runs for me. Just had general questions but LGTM.

node-manager/src/localstorage.rs
```diff
@@ -77,11 +71,20 @@ impl MutinyBrowserStorage {
...
-        let res: gloo_storage::Result<String> = self.get(mnemonic_key);
+    pub(crate) fn get_mnemonic(&self) -> anyhow::Result<Mnemonic> {
+        // TODO: here's another way to write this... but the error conversions end up being a pain in the ass
+        //
+        // self.get(mnemonic_key)
+        //     .and_then(|raw_mnemonic| {
+        //         Ok(Mnemonic::from_str(raw_mnemonic)
+        //             .with_context(|| format!("BIP 39 parse error"))?)
+        //     })
+        //     .with_context(|| format!("storage error"))
```
benthecarman, line 82: I think if you use `.map` instead of `.and_then` then you don't need the extra `with_context`

node-manager/src/wallet.rs
futurepaul: here's how they explain it in the readme:

```
Attach context to help the person troubleshooting the error understand where things went wrong. A low-level error like "No such file or directory" can be annoying to debug without more context about what higher level step the application was in the middle of.
```

node-manager/src/localstorage.rs
```diff
@@ -94,934 +97,20 @@ impl MutinyBrowserStorage {
...
+    pub(crate) fn get_nodes() -> Result<NodeStorage, MutinyStorageError> {
         let res: gloo_storage::Result<NodeStorage> = LocalStorage::get(nodes_key);
         match res {
             Ok(k) => Ok(k),
             Err(e) => match e {
                 StorageError::KeyNotFound(_) => Ok(NodeStorage {
                     nodes: HashMap::new(),
                 }),
-                _ => Err(e),
+                _ => Err(e)?,
```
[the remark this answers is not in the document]
futurepaul: yeah pretty sure... it would be a gloo storage and we're trying to get it automagically converted to a MutinyError... could be more explicit with `with_context` or some other move

not always sure why we don't get automatic "from" errors because we do have a few of those implemented

node-manager/src/localstorage.rs
futurepaul: thanks I'll try that next time

PR #21, opened 2022-11-03
Create logger

### LINKED ISSUE #15: ldk logger
Image

benthecarman:
Closes #15

Seems this is needed for most things and is the easiest so getting this out of the way first

### DIFF
--- a/node-manager/src/logging.rs
+++ b/node-manager/src/logging.rs
@@ -1,0 +1,32 @@
+use chrono::Utc;
+use lightning::util::logger::{Level, Logger, Record};
+use log::*;
+
+#[derive(Default, Debug, Eq, PartialEq, Copy, Clone)]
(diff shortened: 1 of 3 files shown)


node-manager/src/logging.rs
```diff
@@ -0,0 +1,32 @@
+use chrono::Utc;
+use lightning::util::logger::{Logger, Record, Level};
+use log::*;
+
+#[derive(Default, Debug, Eq, PartialEq, Copy, Clone)]
+pub struct MutinyLogger {}
```
TonyGiorgio, line 6: was going to point out `pub` here but you got to it before me. Tested it locally, works well in the browser.

TonyGiorgio (approved):

benthecarman:
Fixed formatting

PR #22, opened 2022-11-03
Mutiny LDK storage

### LINKED ISSUE #17: ldk persister
Image

benthecarman:
Closes #17

I _think_ this is all we need for the ldk storage. A lot of the functions are implemented for us for free using a KV store [here](https://github.com/lightningdevkit/rust-lightning/blob/f4f1093edc506314484c8d52a40dd155e692394b/lightning/src/util/persist.rs)

The main question I had was needing to implement a `read_channel_monitor_updates`. However the default `update_persisted_channel` just overwrites the channel everytime, so I am not sure what the use case of each channel update would be if we are able to safely overwrite while running.

### DIFF
--- a/node-manager/src/error.rs
+++ b/node-manager/src/error.rs
@@ -44,6 +44,8 @@ pub enum MutinyError {
         #[from]
         source: MutinyStorageError,
     },
+    #[error("Failed to read data from storage.")]
+    ReadError { source: MutinyStorageError },
     /// A wallet operation failed.
     #[error("Failed to conduct wallet operation.")]
     WalletOperationFailed,
@@ -71,6 +73,12 @@ pub enum MutinyStorageError {
     },
     #[error(transparent)]
     Other(#[from] anyhow::Error),
+}
+
+impl MutinyError {
+    pub fn read_err(e: MutinyStorageError) -> Self {
+        MutinyError::ReadError { source: e }
+    }
 }
 
 impl From<bdk::Error> for MutinyError {
--- a/node-manager/src/ldkstorage.rs
+++ b/node-manager/src/ldkstorage.rs
@@ -3,6 +3,7 @@ use std::io::Cursor;
 use std::ops::Deref;
 use std::sync::Arc;
 
+use crate::error::MutinyError;
 use bitcoin::BlockHash;
 use lightning::chain::channelmonitor::ChannelMonitor;
 use lightning::chain::keysinterface::{KeysInterface, Sign};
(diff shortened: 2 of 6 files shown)


node-manager/src/ldkstorage.rs
```diff
@@ -0,0 +1,135 @@
...
+    node_id: String,
+    storage: MutinyBrowserStorage,
+}
+
+impl MutinyNodePersister {
+    fn get_key(&self, key: &str) -> String {
+        format!("{}_{}", key, self.node_id)
+    }
+
+    fn read_value(&self, _key: &str) -> Result<Vec<u8>, io::Error> {
```
TonyGiorgio, line 28: Should these be mutiny errors now?

TonyGiorgio, line 28: Also why the underscore here `_key` ?

TonyGiorgio (approved):
Awesome, yeah I think that should do it. Thanks for tackling that, it can be very verbose and tedious. 

You're right about the channel monitor updates. Glad you found the base implementations for it. The main reason I had swapped to saving each individual update is for performance, needed it for hidden lightning network and since I had already done the work I went ahead and did it for lnsploit too. Think that was the right decision for them but these nodes shouldn't get too big of a channel manager where saving the big blobs too fast is a problem. I think you need to implement mutiny errors now but not sure, I'll let @futurepaul make that call.

node-manager/src/ldkstorage.rs
benthecarman: I used _key here because we're gonna concatenate the node id to it so and I didn't want people to accidentally use that key when reading if we refractor this eventually

benthecarman: Yeah I should change to that

futurepaul (approved):

node-manager/src/ldkstorage.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub fn persist_network_graph(&self, network_graph: &NetworkGraph) -> io::Result<()> {
+        self.persist("network", network_graph)
+    }
+
+    pub fn read_network_graph(
+        &self,
+        genesis_hash: BlockHash,
+        logger: Arc<MutinyLogger>,
+    ) -> NetworkGraph {
+        let (already_init, kv_value) = match self.read_value("network") {
```
TonyGiorgio, line 42: We could probably get rid of this `already_init` stuff completely TBH. I forgot why I did that before. But instead of returning an empty vec and later deciding when to initialize from scratch or read the key values, maybe we can put it all in this `match self.read_value` scope.

node-manager/src/ldkstorage.rs
benthecarman: I was converting the `MutinyStorageError` to `std::io::Error` so just instead make it a `MutinyError::ReadError`

benthecarman (commented):
responded to review

node-manager/src/ldkstorage.rs
```diff
@@ -0,0 +1,135 @@
...
+            Ok(kv_value) => (!kv_value.is_empty(), kv_value),
+            Err(_) => (false, vec![]),
+        };
+
+        if already_init {
+            let mut readable_kv_value = Cursor::new(kv_value);
```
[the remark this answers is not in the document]
benthecarman: Yeah this is really to just get around being able to implement `Read`

node-manager/src/error.rs
```diff
@@ -73,6 +75,12 @@ pub enum MutinyStorageError {
     Other(#[from] anyhow::Error),
 }
 
+impl MutinyError {
```
futurepaul, line 78: I feel like this impl shouldn't be necessary but I guess this is because there the extra layer of indirection? io::Error -> MutinyStorageError -> MutinyError

futurepaul (approved):

PR #23, opened 2022-11-04
Add basic BroadcasterInterface

benthecarman:
Didn't end up needing to do the fancy queue & other stuff. Since this just returns `()` we can just ignore the return of the `broadcast` function and not need to deal with the async. The only problem here is that if it fails to broadcast we won't know, this should be fine for POC but left a `fixme` comment for future reference.

### DIFF
--- a/node-manager/src/wallet.rs
+++ b/node-manager/src/wallet.rs
@@ -1,7 +1,8 @@
 use anyhow::Context;
 use futures::lock::Mutex;
-use log::debug;
+use log::{debug, error};
 use std::str::FromStr;
+use std::sync::Arc;
 
 use bdk::blockchain::{Blockchain, EsploraBlockchain};
 use bdk::keys::ExtendedKey;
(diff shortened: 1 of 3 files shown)


node-manager/src/wallet.rs
```diff
@@ -133,6 +134,14 @@ impl MutinyWallet {
     }
 }
 
+#[allow(unused_must_use)]
```
benthecarman, line 137: a warning was being throw from just dropping the return of the function so I had to add this

TonyGiorgio (approved):
Not sure if this will work when it runs in the browser, not sure how await or block will behave or if it needs to await or get sent to a worker thread, but hey it compiles and allows us to use it. Happy to see if it works alter.

node-manager/src/wallet.rs
```diff
@@ -133,6 +136,19 @@ impl MutinyWallet {
...
 
+impl BroadcasterInterface for MutinyWallet {
+    fn broadcast_transaction(&self, tx: &Transaction) {
+        let blockchain = self.blockchain.clone();
+        let tx_clone = tx.clone();
+        spawn_local(async move {
+            blockchain
+                .broadcast(&tx_clone)
+                .await
+                .unwrap_or_else(|_| panic!("failed to broadcast tx! {}", tx_clone.txid()))
```
TonyGiorgio, line 147: no panic

TonyGiorgio (dismissed):

futurepaul (approved):
heck yeah

node-manager/src/wallet.rs
TonyGiorgio: thank you for addressing this to paul's satisfaction

PR #28, opened 2022-11-06
Check if wasm is supported in the browser

TonyGiorgio:


benthecarman (approved):
lgtm, but we should probably have a separate screen for this eventually

TonyGiorgio:
@futurepaul did you say that branches now get their own deployment now? I would like to test this out on my phone to see if it warns me that wasm is disabled.

frontend/src/routes/KitchenSink.tsx
```diff
@@ -3,6 +3,19 @@ import logo from '../images/mutiny-logo.svg';
 import init, { NodeManager, InitOutput } from "node-manager";
 
 function App() {
+    // https://stackoverflow.com/questions/47879864/how-can-i-check-if-a-browser-supports-webassembly
+    const wasmSupported = (() => {
+        try {
```
futurepaul, line 8: this logic should happen inside a useEffect hook to make sure it doesn't get re-run on every re-paint (welcome to frontend!)
https://beta.reactjs.org/apis/react/useEffect

futurepaul (changes requested):

frontend/src/routes/KitchenSink.tsx
TonyGiorgio: I've noticed that node manager runs twice too. Does everything need to be in `useEffect` ?

TonyGiorgio: Moved it

TonyGiorgio:
Doesn't seem to work, the warning doesnt appear on my phone.

frontend/src/routes/App.tsx
```diff
@@ -14,14 +17,43 @@ function App() {
...
+    checkWasm();
+
+  }, [])
+
   return (
     <div className="flex flex-col h-screen">
       <header className='p-8'>
         <img src={logo} className="App-logo" alt="logo" />
         <h2>You're probably looking for <a href="/tests">the tests</a></h2>
+        {wasmSupported ? <p>WASM works!</p> :
```
TonyGiorgio, line 47: We probably don't need anything to appear if it works.

futurepaul: yeah but it's fun for now

PR #29, opened 2022-11-06
Small code clean ups

benthecarman:


node-manager/src/keymanager.rs
```diff
@@ -69,7 +69,7 @@ mod tests {
             second_pubkey.to_string()
         );
 
-        let km = create_keys_manager(mnemonic.clone(), 2);
```
TonyGiorgio, line 72: I wish something would tell me that I didn't need to `clone()`. Sometimes those linger around for me.

TonyGiorgio (approved):

node-manager/src/keymanager.rs
futurepaul: cargo clippy sometimes catches these

futurepaul (approved):
nice, love getting rid of an expect

PR #30, opened 2022-11-07
Add basic socket descriptor with TcpProxy

benthecarman:
I am not using things like `resume_read` but that is probably fine for the POC and possibly the websocket will handle that for us.


TonyGiorgio (approved):
Awesome, LGTM

futurepaul (approved):
nice, this is very exciting. works on my machine and makes ldk feel tantalizingly close!

PR #35, opened 2022-11-07
Websocket docker

TonyGiorgio:
TODO:

- [x] Deploy docker images somewhere

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,3 +123,9 @@ Run the docker image locally
 ```
 docker run -d -p 3001:3001 bitcoindevshop/websocket-tcp-proxy
 ```
+
+Deploy the docker image:
+```
+docker tag bitcoindevshop/websocket-tcp-proxy registry.digitalocean.com/bitcoindevshop-do/websocket-tcp-proxy
+docker push registry.digitalocean.com/bitcoindevshop-do/websocket-tcp-proxy
+```
--- a/README.md
+++ b/README.md
@@ -110,3 +110,16 @@ serve -s build
 ```
 
 They recommend running this in an incognito window because caching can be annoying with this stuff. Works for me in Chrome to install Mutiny as a desktop app.
+
+### Docker
+
+Build the websocket-tcp-proxy image
+
+```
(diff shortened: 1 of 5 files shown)


websocket-tcp-proxy/src/main.rs
```diff
@@ -16,19 +17,24 @@ use tower_http::trace::{DefaultMakeSpan, TraceLayer};
...
+    println!("Running websocket-tcp-proxy");
     tracing_subscriber::fmt::init();
 
     let app = Router::new().route("/v1/:ip/:port", get(ws_handler)).layer(
         TraceLayer::new_for_http().make_span_with(DefaultMakeSpan::default().include_headers(true)),
     );
 
-    // TODO let this be configurable
-    let addr = SocketAddr::from(([127, 0, 0, 1], 3001));
+    let port = match env::var("PORT") {
```
benthecarman, line 27: should make this something like `MUTINY_PROXY_PORT` so it doesn't accidentally overlap with something else

TonyGiorgio: good idea

README.md
```diff
@@ -110,3 +110,16 @@ serve -s build
...
+
+Build the websocket-tcp-proxy image
+
+```
+DOCKER_BUILDKIT=1 docker build -f Dockerfile-Proxy -t bitcoindevshop/websocket-tcp-proxy .
+```
+
+Run the docker image locally
+```
+docker run -d -p 8080:8080 bitcoindevshop/websocket-tcp-proxy
```
benthecarman, line 124: instead of 8080 should use 3001 as that is the default port

TonyGiorgio:
The proxy is at: `wss://websocket-tcp-proxy-fywbx.ondigitalocean.app`, hopefully soon is at `wss://proxy.mutinywallet.com`

You can put that in to the proxy box. If you want to expose a local port so you can connect to, you can run [bore](https://github.com/ekzhang/bore): `bore local {your_local_port} --to bore.pub` and note the port you get.

Then that connection string would be something like `{pubkey}@159.223.171.199:{bore_port}` - just replace the port with the one you got from bore

benthecarman (approved):

futurepaul (approved):
works!

PR #36, opened 2022-11-08
Add chain and channel monitors

TonyGiorgio:


node-manager/src/nodemanager.rs
```diff
@@ -83,13 +88,15 @@ impl NodeManager {
...
-        let wallet = MutinyWallet::new(mnemonic.clone(), storage, Network::Testnet);
+        let wallet = MutinyWallet::new(mnemonic.clone(), storage.clone(), network);
 
         let node_storage = MutinyBrowserStorage::get_nodes().expect("could not retrieve node keys");
 
         NodeManager {
             mnemonic,
+            network,
             wallet,
+            storage: storage.clone(),
```
benthecarman, line 99: this .clone seems redundant

node-manager/src/wallet.rs
```diff
@@ -178,3 +175,17 @@ fn get_tr_descriptors_for_extended_key(
...
+        Network::Bitcoin => Ok("https://blockstream.info/api"),
+        Network::Testnet => Ok("https://blockstream.info/testnet/api"),
+        Network::Signet => Ok("https://mempool.space/signet/api"),
+        Network::Regtest => Err(bdk::Error::Generic(
+            "No esplora client available for regtest".to_string(),
+        )),
+    }
+    .expect("What did I tell you about regtest?");
+    let blockchain = EsploraBlockchain::new(url, 20);
+    blockchain
```
benthecarman, line 190:
```suggestion
    EsploraBlockchain::new(url, 20)
```

benthecarman:
pr title says channel monitor too but i dont see one

node-manager/src/node.rs
```diff
@@ -30,23 +36,73 @@ pub(crate) type PeerManager = LdkPeerManager<
...
+        let chain_monitor: Arc<ChainMonitor> = Arc::new(chainmonitor::ChainMonitor::new(
+            None,
+            wallet.clone(),
+            Arc::new(MutinyLogger {}),
+            wallet,
+            persister.clone(),
+        ));
+
+        // read channelmonitor state from disk
+        let channelmonitors = persister
```
futurepaul, line 97: do we have any idea of how long this takes? sounds a little "blocking" to me. sure it's fine for now maybe just call it out as something we want potentially async / want this whole method to be async

node-manager/src/nodemanager.rs
```diff
@@ -69,6 +71,9 @@ impl NodeManager {
     pub fn new(password: String, mnemonic: Option<String>) -> NodeManager {
```
futurepaul, line 71: I think it makes sense for this to return an `anyhow::Result<NodeManager>`, don't think it helps much to keep punting on it at this point

then we turn the expects into `?`

then we can do a try / catch in the browser

futurepaul (changes requested):

node-manager/src/node.rs
TonyGiorgio: It's just reading one key value, but yeah we'll end up needing async for this function eventually

node-manager/src/nodemanager.rs
TonyGiorgio: fixed

node-manager/src/node.rs
TonyGiorgio, line 97: @benthecarman this is where channel monitors is initialized, it's read in from the persister.

node-manager/src/nodemanager.rs
TonyGiorgio: why `anyhow` ? Why not the mutiny errors?

TonyGiorgio: We can do it like this: https://rustwasm.github.io/wasm-bindgen/reference/types/result.html?highlight=result#resultt-e

```
impl Into<JsValue> for MutinyError {
    fn into(self) -> JsValue {
        todo!()
    }
}
```

TonyGiorgio: I don't know how to turn all the various errors here into `JsValue` right now. And I don't introduce any additional `expect`'s here so I'll open up a ticket and we can handle it soon outside of this PR.

TonyGiorgio: https://github.com/BitcoinDevShop/mutiny-web-poc/issues/38

TonyGiorgio:
Handled what I could/should here, let me know if it's good to go.

benthecarman (approved):

benthecarman:
needs rebase

PR #37, opened 2022-11-08
Initial peer manager

benthecarman:
Will probably wait until #36 is ready so we can replace the channel handler from a `ErroringMessageHandler` to our own.

### DIFF
--- a/node-manager/src/node.rs
+++ b/node-manager/src/node.rs
@@ -12,14 +12,21 @@ use crate::{
 use anyhow::Context;
 use bip39::Mnemonic;
 use bitcoin::secp256k1::PublicKey;
-use lightning::chain::keysinterface::KeysManager;
+use lightning::chain::keysinterface::{KeysInterface, KeysManager, Recipient};
 use lightning::ln::peer_handler::{
-    ErroringMessageHandler, IgnoringMessageHandler, PeerManager as LdkPeerManager,
+    ErroringMessageHandler, IgnoringMessageHandler, MessageHandler as LdkMessageHandler,
+    PeerManager as LdkPeerManager,
 };
 use lightning::routing::gossip;
 use log::info;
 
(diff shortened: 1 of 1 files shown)


node-manager/src/node.rs
```diff
@@ -33,20 +40,32 @@ pub(crate) type PeerManager = LdkPeerManager<
...
         info!("initialized a new node: {}", node_index.uuid);
 
-        let keys_manager = create_keys_manager(mnemonic, node_index.child_index);
+        let logger = Arc::new(MutinyLogger::default());
+
+        let keys_manager = Arc::new(create_keys_manager(mnemonic, node_index.child_index));
         let pubkey = pubkey_from_keys_manager(&keys_manager);
 
+        let ln_msg_handler = MessageHandler {
+            chan_handler: Arc::new(ErroringMessageHandler::new()),
```
TonyGiorgio, line 57: We're going to end up putting the channel manager here, could you just add a comment mentioning that so we don't forget?

TonyGiorgio (approved):
small comment but otherwise lgtm

node-manager/src/node.rs
```diff
@@ -107,6 +126,25 @@ pub(crate) async fn connect_peer_if_necessary(
...
+    lightning_msg_handler: MessageHandler,
+    logger: Arc<MutinyLogger>,
+) -> PeerManager {
+    let current_time = instant::now();
+    let mut ephemeral_bytes = [0u8; 32];
+    getrandom::getrandom(&mut ephemeral_bytes).expect("Failed to generate entropy");
+
+    PeerManager::new(
+        lightning_msg_handler,
+        km.get_node_secret(Recipient::Node).unwrap(),
```
futurepaul, line 140: I'd prefer an expect here if this is an unrecoverable sort of error (getrandom is a great example). otherwise error handling plz

futurepaul (changes requested):

futurepaul (approved):

PR #42, opened 2022-11-10
Implement Filter & helper functions

benthecarman:
Implements some of the required stuff for getting blockchain data for ldk. A lot is copied from https://github.com/lightningdevkit/ldk-lite/pull/9 and adapted to fit in our use case.

The next step would to get a stream of txs and somehow construct a stream of `lightning::chain::Confirm`s and call `chain.sync` on them.

Fixed a couple small clippy errors as well


TonyGiorgio (approved):
didn't look too much at `chain.rs` but the rest looks good and i like the separation of `MutinyWallet` and `MutinyChain`.

PR #49, opened 2022-11-14
Stub out front end apis

benthecarman:


node-manager/src/nodemanager.rs
```diff
@@ -182,6 +233,44 @@ impl NodeManager {
...
+    ) -> Result<JsValue /* TransactionDetails */, MutinyJsError> {
+        todo!()
+    }
+
+    #[wasm_bindgen]
+    pub async fn list_onchain(&self) -> Result<JsValue, MutinyJsError> {
+        let txs = self.wallet.wallet.lock().await.list_transactions(false)?;
+
+        Ok(serde_wasm_bindgen::to_value(&txs)?)
+    }
```
TonyGiorgio, line 249: Can we return the actual type back instead of `JsValue`? That way the frontend has access to the expected types?

I may have sent you down the wrong path with the `serde_wasm_bindgen` approach, I knew it was a thing but I remember now why @futurepaul and I thought against it with the errors.

One approach is to have something like this instead:

```rust
    #[wasm_bindgen]
    struct MutinyTransactionDetails(Vec<TransactionDetails>);


    #[wasm_bindgen]
    pub async fn list_onchain(&self) -> Result<MutinyTransactionDetails, MutinyJsError> {
        match self.wallet.wallet.lock().await.list_transactions(false) {
            Ok(txs) => Ok(MutinyTransactionDetails(txs)),
            Err(e) => Err(MutinyError::from(e).into()),
        }
    }
```

I would say implement `JsValule` for `TransactionDetails` but we can't implement traits for structs that we do not own. 

You can also remove this error conversion too with the above approach:

```
impl From<bdk::Error> for MutinyJsError {
```

I don't know why `into` doesn't know to go `into().into()` but `from().into()` works for now. We really just need to implement `Try` I think so we can get `?` back to avoid `match` requirements. Maybe you could `map_error` instead?

TonyGiorgio (changes requested):
Since we're doing this to conform to type api requirements on both sides, I would like to see if we can return the real objects and not just `JsValue`. We can go back to `todo()` instead of trying to implement them now if we want to punt on it. But I think there's a way to get this in my comment, curious what others think.

node-manager/src/nodemanager.rs
TonyGiorgio: I don't think it's the cleanest what I typed above but if it's just at the wasm conversion layer then that's not the worst thing in the world to have 4 lines of boilerplate per api.

futurepaul: yeah totally agree with the sentiment, not exactly sure the best course of action. returning a generic `JsValue` to the frontend should only be used as a last resort because I lose all typing

TonyGiorgio: ah yeah `struct MutinyTransactionDetails(TransactionDetails);` wouldl probably be better since you can always `vec` it

node-manager/src/error.rs
```diff
@@ -200,6 +203,21 @@ impl From<MutinyError> for MutinyJsError {
     }
 }
 
+impl From<bdk::Error> for MutinyJsError {
+    fn from(e: bdk::Error) -> Self {
```
TonyGiorgio, line 207: I still don't think we need this because it should be turning into `MutinyError` instead of `MutinyJsError` directly.

TonyGiorgio (approved):
LGTM

PR #54, opened 2022-11-16
Allow generating 12 or 24 word seed phrase

benthecarman:
Made it so `generate_seed()` now results a `Result` so we can handle specific errors if something fails there.

For now made it so the app is hardcoded to generates 12 word seeds. I don't know if we want to make this a parameter later but felt this was good to at least get the internals out of the way for now.


futurepaul (approved):
I haven't run this but read through it and everything looked straightforward

TonyGiorgio (approved):

PR #71, opened 2022-11-22
fix double init and add some timers

futurepaul:


node-manager/src/lib.rs
```diff
@@ -35,8 +35,6 @@ cfg_if! {
 #[wasm_bindgen(start)]
 pub async fn main_js() -> Result<(), JsValue> {
     wasm_logger::init(wasm_logger::Config::new(Level::Debug).message_on_new_line());
-    debug!("Main function begins");
-    debug!("Main function ends");
```
TonyGiorgio, line 39: why remove this? sometimes that's the only reason we know that the double init is happening.

TonyGiorgio (approved):

node-manager/src/lib.rs
futurepaul: just feels silly, maybe I'll just make it one line

benthecarman: +1, should at least leave one of the messages

PR #73, opened 2022-11-22
save to and reload from json file

futurepaul:
maybe this is kind of dumb idk but it was stuck in my brain

### DIFF
--- a/frontend/src/routes/Settings.tsx
+++ b/frontend/src/routes/Settings.tsx
@@ -21,7 +21,7 @@ function Settings() {
         let serializable: Record<string, any> = {};
         for (let i = 0; i < localStorage.length; i++) {
             const key = localStorage.key(i)
-            serializable[key!] = JSON.parse(localStorage.getItem(key!)!);
+            serializable[key!] = localStorage.getItem(key!);
         }
         console.log(serializable)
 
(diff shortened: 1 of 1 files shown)


benthecarman:
This is great!

benthecarman (changes requested):
tried copying my storage from firefox to chrome, doesn't seem to work, no error logs either, it did log my entire storage

futurepaul:
hm good idea. I just tested it in place

futurepaul:
okay I think it works now

benthecarman:
> one caveat: it doesn't clear existing storage, so beware of that

We should probably clear storage just to be safe

frontend/src/routes/Settings.tsx
```diff
@@ -17,18 +17,84 @@ function Settings() {
...
+        link.remove()
+    };
+
+    function handleFileChoose(e: React.ChangeEvent) {
+        const fileReader = new FileReader();
+        const target = e.target as HTMLInputElement;
+        const file: File = (target.files as FileList)[0];
+        fileReader.readAsText(file, "UTF-8");
+        fileReader.onload = e => {
+            handleClearState();
```
benthecarman, line 56: lets not clear state until we we absolutely have to, in case something goes wrong. This should happen after the console log I think

benthecarman (approved):

PR #75, opened 2022-11-22
Wire up more wallet

futurepaul:
list channels and peers

kind of wip can do more, but feel free to merge if it's helpful


benthecarman:
For the channels it is going off the edge of my screen and the top spacing is weird
image

benthecarman:
Will look into if we can get the aliases as well

PR #76, opened 2022-11-22
Fixes for channels + other shit

benthecarman:

PR #78, opened 2022-11-22
Correctly convert sats to millisats in invoice creation

benthecarman:
When debugging I broke this in #76

### DIFF
--- a/node-manager/src/event.rs
+++ b/node-manager/src/event.rs
@@ -21,6 +21,7 @@ pub(crate) struct PaymentInfo {
     pub secret: Option<[u8; 32]>,
     pub status: HTLCStatus,
     pub amt_msat: MillisatAmount,
+    pub fee_paid_msat: Option<u64>,
 }
 
 #[derive(Serialize, Deserialize)]
@@ -239,6 +240,7 @@ impl LdkEventHandler for EventHandler {
                             secret: payment_secret,
                             status: HTLCStatus::Succeeded,
                             amt_msat: MillisatAmount(Some(*amount_msat)),
+                            fee_paid_msat: None,
                         };
                         match self
                             .persister
@@ -261,6 +263,7 @@ impl LdkEventHandler for EventHandler {
             Event::PaymentSent {
                 payment_preimage,
                 payment_hash,
+                fee_paid_msat,
                 ..
             } => {
                 self.logger.log(&Record::new(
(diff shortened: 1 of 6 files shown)


benthecarman:
Found some other bugs and threw the fixes in here

node-manager/src/event.rs
```diff
@@ -271,14 +274,15 @@ impl LdkEventHandler for EventHandler {
...
+                match self.persister.read_payment_info(*payment_hash, false) {
                     Some(mut saved_payment_info) => {
                         saved_payment_info.status = HTLCStatus::Succeeded;
                         saved_payment_info.preimage = Some(payment_preimage.0);
+                        saved_payment_info.fee_paid_msat = *fee_paid_msat;
                         match self.persister.persist_payment_info(
                             *payment_hash,
                             saved_payment_info,
-                            true,
+                            false,
```
TonyGiorgio, line 285: i can't believe i screwed some of these up, i thought i triple checked them. nice catch.

node-manager/src/nodemanager.rs
```diff
@@ -502,33 +508,45 @@ impl NodeManager {
...
+            Some(payment_info) => {
+                let mut mutiny_invoice = MutinyInvoice::from(invoice);
+
+                mutiny_invoice.is_send = true;
+                mutiny_invoice.paid = true;
+                mutiny_invoice.fees_paid = payment_info.fee_paid_msat;
+
+                Ok(mutiny_invoice)
+            }
+            None => Err(MutinyJsError::ReadError),
```
TonyGiorgio, line 539: Unfortunately this won't work. the invoice payer is async so it really only indicates that a payment is in progreess.

node-manager/src/nodemanager.rs
```diff
@@ -542,27 +560,34 @@ impl NodeManager {
...
+                    bolt11: "".to_string(),
+                    description: None,
+                    payment_hash: payment_hash.0.to_hex(),
+                    preimage: Some(preimage.0.to_hex()),
+                    amount_sats: Some(amt_sats),
+                    expire: 0,
+                    paid: true,
+                    fees_paid: payment_info.fee_paid_msat,
+                    is_send: true,
+                };
```
TonyGiorgio, line 585: same with this

TonyGiorgio (changes requested):

node-manager/src/nodemanager.rs
TonyGiorgio: In order to have a blocking call, we need a channel between event handler and this that blocks on waiting for the event handler to emit a payment success event.

benthecarman: reverted these changes as they'll be addresses in your PR anyways

PR #79, opened 2022-11-22
Finish the invoice stubs

TonyGiorgio:
Still need to test this but this should finish out the invoices stuff.

Long term, I think there should be a clear distinction between inbound invoices and payments. There's just too much intermingling than I would like to reason with and much of what I did seemed v dirty.

### DIFF
--- a/node-manager/src/ldkstorage.rs
+++ b/node-manager/src/ldkstorage.rs
@@ -27,6 +27,7 @@ use lightning::util::persist::KVStorePersister;
 use lightning::util::ser::{ReadableArgs, Writeable};
 use log::error;
 use serde::Deserialize;
+use serde_json::Value;
 use std::io;
 use std::io::Cursor;
 use std::ops::Deref;
@@ -134,7 +135,8 @@ impl MutinyNodePersister {
 
         // Get all the channel monitor buffers that exist for this node
         let suffix = self.node_id.as_str();
-        let channel_monitor_list = self.storage.scan(MONITORS_PREFIX_KEY, Some(suffix));
+        let channel_monitor_list: Vec<(String, Value)> =
+            self.storage.scan(MONITORS_PREFIX_KEY, Some(suffix));
 
         // TODO probably could use a fold here instead
         for (_, value) in channel_monitor_list {
@@ -224,7 +226,7 @@ impl MutinyNodePersister {
         inbound: bool,
     ) -> io::Result<()> {
         self.storage
-            .set(payment_key(inbound, payment_hash), &payment_info)
+            .set(payment_key(inbound, payment_hash), payment_info)
             .map_err(io::Error::other)
     }
 
@@ -237,6 +239,14 @@ impl MutinyNodePersister {
         let deserialized_value: Result<PaymentInfo, MutinyError> =
             self.storage.get(key).map_err(MutinyError::read_err);
         deserialized_value.ok()
+    }
+
+    pub(crate) fn list_payment_info(&self, inbound: bool) -> Vec<(String, PaymentInfo)> {
+        let prefix = match inbound {
+            true => PAYMENT_INBOUND_PREFIX_KEY,
+            false => PAYMENT_OUTBOUND_PREFIX_KEY,
+        };
+        self.storage.scan(prefix, None)
     }
 }
 
--- a/node-manager/src/ldkstorage.rs
+++ b/node-manager/src/ldkstorage.rs
@@ -26,8 +26,7 @@ use lightning::util::config::UserConfig;
 use lightning::util::persist::KVStorePersister;
 use lightning::util::ser::{ReadableArgs, Writeable};
 use log::error;
-use serde::Deserialize;
-use serde_json::Value;
+use std::collections::HashMap;
 use std::io;
 use std::io::Cursor;
 use std::ops::Deref;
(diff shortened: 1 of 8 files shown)


TonyGiorgio:
there's no doubt lots of clippy warnings so I'll get to those tomorrow.

TonyGiorgio:
rebase wasn't bad at all, going to clean up

node-manager/src/nodemanager.rs
```diff
@@ -73,7 +73,7 @@ impl NodeIdentity {
 #[derive(Serialize, Deserialize, Clone, Eq, PartialEq)]
 #[wasm_bindgen]
 pub struct MutinyInvoice {
-    bolt11: String,
+    bolt11: Option<String>,
```
TonyGiorgio, line 76: Just realized I didn't have to put `Option` here and could have put `""`. 

@benthecarman  @futurepaul  thoughts on returning `Option<String>` vs empty string?

benthecarman: None is more explicit than an empty string so we should use that

node-manager/src/ldkstorage.rs
```diff
@@ -238,6 +240,14 @@ impl MutinyNodePersister {
             self.storage.get(key).map_err(MutinyError::read_err);
         deserialized_value.ok()
     }
+
+    pub(crate) fn list_payment_info(&self, inbound: bool) -> Vec<(String, PaymentInfo)> {
+        let prefix = match inbound {
+            true => PAYMENT_INBOUND_PREFIX_KEY,
+            false => PAYMENT_OUTBOUND_PREFIX_KEY,
+        };
+        self.storage.scan(prefix, None)
```
benthecarman, line 249: matching on a boolean is a little weird but nbd

node-manager/src/nodemanager.rs
```diff
@@ -565,26 +580,52 @@ impl NodeManager {
     }
 
     #[wasm_bindgen]
-    pub async fn decode_invoice(&self, _invoice: String) -> Result<MutinyInvoice, MutinyJsError> {
-        todo!()
+    pub async fn decode_invoice(&self, invoice: String) -> Result<MutinyInvoice, MutinyJsError> {
+        Invoice::from_str(&invoice)
+            .map(|i| i.into())
+            .map_err(|e| e.into())
```
benthecarman, line 586: doing a `map` and `map_err` is equivalent to just doing a match, that would be more clean here

node-manager/src/nodemanager.rs
```diff
@@ -565,26 +580,52 @@ impl NodeManager {
...
     }
 
     #[wasm_bindgen]
-    pub async fn get_invoice(&self, _invoice: String) -> Result<MutinyInvoice, MutinyJsError> {
-        todo!()
+    pub async fn get_invoice(&self, invoice: String) -> Result<MutinyInvoice, MutinyJsError> {
+        let nodes = self.nodes.lock().await;
+        let inv_opt: Option<MutinyInvoice> = nodes
+            .iter()
+            .find_map(|(_, n)| n.get_invoice(invoice.clone()).ok());
```
benthecarman, line 594: We are going to be parsing `invoice` into an `Invoice` n times here. It would be better if `get_invoice` took in an `Invoice` type instead of `String`. Would be more efficient and makes that function more typesafe

node-manager/src/node.rs
```diff
@@ -365,6 +367,114 @@ impl Node {
         Ok(invoice)
     }
 
+    pub fn get_invoice(&self, invoice_str: String) -> Result<MutinyInvoice, MutinyError> {
+        let invoice = Invoice::from_str(&invoice_str).map_err(Into::<MutinyError>::into)?;
```
benthecarman, line 371: Reposting previous comment, but I think this function should take in an `Invoice` instead of `String`, let the parsing be handled on the `wasm` bound functions.

node-manager/src/node.rs
```diff
@@ -365,6 +367,114 @@ impl Node {
...
+                        }
+                        Err(_) => None,
+                    }
+                }
+                None => {
+                    // Constructing MutinyInvoice from no invoice, harder
+                    let paid = matches!(i.status, HTLCStatus::Succeeded);
+                    // TODO finish this properly
+                    let mutiny_invoice =
+                        MutinyInvoice::new(None, None, h, None, None, 0, paid, None, !inbound);
```
benthecarman, line 410: we should have `fees_paid` and `amount_sats` from payment info

node-manager/src/node.rs
```diff
@@ -365,6 +367,114 @@ impl Node {
...
+        match self
+            .persister
+            .read_payment_info(PaymentHash(payment_hash.into_inner()), false)
+        {
+            Some(payment_info) => Ok((payment_info, false)),
+            None => Err(MutinyError::InvoiceInvalid),
+        }
+    }
+
+    // pay_invoice sends off the payment but does not wait for results
```
benthecarman, line 439: This comment should have 3 slashes so it appears as docs for the function

benthecarman (changes requested):

node-manager/src/ldkstorage.rs
TonyGiorgio: keeping it for now

node-manager/src/nodemanager.rs
TonyGiorgio: cleaned up

node-manager/src/nodemanager.rs
TonyGiorgio: fixed, passing it through now

node-manager/src/node.rs
TonyGiorgio: fixed here too

node-manager/src/node.rs
TonyGiorgio: parsing them now, along with the rest of what i can parse

node-manager/src/node.rs
TonyGiorgio: added

node-manager/src/node.rs
```diff
@@ -365,6 +367,114 @@ impl Node {
...
+    pub fn pay_invoice(&self, invoice_str: String) -> Result<MutinyInvoice, MutinyError> {
```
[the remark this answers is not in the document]
TonyGiorgio: done

benthecarman (approved):

PR #86, opened 2022-11-23
Send keysend

futurepaul:


benthecarman:
Worked for me! Would be nice if after sending on-chain it calls `sync()` so our on-chain balance is updated

frontend/src/utility/dumb.ts
```diff
@@ -11,4 +12,33 @@ export async function getFirstNode(nodeManager: NodeManager): Promise<string> {
...
+    "unknown"
+}
+
+export function detectPaymentType(p: string): PaymentType {
+    if (p.startsWith("lnt")) {
+        return PaymentType.invoice
+    } else if (p.startsWith("03") || p.startsWith("02")) {
+        return PaymentType.keysend
+    } else if (p.startsWith("tb1")) {
+        return PaymentType.onchain
```
benthecarman, line 30: the invoice and on-chain ones won't work for mainnet. Also on-chain wont work for older addresses. might be a good idea to bring in a library here, I imagine we'll eventually have, on-chain address, invoice, node pubkey, lnurl, and lightning address, so library could potentially simplify this

frontend/src/routes/SendConfirm.tsx
```diff
@@ -41,6 +41,7 @@ function SendConfirm() {
 
         if (paymentType === PaymentType.onchain) {
           const txid = await nodeManager?.send_to_address(destination, amountInt);
+          await nodeManager?.sync();
           setTxid(txid)
           setSentOnchain(true);
```
benthecarman, line 46: I think it'd be better to sync after we set the UI so the user doesn't have to wait for it

PR #87, opened 2022-11-23
Peer reconnect

TonyGiorgio:


node-manager/src/nodemanager.rs
```diff
@@ -210,12 +214,18 @@ impl NodeManager {
...
         set_panic_hook();
 
-        // TODO get network from frontend
-        let network = Network::Testnet;
-        //let network = Network::Regtest;
+        let websocket_proxy_addr = if let Some(w) = websocket_proxy_addr {
+            w
+        } else {
+            String::from("wss://websocket-tcp-proxy-fywbx.ondigitalocean.app")
+        };
```
benthecarman, line 226: a simpler way to do this would be `websocket_proxy_addr.unwrap_or(...)`

node-manager/src/utils.rs
```diff
@@ -40,3 +40,16 @@ pub fn currency_from_network(network: Network) -> Currency {
...
+pub fn network_from_string(network_str: Option<&str>) -> Network {
+    match network_str {
+        Some("mainnet") => Network::Bitcoin,
+        Some("testnet") => Network::Testnet,
+        Some("regtest") => Network::Regtest,
+        Some("signet") => Network::Signet,
+        Some(net) => {
+            panic!("Unsupported network provided. Options are: `regtest`, `testnet`, and `signet`. Got {}", net);
+        }
+        None => Network::Testnet,
```
benthecarman, line 53: I think `Network` has a utility for this

PR #97, opened 2022-11-23
Pay zero amount invoice

TonyGiorgio:


node-manager/src/node.rs
```diff
@@ -522,8 +522,20 @@ impl Node {
...
+        amt_sats: Option<u64>,
+    ) -> Result<MutinyInvoice, MutinyError> {
+        let pay_result = if invoice.amount_milli_satoshis().is_none() {
+            if amt_sats.is_none() {
+                return Err(MutinyError::InvoiceInvalid);
+            }
+            self.invoice_payer
+                .pay_zero_value_invoice(&invoice, amt_sats.unwrap())
+        } else {
+            self.invoice_payer.pay_invoice(&invoice)
```
benthecarman, line 537: we should probably throw an error if `amt_sats`is some here because the frontend is probably misusing the api

TonyGiorgio: added `is_some()`

benthecarman (approved):

PR #113, opened 2022-11-23
Make OnChainTx.fee optional on front end

benthecarman:
This is optional in the backend so should be the same here.


TonyGiorgio (approved):

futurepaul (approved):
probably a lot of cases like this where I saw data in a certain shape and expect it to be there all the time

PR #114, opened 2022-11-23
Fix TODO use a fold instead of for loop

benthecarman:
This is more inline with functional programming paradigms

### DIFF
--- a/node-manager/src/ldkstorage.rs
+++ b/node-manager/src/ldkstorage.rs
@@ -132,28 +132,29 @@ impl MutinyNodePersister {
     where
         K::Target: KeysInterface<Signer = Signer> + Sized,
     {
-        let mut res = Vec::new();
-
         // Get all the channel monitor buffers that exist for this node
         let suffix = self.node_id.as_str();
         let channel_monitor_list: HashMap<String, Vec<u8>> =
             self.storage.scan(MONITORS_PREFIX_KEY, Some(suffix));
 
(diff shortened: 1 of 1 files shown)


node-manager/src/ldkstorage.rs
```diff
@@ -132,28 +132,29 @@ impl MutinyNodePersister {
...
-                    res.push((blockhash, channel_monitor));
-                }
-                Err(e) => {
-                    return Err(io::Error::new(
-                        io::ErrorKind::InvalidData,
-                        format!("Failed to deserialize ChannelMonitor: {}", e),
-                    ));
+        let res = channel_monitor_list
+            .iter()
+            .fold(Ok(Vec::new()), |current_res, (_, data)| match current_res {
```
TonyGiorgio, line 142: So `fold` starts out by initializing a vec, passing that in as `current_res` via `match` since it starts out `Ok()`. 

Then it goes through the logic, pushes to the `Vec` on success, returns the `Vec`, triggering it's match again and going through the motions until complete. And if any of the iteration fails, it returns an error (by bubbling up the error through the rest of the iterations until it hits the end and returns `Err` through the `?`? 

I think that makes sense just wanted to understand fold.

TonyGiorgio (approved):

node-manager/src/ldkstorage.rs
benthecarman: yes exactly

PR #116, opened 2022-11-23
Optimize peer reconnect logic

benthecarman:
Previously we would have to call `peer_manager.get_peer_node_ids()` everytime we would try to connect to a peer. This has a lock inside the peer manager so we would invoke this lock N times. This isn't optimal so instead I made it so we check once and only connect `connect_peer` on the peers we aren't connected to.


TonyGiorgio (approved):
great change, seems like this also will cut down on spammy logs.

PR #119, opened 2022-11-25
Handle missing port in peer info

### LINKED ISSUE #108: if there's no port default to 9735

benthecarman:
Closes #108


futurepaul (approved):
nice

PR #123, opened 2022-11-26
Bip 21 but more tryhard

futurepaul:
this builds on #112, I got spooked by a big multiply so wanted to get that logic into rust

these nested if statements are getting scary, will probably want to find a way to compress this logic into some more readable form


benthecarman (approved):
> these nested if statements are getting scary, will probably want to find a way to compress this logic into some more readable form

Agreed but lgtm to me for now

PR #124, opened 2022-11-26
use env for bitcoin network

futurepaul:
new setup explained in readme, I already added the env variable to vercel so should "just work" on deploy


benthecarman (commented):
I think it'd be good to add `.env.sample` file so we can fill it out with all available options as we add more things

benthecarman (approved):

PR #127, opened 2022-11-27
Remove mutex around bdk wallet

benthecarman:
Seems we need to remove this mutex to do any DLC stuff as all of rust-dlcs definitions aren't async. I don't totally remember why we used a mutex here, the wallet seems to work fine without it. I did need to do the `unsafe impl Send` and `unsafe impl Sync` for `MutinyWallet` but that should be fine in wasm anyways


benthecarman:
After working on this more, may not need to do this

TonyGiorgio (approved):

PR #140, opened 2022-11-29
Update the proxy URL

TonyGiorgio:


TonyGiorgio:
tested the deployment url and it works

PR #142, opened 2022-11-29
Create MutinyDLCManager

benthecarman:
build is failing for me from some dependency issue, putting up for now though

### DIFF
--- a/node-manager/src/dlc.rs
+++ b/node-manager/src/dlc.rs
@@ -1,0 +1,275 @@
+use crate::localstorage::MutinyBrowserStorage;
+use bdk::blockchain::EsploraBlockchain;
+use bdk::database::Database;
+use bdk::keys::ExtendedKey;
+use bdk::wallet::AddressIndex;
+use bdk::{SignOptions, Wallet};
+use bdk_macros::maybe_await;
+use bip39::Mnemonic;
+use bitcoin::psbt::PartiallySignedTransaction;
+use bitcoin::util::address::Address;
+use bitcoin::util::bip32::{ChildNumber, DerivationPath, ExtendedPrivKey};
+use bitcoin::{Block, Network, Script, Transaction, TxIn, Witness};
+use dlc_manager::error::Error as ManagerError;
+use dlc_manager::error::Error::WalletError;
+use dlc_manager::{Blockchain, Utxo};
+use log::error;
+use secp256k1::Secp256k1;
+use std::str::FromStr;
+use std::sync::Arc;
+use wasm_bindgen_futures::spawn_local;
+
+#[derive(Debug)]
+pub struct MutinyDLCWallet {
+    wallet: Arc<Wallet<MutinyBrowserStorage>>,
+    blockchain: Arc<EsploraBlockchain>,
+    storage: Arc<MutinyBrowserStorage>,
+    mnemonic: Mnemonic,
+}
+
+fn bdk_err_to_manager_err(e: bdk::Error) -> ManagerError {
+    WalletError(Box::new(e))
+}
+
+impl MutinyDLCWallet {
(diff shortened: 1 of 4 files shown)


node-manager/src/dlc.rs
```diff
@@ -0,0 +1,275 @@
...
+        let xkey: ExtendedKey = ExtendedKey::from(xprivkey);
+        let master_xprv = xkey.into_xprv(network).unwrap();
+        let coin_type = match network {
+            Network::Bitcoin => 0,
+            Network::Testnet => 1,
+            Network::Signet => 1,
+            Network::Regtest => 1,
+        };
+
+        let base_path = DerivationPath::from_str("m/585'").unwrap();
```
TonyGiorgio, line 48: we should try to keep track of these derivation paths and where they are use. Phantom also uses a shared derivation path as well. And some of the node derivation paths should probably be hard coded so they aren't used either.

node-manager/src/dlc.rs
```diff
@@ -0,0 +1,275 @@
...
+        let network = self.wallet.network();
+        let xprivkey = ExtendedPrivKey::new_master(network, &entropy).unwrap();
+        let xkey: ExtendedKey = ExtendedKey::from(xprivkey);
+        let master_xprv = xkey.into_xprv(network).unwrap();
+        let coin_type = match network {
+            Network::Bitcoin => 0,
+            Network::Testnet => 1,
+            Network::Signet => 1,
+            Network::Regtest => 1,
+        };
```
TonyGiorgio, line 46: Weird that you had to do this, I put some of this in `utils` though if you want to move it there though. No worries if not.

node-manager/src/dlc.rs
```diff
@@ -0,0 +1,275 @@
...
+}
+
+impl dlc_manager::Signer for MutinyDLCWallet {
+    fn sign_tx_input(
+        &self,
+        tx: &mut bitcoin::blockdata::transaction::Transaction,
+        input_index: usize,
+        tx_out: &bitcoin::blockdata::transaction::TxOut,
+        redeem_script: Option<bitcoin::blockdata::script::Script>,
+    ) -> Result<(), ManagerError> {
```
TonyGiorgio, line 70: Now that we've got multiple "wallets", maybe putting this inside some generic BDK wallet wrapper could be useful. Maybe not now but having multiple "wallets" and "chains" can get rather confusing and probably some leakage in terms of reusing logic. I don't think I see anything DLC specific here.

node-manager/src/dlc.rs
```diff
@@ -0,0 +1,275 @@
...
+            .database()
+            .get_sync_time()
+            .map_err(bdk_err_to_manager_err)?
+            .unwrap();
+
+        Ok(last_sync.block_time.height as u64)
+    }
+
+    fn get_block_at_height(&self, _height: u64) -> Result<Option<Block>, ManagerError> {
+        Ok(None)
```
TonyGiorgio, line 273: did you not need to add async functions after all? Is it not actually needed?

TonyGiorgio (commented):
nothing terribly needed to change but some comments. feel free to skip if you want

node-manager/src/dlc.rs
benthecarman: yeah this is just kinda a hack to sign a specific input, but I agree that things could probably be cleaned up. That is kinda what I want to propose for the spiral grant lol

node-manager/src/dlc.rs
benthecarman: Yeah I just changed it to `Option` instead, was simpler: https://github.com/benthecarman/rust-dlc/commit/c13700440c7a5aeb8edd9ce27cfc3e1c36a92217

This is only needed for the DLC channels stuff so I we'd have to punt on that part for now.

PR #144, opened 2022-11-30
Add mutiny to mutiny serverside proxy

TonyGiorgio:


websocket-tcp-proxy/src/main.rs
```diff
@@ -1,28 +1,69 @@
...
+use std::sync::Arc;
+use std::{env, fmt, str::FromStr};
+use tokio::sync::mpsc::{self, Sender};
 use tokio::{
     io::{AsyncReadExt, AsyncWriteExt},
     net::TcpStream,
 };
 use tower_http::trace::{DefaultMakeSpan, TraceLayer};
 
+const PUBKEY_BYTES_LEN: i8 = 33;
```
benthecarman, line 25: you cast this to `usize` a bunch, should just set this to `usize` here instead of `i8`

node-manager/src/socket.rs
```diff
@@ -0,0 +1,413 @@
...
+        self.connected.load(Ordering::Relaxed)
+    }
+
+    pub fn reconnect(&mut self, conn: Arc<Proxy>) {
+        let (send_to_multi_socket, read_from_sub_socket): (Sender<Message>, Receiver<Message>) =
+            unbounded();
+
+        let socket_map: Arc<Mutex<HashMap<Vec<u8>, Sender<Message>>>> =
+            Arc::new(Mutex::new(HashMap::new()));
+
```
TonyGiorgio, line 149: I think this is one problem i need to explore. I shouldn't create a new map, but delete the old maps and sub sockets..

node-manager/src/socket.rs
```diff
@@ -0,0 +1,413 @@
...
+        let connected_copy = self.connected.clone();
+        debug!("spawning multi socket connection reader");
+        spawn_local(async move {
+            while let Some(msg) = conn_copy.read.lock().await.next().await {
+                if let Ok(Message::Bytes(msg)) = msg {
+                    // parse the msg and see which pubkey it belongs to
+                    if msg.len() < PUBKEY_BYTES_LEN {
+                        debug!("msg not long enough to have pubkey, ignoring...");
+                        continue;
+                    }
```
benthecarman, line 182: Could this just be we need to wait for more bytes? Might need to implement some buffering, probably fine for now if you got payments working

node-manager/src/socket.rs
```diff
@@ -0,0 +1,413 @@
...
+
+                    // now send that data to the right subsocket
+                    let socket_lock = socket_map_copy.lock().await;
+                    let found_subsocket = socket_lock.get(id_bytes);
+                    match found_subsocket {
+                        Some(s) => {
+                            match s.send(Message::Bytes(message_bytes.to_vec())) {
+                                Ok(_) => (),
+                                Err(e) => error!("error sending msg to channel: {}", e),
+                            };
```
benthecarman, line 193: `if let Err(e) = ...` would be better here unless we plan on doing something in the `Ok()` case

node-manager/src/socket.rs
```diff
@@ -0,0 +1,413 @@
...
+                                        );
+                                        }
+                                    }
+                                }
+                                Err(_) => {
+                                    // TODO delete the new subsocket
+                                    error!(
+                                        "peer manager could not handle subsocket for: {:?}",
+                                        id_bytes
+                                    );
```
benthecarman, line 249: should do these todos

node-manager/src/socket.rs
```diff
@@ -0,0 +1,413 @@
...
+                                Ok(_) => {
+                                    debug!("gave new subsocket to peer manager: {:?}", id_bytes);
+                                    schedule_descriptor_read(
+                                        inbound_subsocket,
+                                        peer_manager_copy.clone(),
+                                    );
+
+                                    // now that we have the inbound connection, send the original
+                                    // message to our new subsocket descriptor
+                                    match socket_map_copy.lock().await.get(id_bytes) {
```
benthecarman, line 224: could we have `new_inbound_connection` return the subsocket so we don't need this lookup

node-manager/src/socket.rs
```diff
@@ -0,0 +1,413 @@
...
+
+        let read_channel_copy = self.read_from_sub_socket.clone();
+        let conn_copy_send = self.conn.clone();
+        let connected_copy_send = self.connected.clone();
+        debug!("spawning multi socket channel reader");
+        spawn_local(async move {
+            loop {
+                if let Ok(msg) = read_channel_copy.try_recv() {
+                    match msg {
+                        Message::Text(_) => (),
```
benthecarman, line 268: should probably log a something for the text case

node-manager/src/socket.rs
```diff
@@ -0,0 +1,413 @@
...
+}
+
+impl ReadDescriptor for SubWsSocketDescriptor {
+    async fn read(&self) -> Option<Result<Message, gloo_net::websocket::WebSocketError>> {
+        debug!("starting subsocket channel reader");
+        loop {
+            if let Ok(msg) = self
+                .read_channel
+                .try_recv()
+                .map_err(|_| gloo_net::websocket::WebSocketError::ConnectionError)
```
benthecarman, line 363: I don't think this map_err does anything, talked to you IRL

websocket-tcp-proxy/README.md
```diff
@@ -6,6 +6,8 @@ it's websockify but with the client gets to pick what tcp address to connect to,
 
 ## How to test
 
+You can change default port by setting `MUTINY_PROXY_PORT=3002` or whatever your port should be.
```
benthecarman, line 9: I think we have the default as 3001, should use that here so it's not confusing

websocket-tcp-proxy/src/main.rs
```diff
@@ -1,28 +1,71 @@
...
+    D: Deserializer<'de>,
+    T: FromStr,
+    T::Err: fmt::Display,
+{
+    let opt = Option::<String>::deserialize(de)?;
+    match opt.as_deref() {
+        None | Some("") => Ok(None),
+        Some(s) => FromStr::from_str(s).map_err(de::Error::custom).map(Some),
+    }
+}
```
benthecarman, line 52: nit: found a more generalized version online: https://github.com/serde-rs/serde/issues/1425#issuecomment-462282398

could try to do that instead incase we have more complex params in the future

websocket-tcp-proxy/src/main.rs
```diff
@@ -105,6 +148,152 @@ async fn handle_socket(mut socket: WebSocket, ip: String, port: String) {
...
+    // This is important so that only the node with the
+    // private key can read and send messages through
+    // this socket.
+    let owner_id_bytes: Vec<u8> = if let Ok(b) = FromHex::from_hex(identifier.as_str()) {
+        b
+    } else {
+        // drop the connection if we didn't get a good pubkey hex
+        tracing::error!("could not parse hex string identifier");
+        return;
+    };
```
benthecarman, line 211: `unwrap_or_else` would be cleaner here

websocket-tcp-proxy/src/main.rs
```diff
@@ -105,6 +148,152 @@ async fn handle_socket(mut socket: WebSocket, ip: String, port: String) {
...
+
+    tracing::debug!("listening for {identifier} websocket or consumer channel",);
+    loop {
+        tokio::select! {
+            // The websocket owner is sending a message to some peer
+            res  = socket.recv() => {
+                if let Some(msg) = res {
+                    if let Ok(Message::Binary(msg)) = msg {
+                        // parse the first 33 bytes to find the ID to send to
+                        if msg.len() < PUBKEY_BYTES_LEN{
```
benthecarman, line 232: nit: looks like a spacing issue with the `{`

websocket-tcp-proxy/src/main.rs
```diff
@@ -105,6 +148,152 @@ async fn handle_socket(mut socket: WebSocket, ip: String, port: String) {
...
+    // to later. The consumer here is to listen to events
+    // that should be sent down the websocket that owns this.
+    let (tx, mut rx) = mpsc::channel::<MutinyWSCommand>(32);
+    state
+        .lock()
+        .await
+        .insert(owner_id_bytes.clone(), tx.clone());
+
+    tracing::debug!("listening for {identifier} websocket or consumer channel",);
+    loop {
```
benthecarman, line 225: there's a bunch of TODOs in this loop, idk how needed those are now, might be good to bang out any of the easy ones

node-manager/src/node.rs
```diff
@@ -591,7 +704,12 @@ impl Node {
                 mutiny_invoice.is_send = true;
                 Ok(mutiny_invoice)
             }
-            Err(_) => {
+            Err(e) => {
+                error!("failed to make payment: {:?}", e);
+                // call list channels to see what our channels are
+                let current_channels = self.channel_manager.list_channels();
+                debug!("current channel details: {:?}", current_channels);
```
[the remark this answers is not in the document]
TonyGiorgio: I would agree but then we'd get no info of the channel if it was not usable. Like why was it not usable? The usability of the channel or not would be inclusive by listing all channels.

node-manager/src/socket.rs
TonyGiorgio: I don't think this should ever happen. Buffering shouldn't be needed with these websocket libraries, it should handle that.

PR #153, opened 2022-12-02
Add DLC storage implementation

benthecarman:
implement DLC storage traits for browser storage

### DIFF
--- a/node-manager/src/dlcstorage.rs
+++ b/node-manager/src/dlcstorage.rs
@@ -1,0 +1,365 @@
+use crate::error::{MutinyError, MutinyStorageError};
+use crate::localstorage::MutinyBrowserStorage;
+use anyhow::Context;
+use bitcoin_hashes::hex::ToHex;
+use dlc_manager::chain_monitor::ChainMonitor;
+use dlc_manager::channel::accepted_channel::AcceptedChannel;
+use dlc_manager::channel::offered_channel::OfferedChannel;
+use dlc_manager::channel::signed_channel::{SignedChannel, SignedChannelStateType};
+use dlc_manager::channel::{Channel, FailedAccept, FailedSign};
+use dlc_manager::contract::accepted_contract::AcceptedContract;
+use dlc_manager::contract::offered_contract::OfferedContract;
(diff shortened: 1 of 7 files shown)


node-manager/src/dlcstorage.rs
```diff
@@ -0,0 +1,365 @@
...
+
+    fn update_contract(&mut self, contract: &Contract) -> Result<(), Error> {
+        let serialized = serialize_contract(contract)?;
+        match contract {
+            a @ Contract::Accepted(_) | a @ Contract::Signed(_) => {
+                let key = format! {"{DLC_CONTRACT_KEY_PREFIX}{}", a.get_temporary_id().to_hex()};
+                MutinyBrowserStorage::delete(key);
+            }
+            _ => {}
+        };
```
TonyGiorgio, line 69: Should you delete after `set` instead? What if there's a problem saving the contract?

node-manager/src/dlcstorage.rs
```diff
@@ -0,0 +1,365 @@
...
+        let result: Result<Vec<u8>, MutinyStorageError> = self.get(DLC_CHAIN_MONITOR_KEY);
+
+        match result {
+            Ok(bytes) => {
+                let mon = ChainMonitor::deserialize(&mut Cursor::new(bytes)).map_err(|_| {
+                    Error::StorageError("Failed to decode ChainMonitor".to_string())
+                })?;
+                Ok(Some(mon))
+            }
+            Err(_) => Ok(None),
```
TonyGiorgio, line 189: Could you match on error type here? To distinguish a problem getting the value vs if a value does not currently exist?

node-manager/src/localstorage.rs
```diff
@@ -85,6 +87,36 @@ impl MutinyBrowserStorage {
         map
     }
 
+    pub(crate) fn get_data_with_prefix<T: Serializable>(
```
TonyGiorgio, line 90: how is this different than `scan`?

benthecarman: scan checks the key, this checks the value

TonyGiorgio (approved):

PR #165, opened 2022-12-04
wip in-app logger

futurepaul:


TonyGiorgio:
can we still get this? would be useful to have. not sure what it looks like but maybe hidden in settings or something.

TonyGiorgio:

PR #167, opened 2022-12-05
Revert "Remove mutex around bdk wallet"

benthecarman:
This broke several things, will need to figure out how to properly initate the dlc wallet


TonyGiorgio (approved):
Code changes look good. For DLC's, I'm guessing it's because of the lack of async traits. Would a fork be in order or could we use `std::sync::Mutex` just for DLCs?

benthecarman:
> Code changes look good. For DLC's, I'm guessing it's because of the lack of async traits. Would a fork be in order or could we use `std::sync::Mutex` just for DLCs? 

I think we'll just need to make a copy of the bdk wallet, will see

Issue #176, opened 2022-12-06
if network changes it should clear state, then save settings


futurepaul:
solve chicken egg problem! probably warn about clearing too

PR #178, opened 2022-12-06
clear state if the network changes

futurepaul:


futurepaul:
closes #176

frontend/src/components/SettingStringsEditor.tsx
```diff
@@ -12,6 +12,13 @@ export default function SettingStringsEditor() {
 
     async function handleSaveSettings() {
         try {
+            let existingNetwork = getExistingSettings().network;
+            if (existingNetwork !== nodeManagerSettings.network) {
+                if (window.confirm("Changing networks will delete your node's state. This can't be undone!")) {
+                    localStorage.clear();
+                    window.location.reload();
```
benthecarman, line 19: Do we need to write to storage the network on confirm?

TonyGiorgio: Yeah I think after localStorage.clear() it needs to write network.

TonyGiorgio (changes requested):

frontend/src/components/SettingStringsEditor.tsx
futurepaul: yeah this happens in the `setup` function

PR #185, opened 2022-12-06
fix receive screen skip

futurepaul:
the magic line was 
```
cacheTime: 0
```

### DIFF
--- a/frontend/src/components/MainBalance.tsx
+++ b/frontend/src/components/MainBalance.tsx
@@ -15,7 +15,7 @@ function prettyPrintUnconfirmed(b: MutinyBalance): string {
 export default function MainBalance() {
     const { nodeManager } = useContext(NodeManagerContext);
 
-    const { data: balance } = useQuery({
+    const { isPreviousData, data: balance } = useQuery({
         queryKey: ['balance'],
         queryFn: () => {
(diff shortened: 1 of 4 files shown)


frontend/src/components/MainBalance.tsx
```diff
@@ -24,6 +24,9 @@ export default function MainBalance() {
         enabled: !!nodeManager,
     })
     return (<div className="flex flex-col gap-4 cursor-pointer">
+        <p>
+            {isPreviousData && "is previous"}
+        </p>
```
benthecarman, line 29: I think this is just debug code

frontend/src/routes/ReceiveQR.tsx
```diff
@@ -55,7 +55,8 @@ export default function ReceiveQR() {
         },
         enabled: !!nodeManager,
         // Don't want a new address each time they focus the window
-        refetchOnWindowFocus: false
+        refetchOnWindowFocus: false,
+        cacheTime: 0
```
benthecarman, line 59: maybe add a comment here on why this is important

frontend/src/components/MainBalance.tsx
futurepaul: thank you

PR #199, opened 2022-12-11
Potential fixes for MutinyChain

benthecarman:
Copies some changes made in https://github.com/lightningdevkit/rust-lightning/pull/1870

We still can't directly copy paste as we need some deps to update

### DIFF
--- a/node-manager/src/chain.rs
+++ b/node-manager/src/chain.rs
@@ -10,7 +10,7 @@ use lightning::chain::chaininterface::{
     BroadcasterInterface, ConfirmationTarget, FeeEstimator, FEERATE_FLOOR_SATS_PER_KW,
 };
 use lightning::chain::{Confirm, Filter, WatchedOutput};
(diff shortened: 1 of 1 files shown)


node-manager/src/chain.rs
```diff
@@ -343,7 +345,10 @@ impl MutinyChain {
             }
         }
 
-        *self.watched_transactions.lock().unwrap() = to_watch;
+        let mut locked_watched_transactions = self.watched_transactions.lock().unwrap();
+        for txid in to_watch {
+            locked_watched_transactions.insert(txid);
+        }
```
TonyGiorgio, line 351: Should we `.clear()` first or do we really need to just be inserting new ones to add to it? The old behavior did a replacement.

benthecarman: I think we are supposed to add, that's what is done in the ldk PR

TonyGiorgio (approved):

PR #201, opened 2022-12-11
Exclude expired invoices from list_invoices

benthecarman:


node-manager/src/node.rs
```diff
@@ -614,20 +615,25 @@ impl Node {
...
-                            mutiny_invoice.is_send = !inbound;
-                            mutiny_invoice.paid = matches!(i.status, HTLCStatus::Succeeded);
-                            mutiny_invoice.amount_sats =
-                                if let Some(inv_amt) = invoice.amount_milli_satoshis() {
-                                    if inv_amt == 0 {
-                                        i.amt_msat.0.map(|a| a / 1_000)
+                            let paid = matches!(i.status, HTLCStatus::Succeeded);
+                            if invoice.would_expire(now) && !paid {
+                                None
+                            } else {
```
TonyGiorgio, line 621: What about invoices that are stuck passed the expire date (so would be with `Pending` status or maybe there's another status like `InFlight`?)

Wouldn't want someone's pending in flight payment to disappear and they arent sure what happened.

TonyGiorgio (changes requested):

node-manager/src/node.rs
benthecarman: Added an `InFlight` status that should fix this

TonyGiorgio (approved):
Checked all the other areas where `HTLCStatus` is used and looks good to me.

Issue #190, opened 2022-12-06
camera doesn't stop on browser


TonyGiorgio:
When I am in the browser and go through the send phase, it doesn't stop the camera and it persists until restart.

PR #203, opened 2022-12-11
Turn off Qr scanner after pasting from clipboard

benalleng:
As noted in issue #190 the qr scanner doesn't properly turn off when pasting an address from a clipboard and only on a successful qr scan.  
Upon troubleshooting this problem i discovered it seemed to do with the fact that the input box and the QrCodeSanner component were separate and joined together in the send route. The qr scanner is only looking for valid data scanned from the qr code and not valid data typed in the input field. At first I attempted to import the qrCodeRef to find the existing qr scan and then kill it upon calling the handleContinue() function of the send route with,
`qrCodeRef.current?.stop()`
however, I was not able to get react to successfully identify the qr scanner that was started in the separate component so I instead decide to build matching logic for the input field inside of the QrCodeScanner component to ensure that when the continue button is pressed it can find and kill the existing qr scanner.
Lastly I added a small amount of styling because that qr scanner is too damn BIG!

I am happy that it works but would have liked to simplify the logic and not double up the code :sweat_smile: just thought I'd give a hand at fixing this bug that i thought could be fixed more easily.
If there is a way to find and kill the Qr scanner from outside of the component this code could be much more simplified.

### DIFF
--- a/frontend/src/components/ReceiveUnified.tsx
+++ b/frontend/src/components/ReceiveUnified.tsx
@@ -14,8 +14,8 @@ export default function ReceiveUnified({ bip21String }: { bip21String: string })
 
     const { address, options } = bip21.decode(bip21String) as MutinyBip21;
 
-    const { isLoading: isCheckingAddress } = useQuery({
-        queryKey: ['checktransaction'],
+    useQuery({
+        queryKey: ['checktransaction', address],
         queryFn: async () => {
             console.log("Checking address:", address);
             const tx = await nodeManager?.check_address(address);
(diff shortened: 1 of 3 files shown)


frontend/src/routes/Send.tsx
```diff
@@ -129,10 +127,6 @@ function Send() {
       </header>
```
futurepaul, line 127: Seems like it would help if the <Close /> component would accept some onClose callback that overrides the default behavior of navigating to root, that way we can pass cleanup logic when necessary

futurepaul (changes requested):
thanks for digging into this, def a hairy part of the codebase

I think moving the ref state into Send.tsx and passing it as a prop to QrCodeScanner solves your problem, along with giving <Close /> an optional callback. let me know if that won't work for some reason

futurepaul:
okay #221 should hopefully unblock you on this, let me know if you have any qs!

benalleng:
Going to give this another crack this weekend!

PR #207, opened 2022-12-12
Better mutiny to mutiny

TonyGiorgio:


node-manager/src/socket.rs
```diff
@@ -140,17 +141,19 @@ impl MultiWsSocketDescriptor {
...
+        // if reconnecting master socket, disconnect and clear all subsockets
+        for (_id, (subsocket, _sender)) in self.socket_map.lock().await.iter_mut() {
+            // tell the subsocket to stop processing
+            // and ldk to disconnect that peer
+            subsocket.disconnect_socket();
+            self.peer_manager
+                .socket_disconnected(&WsSocketDescriptor::Mutiny(subsocket.clone()));
+        }
 
+        self.socket_map.lock().await.clear();
```
benthecarman, line 155: you call `self.socket_map.lock()` twice here, would be better to do `let socket_map = self.socket_map.lock()`, there's probably a potential race here with that

benthecarman (changes requested):

node-manager/src/node.rs
```diff
@@ -925,7 +940,7 @@ pub(crate) async fn connect_peer(
     debug!("connected to peer: {:?}", peer_connection_info);
 
     let sent_bytes = descriptor.send_data(&initial_bytes, true);
-    trace!("sent {sent_bytes} to node: {}", peer_connection_info.pubkey);
+    debug!("sent {sent_bytes} to node: {}", peer_connection_info.pubkey);
```
benthecarman, line 943: should revert this back to trace

node-manager/src/socket.rs
```diff
@@ -4,15 +4,23 @@ use crate::utils;
...
 use std::hash::Hash;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::Arc;
 use wasm_bindgen_futures::spawn_local;
 
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize)]
+enum MutinyProxyCommand {
+    Disconnect { to: Vec<u8>, from: Vec<u8> },
```
benthecarman, line 21: I think for this json it will be `{"to": [1,2,1], "from": [23,4,2]}` so we could have some potential overlap, might be worth adding a tag to this so we know which command and don't have potential overlap

node-manager/src/socket.rs
```diff
@@ -161,96 +174,157 @@ impl MultiWsSocketDescriptor {
...
+                            let (id_bytes, message_bytes) = msg.split_at(PUBKEY_BYTES_LEN);
+
+                            // now send that data to the right subsocket;
+                            let socket_lock = socket_map_copy.lock().await;
+                            let found_subsocket = socket_lock.get(id_bytes);
+                            match found_subsocket {
+                                Some((_subsocket, sender)) => {
+                                    match sender.send(Message::Bytes(message_bytes.to_vec())) {
+                                        Ok(_) => {
+                                            debug!(
```
benthecarman, line 250: this should probably be `trace`

node-manager/src/socket.rs
```diff
@@ -161,96 +174,157 @@ impl MultiWsSocketDescriptor {
...
+
+                                            // now that we have the inbound connection, send the original
+                                            // message to our new subsocket descriptor
+                                            match socket_map_copy.lock().await.get(id_bytes) {
+                                                Some((_subsocket, sender)) => {
+                                                    match sender.send(Message::Bytes(
+                                                        message_bytes.to_vec(),
+                                                    )) {
+                                                        Ok(_) => {
+                                                            debug!("sent incoming message to new subsocket channel: {:?}", id_bytes)
```
benthecarman, line 297: this should probably be `trace`

PR #204, opened 2022-12-12
Bring upload state from file input into view

benalleng:
As mentioned in issue #192 the upload state from file was not rendering. Adding a label and hiding the default input file upload boilerplate I styled it to look like the buttons above it.

PR #209, opened 2022-12-12
Fix inputting a keysend on the send page

benalleng:
As mention in issue #200 I added the logic to handle p2pkh as an option for the payment type. 
As you can see in the commit history I'm still a green coder! lol


frontend/src/routes/Send.tsx
```diff
@@ -93,6 +93,14 @@ function Send() {
       await navigateForInvoice(destination)
     } else if (paymentType === PaymentType.bip21) {
       await navigateForBip21(destination)
+    } else if (paymentType === PaymentType.keysend) {
+      if (sendAll === "true") {
+        const params = objectToSearchParams<SendConfirmParams>({ destination, all: "true" })
+        navigate(`/send/confirm?${params}`)
+      } else {
+        const params = objectToSearchParams<SendConfirmParams>({ destination })
+        navigate(`/send/amount?${params}`)
```
benthecarman, line 102: Doing send all for keysend is tricky and not something we currently support, can just exclude that do just use the `else` case here

benthecarman:
Looks like your commit history got merged with #204

frontend/src/routes/Send.tsx
benalleng: Ok! should i go ahead and remove the if statement all together or should it send some sort of UI for the user to understand they can't sendAll for keysend

benthecarman (approved):

frontend/src/routes/Send.tsx
benthecarman: This looks good! The sendAll parameter is never exposed through the api so you should be okay.

benthecarman:
Updated the title to be more correct. P2PKH is an onchain address type, whereas a key send is for sending to a lightning node.

TonyGiorgio (approved):
LGTM, though not sure if we want @futurepaul 's look at it first.

Also note to the PR merger, this should be squashed and then merged.

benthecarman:
> sweet thank you... miss that exhaustive pattern matching in rust!

we could use a `switch` here instead

futurepaul:
true but I think we'll want to do more cleanup / logic centralizing / refactoring in the future so this is good for now

PR #215, opened 2022-12-13
List disconnected peers

### LINKED ISSUE #147: list peers should show all peers, even one we aren't connected to
so we can clear out peers that are disconnected for instance

benthecarman:
Closes #147

The return type for this is a `Vec<String>`, maybe we should create a `MutinyPeer` type that has information like `connection_string` and `is_connected` etc

### DIFF
--- a/frontend/src/routes/Peers.tsx
+++ b/frontend/src/routes/Peers.tsx
@@ -20,9 +20,11 @@ function SinglePeer({ peer }: { peer: MutinyPeer }) {
 
         console.log(myNodes);
 
-        const myNode = myNodes[0]
-        await nodeManager?.disconnect_peer(myNode, peer.pubkey);
-        queryClient.invalidateQueries({ queryKey: ['peers'] })
+        if (window.confirm("Are you sure you want to delete this peer?")) {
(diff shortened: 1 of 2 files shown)


TonyGiorgio (changes requested):
Yeah actually I think we might need to pass back the object with whether or not it is connected. The frontend has an X button on it so what would happen (or the intention to what should happen) when they click an X on a disconnected peer? 

For one, I wouldn't expect disconnected peers showing up exactly the same as connected ones. That's going to be hard to troubleshoot when issues arise due to disconnected peers. 

For two, if a peer is disconnected and I see it is disconnected, when I click X, I would probably intend for it to be deleted from the DB.

benthecarman:
Updates: 
image

image

TonyGiorgio (approved):
Rust code looks good, thanks for doing that. @futurepaul might have some additional thoughts when we were testing this.

futurepaul (approved):
awesome, made a small commit hope you don't mind

benthecarman:
> awesome, made a small commit hope you don't mind

Fixed a small typo

PR #217, opened 2022-12-13
Hex encode ldk sorage items

benthecarman:
Before it would store things byte arrays like: `[1,1,0,0,0,0,0,0,0,15,108,103,13...]`
Now it will be hex, should use less space

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -8,6 +8,7 @@ on:
 
 jobs:
   Deploy-Preview:
+    environment: Preview
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


futurepaul:
will this make old save states incompatible with new ones? not that that should block, just curious

benthecarman:
I tested switching from previous encoding to the new encoding and it did not break, I think rust just supports reading byte arrays as both formats

TonyGiorgio (approved):

futurepaul:
Screen Shot 2022-12-13 at 4 01 12 PM

Screen Shot 2022-12-13 at 4 01 25 PM

PR #221, opened 2022-12-13
give close button more powers

futurepaul:
should help with #203 

for now it just means the sub-pages in the manager will nav to their parent on close


frontend/src/components/Close.tsx
```diff
@@ -1,8 +1,29 @@
...
+
+    let navigate = useNavigate();
+
+    function handleClose() {
+        if (onClose) {
+            onClose();
+        }
+        if (to) {
+            navigate(to);
+        } else {
```
benthecarman, line 19: should probably put a comment on why you're not using an `else if` here

TonyGiorgio (approved):

PR #233, opened 2022-12-16
Update ldk and delete DLC stuff

benthecarman:


futurepaul (approved):
nice. seems like this is the first one to get in because other ones build on it?

PR #240, opened 2022-12-17
Fix bug with receiving zero amount invoices

benthecarman:
This would be saved as 0 in storage instead of being updated with the amount the user paid.


benthecarman:
This has a silent merge conflict with #233

benthecarman:
just gonna do this in #233

Issue #229, opened 2022-12-15
Some UX feedback


sahilc0:
_Note1: I can make these separate tickets if that's easier!_
_Note2: If there's a Figma for this I can make specific change suggestions. If not we could start one!_

**Goals**
- Identify quick wins

**Non-goals**
- Rethink brand identity (colors, type, etc)

**A few super quick wins**
- Trigger numeric keyboard any time a number field is focused on (see figma)
- Copy button is going outside of the margin (see figma)
- Move menu button trigger to top right, so position is consistent
- Make all buttons full width on mobile (instead of staggered and left aligned)
- Center sats amount on homepage - this is the hero number!

**Bigger changes**
- Font readability on some pages (eg. Transaction page, font weight on buttons)
- Warning banners for state/seed backups
- Confusing guidance around saving seed vs state (see figma/video)

**Figma notes**
https://www.figma.com/file/TAhivKvpgOl6C4oInznnuS/Mutiny-feedback?node-id=0%3A1&t=a2ZqrUF7wq8W6LbV-11

**Video walkthrough**
https://photos.app.goo.gl/2D3WVKpxujtorZcT8

futurepaul:
sweet this is good stuff. gonna push back on centering / full-width buttons for now, those were deliberate decisions to be different... will have to think through it more. everything else is great and gonna make it a checklist:

## A few super quick wins

- [x] Trigger numeric keyboard any time a number field is focused on (see figma)
- [ ] Copy button is going outside of the margin (see figma)
- [ ] Move menu button trigger to top right, so position is consistent

## Bigger changes

- [ ] Font readability on some pages (eg. Transaction page, font weight on buttons)
- [ ] Warning banners for state/seed backups
- [ ] Confusing guidance around saving seed vs state (see figma/video)

sahilc0:
Niceeeee. Ping me if it would be useful to have help on the specifics for "Bigger changes" (eg. font readability), warning banner design, etc

PR #242, opened 2022-12-17
change input type=number in send and recieve amount

benalleng:
as referenced in #229 the inputs for where only numbers should be are actually text currently.  Hoping to prevent non-numbers and negative numbers  where  those would cause errors

### DIFF
--- a/frontend/src/routes/Receive.tsx
+++ b/frontend/src/routes/Receive.tsx
@@ -23,12 +23,8 @@ function Receive() {
             setAmount(' ')
             toast("That doesn't look right")
             return
-        } else if (parseInt(amount) <= 0) {
-            setAmount('')
-            toast("You can't receive nothing")
-            return
         }
-        if (parseInt(amount) > 0) {
+        if (amount) {
             const params = objectToSearchParams<ReceiveParams>({ amount, description })
             // Important! Otherwise we might see a stale bip21 code
             queryClient.invalidateQueries({ queryKey: ['bip21'] })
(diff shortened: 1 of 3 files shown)


frontend/src/routes/Receive.tsx
```diff
@@ -16,11 +17,20 @@ function Receive() {
...
+            navigate('/')
+            toast("You can't receive nothing")
+            return
+        }
+        if (amount) {
         const params = objectToSearchParams<ReceiveParams>({ amount, description })
         // Important! Otherwise we might see a stale bip21 code
         queryClient.invalidateQueries({ queryKey: ['bip21'] })
         navigate(`/receive/qr?${params}`)
+          }
```
benthecarman, line 33: this formatting looks weird, should run the linter

benalleng: I just briefly went over the error handling again for the toast strings and I realized they refuse to display on the receive page. they work everywhere else but to see them on receive you must first navigate home

benalleng:
I just briefly went over the error handling again for the toast strings and I realized they refuse to display on the receive page. they work everywhere else but to see them on receive you must first navigate home. I'll need to figure out why it is happening on this page before this pr is ready

benalleng:
sorry for the wave of commits, I kept thinking of more and more edge cases...
It should now accept only numbers while allowing you to still create zero-amount invoices along with adding commas for when you need to keep track of typing larger numbers.
As well on phones it appears to be a number text field but still behaves like a text field so any non-numeric characters pasted in accidentally will still be treated with the same logic as a on desktop.
Shout out to chatGPT for coming in clutch.

futurepaul (approved):
nice, thank you! I have some followup changes so we can format this with underscores like in the ui (`100_000`) but this is a great start

PR #243, opened 2022-12-18
Async events

### LINKED ISSUE #241: Switch to async background processor in ldk
This could be a potential cause of our force closures as we just always write our channel state with the background processor

benthecarman:
Closes #241

Changes to use async event processing, hopefully helps with force closures

### DIFF
--- a/node-manager/src/node.rs
+++ b/node-manager/src/node.rs
@@ -28,7 +28,7 @@ use crate::error::MutinyStorageError;
 use crate::proxy::Proxy;
 use crate::socket::WsTcpSocketDescriptor;
 use crate::{
-    background::{BackgroundProcessor, GossipSync},
+    background::{process_events_async, GossipSync},
     error::MutinyError,
     keymanager::{create_keys_manager, pubkey_from_keys_manager},
     logging::MutinyLogger,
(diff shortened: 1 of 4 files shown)


node-manager/src/node.rs
```diff
@@ -275,27 +274,39 @@ impl Node {
...
+                    |e| ev.handle_event(e),
+                    background_chain_monitor.clone(),
+                    background_processor_channel_manager.clone(),
+                    gs,
+                    background_processor_peer_manager.clone(),
+                    background_processor_logger.clone(),
+                    Some(scorer.clone()),
+                    |d| async move {
+                        sleep(d.as_millis() as i32).await;
+                        true
```
TonyGiorgio, line 303: What is this `true` here for? Do we need to handle some instances where it should be false?

benthecarman: For the sleeper is has this description
```rust
/// `sleeper` should return a future which completes in the given amount of time and returns a
/// boolean indicating whether the background processing should exit. Once `sleeper` returns a
/// future which outputs true, the loop will exit and this function's future will complete.
```

I am not too sure by when it should exit, we don't have an extra processing that needs to be done so I think just `true` is fine.

benthecarman: image

TonyGiorgio (approved):

PR #244, opened 2022-12-19
Resolve dns on proxy side

benthecarman:
This changes so we resolve the dns on the proxy side, this should allow us to connect to nodes using dns as well as set us up for tor support in the future

can test on mainnet with 
```
0230a5bca558e6741460c13dd34e636da28e52afd91cf93db87ed1b0392a7466eb@node.blixtwallet.com
```

### DIFF
--- a/node-manager/src/node.rs
+++ b/node-manager/src/node.rs
@@ -17,6 +17,7 @@ use lightning::util::logger::{Logger, Record};
 use lightning::util::ser::Writeable;
 use lightning_invoice::utils::create_invoice_from_channelmanager_and_duration_since_epoch;
 use std::collections::HashMap;
+use std::net::SocketAddr;
 use std::str::FromStr;
 use std::sync::{Arc, Mutex};
 use wasm_bindgen_futures::spawn_local;
@@ -42,6 +43,7 @@ use instant::SystemTime;
 use lightning::chain::keysinterface::{
     InMemorySigner, KeysInterface, PhantomKeysManager, Recipient,
 };
+use lightning::ln::msgs::NetAddress;
 use lightning::ln::peer_handler::{
     IgnoringMessageHandler, MessageHandler as LdkMessageHandler, PeerManager as LdkPeerManager,
     SocketDescriptor as LdkSocketDescriptor,
@@ -889,11 +891,11 @@ pub(crate) async fn connect_peer(
     // first make a connection to the node
     debug!("making connection to peer: {:?}", peer_connection_info);
     let (mut descriptor, socket_addr) = match peer_connection_info.connection_type {
-        ConnectionType::Tcp(_) => {
+        ConnectionType::Tcp(ref t) => {
             let proxy = Proxy::new(websocket_proxy_addr, peer_connection_info.clone()).await?;
             (
                 WsSocketDescriptor::Tcp(WsTcpSocketDescriptor::new(Arc::new(proxy))),
-                None,
+                try_get_net_addr_from_socket(&t),
             )
         }
         ConnectionType::Mutiny(_) => (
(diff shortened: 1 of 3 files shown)


websocket-tcp-proxy/src/main.rs
```diff
@@ -100,20 +101,30 @@ fn format_addr_from_url(ip: String, port: String) -> String {
...
+    let server_stream = addrs.find_map(|addr| {
+        let connection = block_on(TcpStream::connect(&addr));
+        if let Err(error) = &connection {
+            tracing::error!("Could not connect to {addr}: {error}");
+        };
+
+        connection.ok()
+    });
+
     let mut server_stream = server_stream.unwrap();
```
futurepaul, line 124: this unwrap is bad now because the .is_err case isn't being handled. if the connection fails we panic:

2022-12-19T18:14:26.541204Z ERROR websocket_tcp_proxy: Could not connect to 52.41.246.129:9735: Connection refused (os error 61)
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', websocket-tcp-proxy/src/main.rs:124:43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

node-manager/src/proxy.rs
```diff
@@ -56,12 +56,11 @@ impl Proxy {
...
 
-pub fn tcp_proxy_to_url(proxy_url: String, peer_addr: SocketAddr) -> String {
-    format!(
-        "{proxy_url}/v1/{}/{}",
-        peer_addr.ip().to_string().replace('.', "_"),
-        peer_addr.port()
-    )
+pub fn tcp_proxy_to_url(proxy_url: String, peer_addr: String) -> String {
+    let mut parts = peer_addr.split(':');
+    let host = parts.next().expect("host not found");
```
futurepaul, line 61: seems like this should return a result so we can handle malformed urls

websocket-tcp-proxy/src/main.rs
```diff
@@ -100,20 +101,30 @@ fn format_addr_from_url(ip: String, port: String) -> String {
 
 // Big help from https://github.com/HsuJv/axum-websockify
 async fn handle_socket(mut socket: WebSocket, ip: String, port: String) {
-    let addr = format_addr_from_url(ip, port);
-    let server_stream = TcpStream::connect(&addr).await;
-    tracing::info!("Connect to remote {:#?}", server_stream);
+    let addr_str = format_addr_from_url(ip, port);
+    let mut addrs = addr_str.to_socket_addrs().expect("Failed to resolve");
```
futurepaul, line 105: seems to me like this would panic if we get a bad address, yeah?

futurepaul (changes requested):

benthecarman:
Resolved @futurepaul's comments

node-manager/src/node.rs
```diff
@@ -891,11 +889,11 @@ pub(crate) async fn connect_peer(
     // first make a connection to the node
     debug!("making connection to peer: {:?}", peer_connection_info);
     let (mut descriptor, socket_addr) = match peer_connection_info.connection_type {
-        ConnectionType::Tcp(t) => {
+        ConnectionType::Tcp(_) => {
             let proxy = Proxy::new(websocket_proxy_addr, peer_connection_info.clone()).await?;
             (
                 WsSocketDescriptor::Tcp(WsTcpSocketDescriptor::new(Arc::new(proxy))),
-                Some(get_net_addr_from_socket(t)),
+                None,
```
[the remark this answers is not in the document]
TonyGiorgio: I think `None` is okay but I wonder how `peer_manager.new_outbound_connection` is using it in the first place?

Maybe instead of strictly passing `None`, you could make a `try_get_net_addr_from_string` method that returns Some or None, depending on if it's a socket address or not.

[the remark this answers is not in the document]
benthecarman: Good idea, done

TonyGiorgio (approved):

PR #246, opened 2022-12-19
Better amount inputs

futurepaul:
builds on @benalleng 's work in #242 

switching back and forth between usd and sats was more annoying to code than I anticipated but finally happy with how it works. keeps a source of truth in sats for minimal infinite loop opportunities

Screen Shot 2022-12-19 at 5 54 41 PM
Screen Shot 2022-12-19 at 5 54 34 PM
Screen Shot 2022-12-19 at 5 54 05 PM

### DIFF
--- a/frontend/src/components/AmountInput.tsx
+++ b/frontend/src/components/AmountInput.tsx
@@ -1,0 +1,113 @@
+import { useQuery } from "@tanstack/react-query";
+import prettyPrintAmount from "@util/prettyPrintAmount";
+import { Dispatch, SetStateAction, useContext, useState } from "react";
+import { inputStyle, selectStyle } from "../styles";
+import { NodeManagerContext } from "./GlobalStateProvider";
+
+type Props = {
+    amountSats: string,
+    setAmount: Dispatch<SetStateAction<string>>,
+    accent: "green" | "blue" | "red",
+    placeholder: string,
+}
+
+export enum Currency {
+    USD = "USD",
+    SATS = "SATS"
+}
+
+function satsToUsd(amount: number, price: number): string {
+    let btc = amount / 100000000;
+    let usd = btc * price;
+    return usd.toFixed(2);
+}
+
+function usdToSats(amount: number, price: number): string {
+    let btc = amount / price;
+    let sats = btc * 100000000;
+    return sats.toFixed(0);
+}
+
+export default function AmountInput({ amountSats, setAmount, accent, placeholder }: Props) {
+    const [currency, setCurrency] = useState(Currency.SATS)
+
+    // amountSats will be our source of truth, this is just for showing to the user
+    const [localDisplayAmount, setLocalDisplayAmount] = useState(amountSats)
+
+    const { nodeManager } = useContext(NodeManagerContext);
+
+    const { data: price } = useQuery({
+        queryKey: ['price'],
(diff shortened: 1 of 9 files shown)


frontend/src/components/AmountInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+function satsToUsd(amount: number, price: number): string {
+    let btc = amount / 100000000;
+    let usd = btc * price;
+    return usd.toFixed(2);
+}
+
+function usdToSats(amount: number, price: number): string {
+    let btc = amount / price;
+    let sats = btc * 100000000;
+    return sats.toFixed(0);
```
benthecarman, line 28: we have a function to do btc <-> sats conversion in rust, not sure if we want to do that here instead?

futurepaul: oh right forgot about that one

futurepaul:
can't say this was fun but it did end up battle testing the rust conversion function so that's nice

frontend/src/components/AmountInput.tsx
```diff
@@ -0,0 +1,99 @@
...
+            setLocalDisplayAmount('')
+            return
+        }
+        //Use a regex to replace all commas and underscores with empty string
+        const amount = value.replace(/[_,]/g, "");
+        let parsedAmount = parseInt(amount)
+        if (typeof parsedAmount === "number" && parsedAmount !== 0 && price) {
+            // If the currency is set to usd, we need to convert the input amount to sats and store that in setAmount
+            if (currency === Currency.USD) {
+                setAmount(usdToSats(parsedAmount, price))
```
benalleng, line 42: should this be?
```
setAmount(satsToUSD(parsedAmount, price))
```
This is for where the input is formatting for USD when the Currency.USD option is selected correct?

benalleng: If this is correct I think this is what is causing the input to not be able to have a pre-formatted value of less than $1

node-manager/src/nodemanager.rs
```diff
@@ -921,15 +921,19 @@ impl NodeManager {
 
     #[wasm_bindgen]
     pub fn convert_btc_to_sats(btc: f64) -> Result<u64, MutinyJsError> {
-        if let Ok(amount) = bitcoin::Amount::from_btc(btc) {
+        // rust bitcoin doesn't like extra precision in the float
+        // so we round to the nearest satoshi
+        // explained here: https://stackoverflow.com/questions/28655362/how-does-one-round-a-floating-point-number-to-a-specified-number-of-digits
+        let truncated = 10i32.pow(8) as f64;
+        let btc = (btc * truncated).round() / truncated;
+        if let Ok(amount) = bitcoin::Amount::from_btc(btc as f64) {
```
benthecarman, line 929: Oof this does not sound fun.

benthecarman (commented):
This looks a lot cleaner 👍

frontend/src/components/AmountInput.tsx
futurepaul: yeah it's correct, the "source of truth" is sats so we always need to set that.

yeah I gave up before I could manage to support less than $1. maybe a good follow-on pr?

benthecarman (approved):

PR #247, opened 2022-12-20
encoded string margin

benalleng:
As reference in #229 the copy buttons would go off screen at a certain screen width, this included almost all of the encoded values throughout the app not only the copy buttons. I decided to add some responsiveness so that the lengths change with the screen width. it works by taking the screen width as an argument for the takeN function and multiplying it as the screen size changes.

image

image

I think it could use some slight adjustments but that can be done by simply tweaking the numbers slightly.
hopefully this checks another ux box :heavy_check_mark:


benalleng:
I have a question about where I should have put the screenWidth function. Is it fine where it is as utility function should it be a part of the takeN function? just not sure if I actually should have separated them like I did.

benthecarman:
> I have a question about where I should have put the screenWidth function. Is it fine where it is as utility function should it be a part of the takeN function? just not sure if I actually should have separated them like I did.

Separation is pretty much always better imo

benthecarman (approved):
This looks good to me, this seems like a better way to truncate vs the way we were before.

futurepaul (changes requested):
thanks for working on this! this has been an annoying bit of the design that I keep not fixing

I think this could be DRY-er by making a component that returns a span with all the screen measuring logic internal to it (so you don't have to import useScreenWidth everywhere you want to use this.

Also "n" used to represent the number of characters we wanted to show, now it seems kind of like a magic number. Is there a way to make it more clear what that number does?

benalleng:
@futurepaul You're right about it being a magic number! I really had to think about how the algebra was working there. Pretty easily set up the CodeTruncation component but took a lot of head scratching to figure out what n actually did now. I made a small writeup in the takeNWidth function to explain what it does, though I should probably rename it now that I have it figured out. Probably just good to name it truncator.tsx

For the future: it might be worth diving deeper to allow it to work with all font sizes as it relies on the string length at the moment and not width of the string

futurepaul:
eslint fail:
```
[eslint] 
src/routes/ConnectPeer.tsx
  Line 11:8:  'ScreenMain' is defined but never used  @typescript-eslint/no-unused-vars
  ```

futurepaul:
btw if you'd like more synchronous discussion with us please join our matrix! https://matrix.to/#/!dbavcYAnPJkoAGuMKw:lightninghackers.com?via=lightninghackers.com&via=matrix.org&via=hot.irish

TonyGiorgio:

PR #248, opened 2022-12-22
Use utils helper function for time

benthecarman:
removing some code duplication

### DIFF
--- a/node-manager/src/node.rs
+++ b/node-manager/src/node.rs
@@ -9,7 +9,6 @@ use crate::wallet::MutinyWallet;
 use bitcoin::hashes::sha256::Hash as Sha256;
 use bitcoin::hashes::Hash;
 use bitcoin::Network;
-use instant::Duration;
 use lightning::chain::{chainmonitor, Filter, Watch};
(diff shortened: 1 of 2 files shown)


node-manager/src/node.rs
```diff
@@ -476,10 +474,7 @@ impl Node {
         let amount_msat = amount_sat.map(|s| s * 1_000);
         let invoice = match route_hints.len() {
             0 => {
-                let now = SystemTime::now()
-                    .duration_since(SystemTime::UNIX_EPOCH)
-                    .unwrap()
-                    .as_secs();
+                let now = crate::utils::now();
```
TonyGiorgio, line 477:
```suggestion
                let now = crate::utils::now().as_secs();
```

Should that have `as_secs()` like the rest of them do?

TonyGiorgio (changes requested):

node-manager/src/node.rs
benthecarman: We don't need to here, before we were getting the seconds and then turning back into a duration. This just removes the extra conversion

TonyGiorgio (approved):

PR #250, opened 2022-12-22
Settings menu button moved

benalleng:
Quick pr to see how we like the menu button to match the close button. This makes the two buttons in the same area.

image

image


futurepaul (approved):
cool let's give this a shot

PR #251, opened 2022-12-23
Move to next screen on Enter

benthecarman:


benalleng (approved):
Looks good and works for me!

futurepaul:
hey, this works but the "proper" way to do this is to use a `<form>` element so I'm going to do some add-on commits to this, will involve some layout changes

frontend/src/routes/ConnectPeer.tsx
```diff
@@ -4,7 +4,7 @@ import MutinyToaster from "@components/MutinyToaster";
 import { useQuery } from "@tanstack/react-query";
 import { getFirstNode, getHostname, toastAnything } from "@util/dumb";
 import takeN from "@util/takeN";
-import { useContext, useState } from "react";
+import React, { useContext, useState } from "react";
```
futurepaul, line 7: react types are global I think you don't need to import React for them

futurepaul (changes requested):

futurepaul (approved):
this looks good to me now... tested each screen as I went

PR #271, opened 2022-12-29
Verify network of lightning invoice & addresses

### LINKED ISSUE #270: Fail invoices that are on a different network

benthecarman:
Closes #270

### DIFF
--- a/node-manager/src/nodemanager.rs
+++ b/node-manager/src/nodemanager.rs
@@ -475,6 +475,11 @@ impl NodeManager {
         fee_rate: Option<f32>,
     ) -> Result<String, MutinyJsError> {
         let send_to = Address::from_str(&destination_address)?;
+
+        if send_to.network != self.network {
+            return Err(MutinyJsError::IncorrectNetwork);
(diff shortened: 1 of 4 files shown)


node-manager/src/nodemanager.rs
```diff
@@ -455,11 +456,13 @@ impl NodeManager {
...
-        match self
-            .wallet
-            .send(destination_address, amount, fee_rate)
-            .await
-        {
+        let send_to = Address::from_str(&destination_address)?;
+
+        if send_to.network != self.network {
+            return Err(MutinyJsError::IncorrectNetwork);
+        }
```
TonyGiorgio, line 463: `sweep_wallet` and `check_address` should probably have this check too, right?

TonyGiorgio (changes requested):

node-manager/src/nodemanager.rs
benthecarman: Added

TonyGiorgio (approved):

PR #274, opened 2022-12-30
Copy toast

### LINKED ISSUE #269: Add toast for when you press copy

benalleng:
Should close #269. I also added a cursor pointer class to the copy icon svg to just drive home it is interactable UI just in case.

### DIFF
--- a/frontend/src/components/Copy.tsx
+++ b/frontend/src/components/Copy.tsx
@@ -1,4 +1,6 @@
+import { toast } from "react-hot-toast"
 import { ReactComponent as CopyIcon } from "../images/icons/copy.svg"
+import MutinyToaster from "./MutinyToaster"
 
 type Props = {
   copyValue: string
@@ -8,10 +10,12 @@ const Copy: React.FC<Props> = ({ copyValue }) => {
 
   function handleCopy() {
     navigator.clipboard.writeText(copyValue)
+    toast('Copied to Clipboard!')
   }
 
   return (<div onClick={handleCopy} className="h-10 w-10 min-w-10 text-white active:text-half-faint">
     <CopyIcon />
+    <MutinyToaster />
   </div>)
 
 }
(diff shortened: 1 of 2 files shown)


benalleng:
> Preview URL: https://mutiny-testnet-iz0disa43-paulteam.vercel.app

Neat! fastest place to find a copy is to make a receive for 0 sats

frontend/src/components/Copy.tsx
```diff
@@ -1,4 +1,6 @@
+import { toast } from "react-hot-toast"
 import { ReactComponent as CopyIcon } from "../images/icons/copy.svg"
+import MutinyToaster from "./MutinyToaster"
```
futurepaul, line 3: this means there will be two mutinytoasters on some pages... but it doesn't seem to cause double toasts so no harm.

we can probably end up with one global mutinytoaster at some point but this is fine for now

frontend/src/images/icons/copy.svg
```diff
@@ -1 +1 @@
-<svg fill="none" aria-label="Copy" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><path d="M4 2h11v2H6v13H4V2zm4 4h12v16H8V6zm2 2v12h8V8h-8z" fill="currentColor"/></svg>
\ No newline at end of file
+<svg class="cursor-pointer" fill="none" aria-label="Copy" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><path d="M4 2h11v2H6v13H4V2zm4 4h12v16H8V6zm2 2v12h8V8h-8z" fill="currentColor"/></svg>
```
futurepaul, line 1: I would've added this class to the wrapper div not the svg, but no biggie

futurepaul (approved):
sweet, love having the preview url!!

PR #277, opened 2022-12-30
Small simplification in MutinyError

benthecarman:
Reduces some code duplication

### DIFF
--- a/node-manager/src/error.rs
+++ b/node-manager/src/error.rs
@@ -210,7 +210,7 @@ pub enum MutinyJsError {
     /// A failure to generate a mnemonic seed.
     #[error("Failed to generate seed")]
     SeedGenerationFailed,
-    /// User provided invalid menmonic.
+    /// User provided invalid mnemonic.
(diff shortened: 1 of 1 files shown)


node-manager/src/error.rs
```diff
@@ -284,11 +284,7 @@ impl From<bitcoin::util::address::Error> for MutinyJsError {
 
 impl From<PaymentError> for MutinyJsError {
     fn from(e: PaymentError) -> Self {
-        match e {
-            PaymentError::Invoice(_) => Self::InvoiceInvalid,
-            PaymentError::Routing(_) => Self::RoutingFailed,
-            PaymentError::Sending(_) => Self::RoutingFailed,
-        }
+        MutinyError::from(e).into()
```
TonyGiorgio, line 287: Ah, convert to `MutinyError` and then convert to `MutinyJsError`, yeah that makes sense, nice catch.

TonyGiorgio (approved):

PR #281, opened 2022-12-30
Fix firefox styling

### LINKED ISSUE #258: Manager nav looks like trash on Firefox
from @benalleng 's recent pr:
Screen Shot 2022-12-27 at 11 53 23 AM

here's what it looks like for me on chrome:

Screen Shot 2022-12-27 at 11 54 39 AM

guessing it's due to me using the `:has` selector which isn't supported by ff

Screen Shot 2022-12-27 at 11 56 02 AM

futurepaul:
closes #258 

Also somehow magically in the history of these layout components we no longer need the logic to detect whether it's a scrollable component, so that's a nice little win. Used this opportunity to get rid of all the <ScreenMain> usage and just use <main> because it's just tailwind style + cva


benthecarman (approved):
Looks good to me
image

TonyGiorgio:
Looks good on my firefox. Can't believe I never noticed. I guess the badness just blended in to the 8 bit stuff to my eyes.

TonyGiorgio (approved):

PR #282, opened 2022-12-30
confirm channel close and also await async stuff

futurepaul:
I think this all works but all of a sudden I'm having a ton of problem doing anything on my local setup, connecting to peers, disconnecting, etc.

might blow it away and start over


futurepaul:
closes #268

frontend/src/routes/OpenChannel.tsx
```diff
@@ -58,7 +74,22 @@ export default function OpenChannel() {
...
+									</select>
+								</div>
+							</div>
+						}
+						{(!peers || !peers.length) &&
+							<div className="flex flex-col gap-2">
+								<label>New Peer</label>
+								<input onChange={(e) => setPeerPubkey(e.target.value)} value={peerPubkey} className={`w-full ${inputStyle({ accent: "blue" })}`} type="text" placeholder='Target node pubkey' />
+							</div>
+						}
```
benthecarman, line 92: I think this won't work because we currently don't allow opening a channel to a peer we aren't connected to

futurepaul: oh TIL

TonyGiorgio:
You're probably going to have peer connection problems trying to close a channel after blocking the thread with the dialog box. We should really have our own popup/dialog box that isnt the browser's blocking one. I can't remember the last time a website used that one anyways tbh.

TonyGiorgio:
related https://github.com/BitcoinDevShop/mutiny-web-poc/issues/301

PR #284, opened 2022-12-30
More mutiny to mutiny cleanup

### LINKED ISSUE #262: should revert this back to trace
should revert this back to trace

_Originally posted by @benthecarman in https://github.com/BitcoinDevShop/mutiny-web-poc/pull/207#discussion_r1058663425_

TonyGiorgio:
Handles the rest of the PR feedback in https://github.com/BitcoinDevShop/mutiny-web-poc/pull/144

- Fixes #262

### DIFF
--- a/node-manager/src/node.rs
+++ b/node-manager/src/node.rs
@@ -51,7 +51,7 @@ use lightning::routing::router::DefaultRouter;
 use lightning::routing::scoring::{ProbabilisticScorer, ProbabilisticScoringParameters};
 use lightning::util::config::{ChannelHandshakeConfig, ChannelHandshakeLimits, UserConfig};
 use lightning_invoice::{payment, Invoice};
-use log::{debug, error, info};
+use log::{debug, error, info, trace};
 
 pub(crate) type NetworkGraph = gossip::NetworkGraph<Arc<MutinyLogger>>;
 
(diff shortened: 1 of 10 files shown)


node-manager/src/node.rs
```diff
@@ -108,7 +108,7 @@ impl PubkeyConnectionInfo {
                 .context("connect_peer requires peer connection info")?;
         };
 
-        if connection.starts_with("mutiny:") {
+        if connection.to_lowercase().starts_with("mutiny:") {
             let stripped_connection = connection.strip_prefix("mutiny:").expect("should strip");
```
benthecarman, line 112: I think this strip_prefix will fail if it isn't all lowercase now

node-manager/src/socket.rs
```diff
@@ -224,10 +234,10 @@ impl MultiWsSocketDescriptor {
                             };
                         }
                         Message::Bytes(msg) => {
-                            debug!("received a binary message on multi socket...");
                             // This is a mutiny to mutiny connection with pubkey + bytes
                             // as the binary message. Parse the msg and see which pubkey
                             // it belongs to.
+                            debug!("received a binary message on multi socket...");
```
benthecarman, line 240: should this be trace?

benthecarman:
Looks like CI is throwing some clippy errors now

node-manager/src/node.rs
TonyGiorgio: good catch, doing it at the beginning now so it should apply to both

node-manager/src/socket.rs
TonyGiorgio: moved to trace

benthecarman (approved):

TonyGiorgio:
Tested the vercel deployment and it worked well for me.

PR #285, opened 2022-12-30
Add missing clippy command on mac

TonyGiorgio:
@futurepaul, make sure that the just file works for you. Stuff like clippy and tests. I've changed some of the commands and tried to make sure they work on mac too but Im not sure.


futurepaul (approved):
worked on my machine

PR #297, opened 2023-01-02
Client Proxy Refactor

TonyGiorgio:


TonyGiorgio:
So I did all of this because I thought something in LDK land was thinking that my master socket -> subsocket stuff was equal to other subsockets and causing one socket to clash and disconnect when another one was being disconnected. Turns out the disconnection was due to https://github.com/BitcoinDevShop/mutiny-web-poc/pull/300

I think it's still good to have some mocking, abstraction, and at least some `eq` tests since we have that implemented ourselves anyways. Later we can do more tests with the mocked out connection.. 

Let me think if it's still good to merge @benthecarman or if we should throw it away as not being necessary.

benthecarman (approved):
yeah this looks like a cleaner way to do it

PR #300, opened 2023-01-02
Allow extra time to disconnect all peers

TonyGiorgio:
Whenever disconnecting from one peer, it would seemingly instantly disconnect all peers. Turns out it was from this part of the code. It seems weird that it would trigger almost instantly, not ~1s. Not sure if some weird f64 thing going on or what but doubling it seems to make it work for me.

Fixes https://github.com/BitcoinDevShop/mutiny-web-poc/issues/299

I would be curious if anyone else can reproduce this on their machines / phones and if this PR fixes it for them.

### DIFF
--- a/node-manager/src/background.rs
+++ b/node-manager/src/background.rs
@@ -26,7 +26,7 @@ use lightning::util::events::{Event, EventHandler};
 use lightning::util::logger::Logger;
 use lightning::util::persist::Persister;
 use lightning_rapid_gossip_sync::RapidGossipSync;
-use log::{error, trace};
+use log::{debug, error, trace};
 use std::ops::Deref;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


TonyGiorgio:
The comments in the code is interesting too. I wonder if it would be related to https://github.com/BitcoinDevShop/mutiny-web-poc/issues/272

futurepaul (approved):
I was able to repro the bug and this does fix it, although I am also totally confused by why this fixes it

futurepaul:
#301

TonyGiorgio:
If someone can test the behavior on iOS, that'll help. Based on discussions with Matt, I'm curious if reconnections will work after being backgrounded.

benthecarman (approved):

Issue #150, opened 2022-12-02
publish node_manager as a standalone npm package


futurepaul:
this will help with #136

futurepaul:
I ran into some problems getting testing working today, both in webpack + jest land and in a vite project with vitest. Part of the problem seems to be running these tests in node... so might be a total non-starter getting `node-manager` to test well. But another part of it seems to be to do with module resolution and how we initialize the wasm lib. If we had a really nice easy-to-import js package it would help with #136 but also these kinds of problems.

My new best friend for this process is this guy:
https://nickb.dev/blog/recommendations-when-publishing-a-wasm-library/

Who has a sample library here that shows best practices for making a lib:
https://github.com/nickbabcock/jomini/

(including testing with vitest)

PR #302, opened 2023-01-03
Switch to vite

futurepaul:
Probably blocked on #150 

I wanted to use vite in the first place but I wasn't doing well with importing wasm.

We end up with a lot fewer dev dependencies (react-scripts is a nightmare) and we have a stack that's both easier to customize than CRA and better at staying modern.

WIP still need to get testing really working with our actual app, not just a one-off function.

Also this removes PWA right now so need to get that back in.

Will probably use Vite PWA which even has built-in prompt for refresh: https://vite-pwa-org.netlify.app/guide/prompt-for-update.html


TonyGiorgio:

PR #304, opened 2023-01-03
non-native confirm dialog

futurepaul:
only added it to peer screen so far, gotta add it to the other places

a lot less ergonomic than window.confirm but \o/

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "mw-frontend",
-  "version": "0.1.0",
+  "version": "0.2.0",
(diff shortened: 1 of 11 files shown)


futurepaul:
okay got rid of all the other existing window.confirms

I don't think this will be a silver bullet for the peer disconnects due to the browser blocking... could happen due to backgrounding, or the file upload dialog for instance. but it at least fixes the most common one: messing around on the peers and channels pages

benthecarman (approved):
Looks good to me

TonyGiorgio:
Works for me, nice work

TonyGiorgio (approved):

PR #306, opened 2023-01-04
Persist all peers and make peer manager abstracted

TonyGiorgio:
Gonna need to do this for all the structures being saved onto the `node` so we can test things better on our node implementation. I'll do those in separate PR's just like this if it's okay.

Need to do this in order to test the thing I was changing originally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ debug/
 target/
 ./*/Cargo.lock
 Cargo.lock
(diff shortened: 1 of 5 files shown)


TonyGiorgio:
Fixes https://github.com/BitcoinDevShop/mutiny-web-poc/issues/305

benthecarman (approved):

PR #307, opened 2023-01-04
Update bdk to v0.26.0

benthecarman:
In 0.26 bdk added the transaction sorting we were doing, so now we can just call `.sort`

It also updated the esplora dep so we can now use the same code from https://github.com/lightningdevkit/rust-lightning/pull/1870

### DIFF
--- a/node-manager/Cargo.toml
+++ b/node-manager/Cargo.toml
@@ -22,7 +22,7 @@ js-sys = "0.3.60"
 secp256k1 = "0.24.0"
 bitcoin_hashes = { version = "0.11", default-features = false }
 bitcoin = { version = "0.29.2", features = ["serde"] }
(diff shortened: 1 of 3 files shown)


TonyGiorgio:
will it eventually get to the point where we can use that directly? I guess when that PR merges or can we import that branch until the meantime?

benthecarman:
> will it eventually get to the point where we can use that directly? I guess when that PR merges or can we import that branch until the meantime?

once that's merged and released we can use that, for now copy-paste should be stable enough

TonyGiorgio (approved):
code LGTM, tested this PR locally and did all the on chain sending / receiving and channel opening / sending and all that worked for me too. nice work keeping the deps fresh.

PR #308, opened 2023-01-09
Compress logo png files

benthecarman:
Ran these through https://compresspng.com/, should reduce size by about 50%


futurepaul (approved):
great job

PR #313, opened 2023-01-18
fix select element clickability

### LINKED ISSUE #309: Testnet dropdown button does not react when you click arrow
The arrow next to the testnet dropdown is not clickable, you have to click on the words for the thing to show you the dropdown list. That was the first thing I clicked on and Im sure others will as well. Very minor tho but if we reuse that component anywhere else then it'll be weird.

futurepaul:
closes #309


benthecarman (approved):
Works for me!

TonyGiorgio (approved):
goood job

PR #317, opened 2023-01-31
Handle channel open manually

TonyGiorgio:
This allows others to open a channel to us, useful for LSP's.

A later PR can handle 0 conf LSP stuff.

### DIFF
--- a/node-manager/src/event.rs
+++ b/node-manager/src/event.rs
@@ -326,15 +326,55 @@ impl EventHandler {
                     }
                 }
             }
-            Event::OpenChannelRequest { .. } => {
-                // Unreachable, we don't set manually_accept_inbound_channels
+            Event::OpenChannelRequest {
+                temporary_channel_id,
+                counterparty_node_id,
+                ..
+            } => {
                 self.logger.log(&Record::new(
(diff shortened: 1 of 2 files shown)


node-manager/src/event.rs
```diff
@@ -326,15 +326,43 @@ impl EventHandler {
...
+                    format_args!("EVENT: OpenChannelRequest incoming"),
                     "event",
                     "",
                     0,
                 ));
+
+                match self.channel_manager.accept_inbound_channel(
+                    &temporary_channel_id,
+                    &counterparty_node_id,
+                    0, // we don't track our channels currently
```
benthecarman, line 345: if we accept multiple channels will there be a problem with overlap, should we generate a random number here instead?

benthecarman (approved):
looks good to me otherwise

node-manager/src/event.rs
TonyGiorgio: I haven't ran into it yet but shouldn't be difficult to already do a random number, will add that.

TonyGiorgio: Added random number generation here.

benthecarman (approved):

PR #321, opened 2023-02-04
Basic LNURL support

benthecarman:
Not hooked up to the front end.

This adds basic support for lnurl. The flow would be to call `decode_lnurl` to see what kind of lnurl it is and then call the appropriate function (pay or withdraw) based on that result.

This does result in calling the lnurl twice, otherwise we'd have to duplicate the lnurl structures for wasm bindgen which I didn't want to have to do. However I think this should be okay because normally you'd know the what the lnurl is by the context (on the pay screen already).

### DIFF
--- a/node-manager/src/nodemanager.rs
+++ b/node-manager/src/nodemanager.rs
@@ -23,6 +23,8 @@ use lightning::chain::keysinterface::{KeysInterface, Recipient};
 use lightning::chain::Confirm;
 use lightning::ln::channelmanager::{ChannelDetails, PhantomRouteHints};
 use lightning_invoice::{Invoice, InvoiceDescription};
+use lnurl::lnurl::LnUrl;
+use lnurl::{AsyncClient as LnUrlClient, LnUrlResponse, Response};
 use log::{debug, error, info};
 use reqwest::Client;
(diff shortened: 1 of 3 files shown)


node-manager/src/nodemanager.rs
```diff
@@ -788,6 +809,79 @@ impl NodeManager {
...
+        let lnurl = LnUrl::from_str(&lnurl)?;
+
+        let response = self.lnurl_client.make_request(&lnurl.url).await?;
+
+        match response {
+            LnUrlResponse::LnUrlPayResponse(_) => Err(MutinyJsError::IncorrectLnUrlFunction),
+            LnUrlResponse::LnUrlWithdrawResponse(withdraw) => {
+                let description = withdraw.default_description.clone();
+                let mutiny_invoice = self.create_invoice(Some(amount_sats), description).await?;
+                let invoice_str = mutiny_invoice.bolt11.expect("Invoice should have bolt11");
```
TonyGiorgio, line 872: Probably shouldn't be an `expect` but instead an error.

TonyGiorgio (changes requested):

TonyGiorgio (approved):

PR #328, opened 2023-03-03
Move to LDK 0.0.114

benthecarman:


node-manager/src/node.rs
```diff
@@ -923,12 +938,13 @@ pub(crate) fn create_peer_manager(
 
     PeerManagerImpl::new(
         lightning_msg_handler,
-        km.get_node_secret(Recipient::Node)
-            .expect("Failed to get node secret"),
+        // km.get_node_secret(Recipient::Node)
+        //     .expect("Failed to get node secret"),
```
TonyGiorgio, line 942:
```suggestion
```

TonyGiorgio (commented):
minor nit, going to test now to see how this works in the app

node-manager/src/node.rs
```diff
@@ -288,9 +275,11 @@ impl Node {
...
                     background_processor_logger.clone(),
                     Some(scorer.clone()),
-                    |d| async move {
-                        sleep(d.as_millis() as i32).await;
-                        false
+                    |d| {
+                        Box::pin(async move {
+                            sleep(d.as_millis() as i32).await;
+                            false
+                        })
```
TonyGiorgio, line 282: sometimes i see this error that I don't remember really seeing before: 

```
node_manager.js:1681 panicked at '`async fn` resumed after completion', node-manager/src/node.rs:279:34
```

benthecarman: yeah needed the Box::pin to make it compile, not sure if that was correct or not

TonyGiorgio: this was fixed with https://github.com/MutinyWallet/mutiny-node/pull/328/commits/10f2353fea9cb2071a7f14cec75f75230914bbc1 and LDK is merging it upstream

node-manager/Cargo.toml
```diff
@@ -22,17 +22,21 @@ bitcoin-bech32 = "0.12"
 js-sys = "0.3.60"
 secp256k1 = "0.24.0"
 bitcoin_hashes = { version = "0.11", default-features = false }
-bitcoin = { version = "0.29.2", features = ["serde"] }
-bdk = { version = "0.27.0", default-features = false, features = ["keys-bip39", "esplora", "use-esplora-reqwest", "async-interface"] }
+bitcoin = { version = "0.29.2", default-features = false, features = ["serde", "secp-recovery"] }
+# TODO waiting for esplora version 0.4.0
+bdk = { git = "https://github.com/benthecarman/bdk", branch = "tmp", default-features = false, features = ["keys-bip39", "esplora", "use-esplora-reqwest", "async-interface"] }
```
benthecarman, line 27: we can wait for bdk v0.27.2 if we want, should be any day now

benthecarman (commented):
otherwise lgtm

PR #330, opened 2023-04-03
Basic RGS support

benthecarman:
This adds basic RGS support by having it download the full snapshot on startup and use it across all the nodes. We should later add it so it stores the snapshot for later so we dont need to download the full 4mb every boot.

We also don't have anything for regtest and signet. We should maybe add to our docs how to setup a rgs dev server.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,6 +2,7 @@ use bdk::esplora_client;
 use lightning::ln::peer_handler::PeerHandleError;
 use lightning_invoice::payment::PaymentError;
 use lightning_invoice::ParseOrSemanticError;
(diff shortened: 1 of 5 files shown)


TonyGiorgio (approved):
Nice! Should there be a TODO as well to keep it in sync? or really that's probably not a big deal, there's really not going to be much channel changes before they probably refresh the page again, right?

benthecarman:
> Should there be a TODO as well to keep it in sync? or really that's probably not a big deal, there's really not going to be much channel changes before they probably refresh the page again, right?

Will file an issue

PR #332, opened 2023-04-04
Remove unneeded clone

benthecarman:
Clippy didn't catch this in #330


TonyGiorgio (approved):
Always a cloner, never a taker

PR #333, opened 2023-04-04
Makes the bdk & ldk forks simpler

TonyGiorgio:
I moved our bdk & ldk forks into the MutinyWallet github for easier access to the whole team.

The BDK fork is here: https://github.com/MutinyWallet/bdk/pull/1
It is copied from ben's tmp fork. Mostly dependency updates.

Actually didn't need the LDK fork, so I forced pushed. Since we are using our own `lightning-transaction-sync`, we have no reason for updating esplora.

I have copied the `lightning-transaction-sync` files into our own `esplora.rs` because of some `Send + Sync` issues. It's really just like 5 lines of code changed there but it would have been a fork nevertheless.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,19 +22,18 @@ secp256k1 = "0.24.0"
 bitcoin_hashes = { version = "0.11", default-features = false }
 bitcoin = { version = "0.29.2", default-features = false, features = ["serde", "secp-recovery"] }
 # TODO waiting for esplora version 0.4.0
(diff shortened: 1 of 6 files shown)


TonyGiorgio:
Removed the LDK fork since we don't need it to have esplora yet (removed `lightning-transaction-sync` package for now.

Could probably remove the BDK fork with bdk core upgrade so I didn't bother touching that besides to move it over to our repo.

benthecarman:
> Could probably remove the BDK fork with bdk core upgrade so I didn't bother touching that besides to move it over to our repo.

Looks like bdk 28.0 will allow us to remove the fork. Eventually when bdk core is ready for no-std we can move to that as well.

PR #335, opened 2023-04-04
Save RGS data into indexeddb

benthecarman:
Made a change as well so when creating a node manager in test we don't actually sync gossip, this should make tests faster and not error out incorrectly.

### DIFF
--- a/src/gossip.rs
+++ b/src/gossip.rs
@@ -1,5 +1,5 @@
-use crate::error::{MutinyError, MutinyStorageError};
-use crate::logging::MutinyLogger;
+use std::sync::Arc;
+
 use anyhow::anyhow;
 use bitcoin::Network;
 use bitcoin_hashes::hex::{FromHex, ToHex};
@@ -7,9 +7,10 @@ use lightning::util::ser::{ReadableArgs, Writeable};
 use log::{debug, info, warn};
 use reqwest::Client;
 use rexie::{ObjectStore, Rexie, TransactionMode};
-use std::sync::Arc;
 use wasm_bindgen::JsValue;
 
+use crate::error::{MutinyError, MutinyStorageError};
+use crate::logging::MutinyLogger;
 use crate::node::{NetworkGraph, RapidGossipSync};
 use crate::wallet::get_rgs_url;
 
@@ -212,9 +213,9 @@ pub fn get_dummy_gossip(
 mod test {
     use wasm_bindgen_test::{wasm_bindgen_test as test, wasm_bindgen_test_configure};
 
-    wasm_bindgen_test_configure!(run_in_browser);
-
     use super::*;
+
+    wasm_bindgen_test_configure!(run_in_browser);
 
     #[test]
     async fn test_gossip() {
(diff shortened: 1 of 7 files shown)


src/gossip.rs
```diff
@@ -0,0 +1,235 @@
...
+    let last_sync_timestamp_js = store.get(&JsValue::from(GOSSIP_SYNC_TIME_KEY)).await?;
+    let last_sync_timestamp: u32 =
+        last_sync_timestamp_js
+            .as_f64()
+            .ok_or_else(|| MutinyError::ReadError {
+                source: MutinyStorageError::Other(anyhow!(
+                    "could not read last_sync_timestamp, got {:?}",
+                    last_sync_timestamp_js
+                )),
+            })? as u32;
```
TonyGiorgio, line 65: Should we just default to 0 here to not cause any client errors? Worst case it fetches it down entirely but that's just 4mb or so, not the end of the world.

src/gossip.rs
```diff
@@ -0,0 +1,235 @@
...
+            &JsValue::from(last_sync_timestamp),
+            Some(&JsValue::from(GOSSIP_SYNC_TIME_KEY)),
+        )
+        .await?;
+
+    // Save the network graph
+    let network_graph_js = serde_wasm_bindgen::to_value(&network_graph.encode().to_hex())?;
+    store
+        .put(&network_graph_js, Some(&JsValue::from(NETWORK_GRAPH_KEY)))
+        .await?;
```
TonyGiorgio, line 105: Should we delete the `GOSSIP_SYNC_TIME_KEY` we just saved? Or reset it back to the last value? or does `transaction.done()` make sure that both happen?

src/gossip.rs
```diff
@@ -0,0 +1,236 @@
...
+
+use crate::error::{MutinyError, MutinyStorageError};
+use crate::logging::MutinyLogger;
+use crate::node::{NetworkGraph, RapidGossipSync};
+use crate::wallet::get_rgs_url;
+
+pub(crate) const GOSSIP_DATABASE_NAME: &str = "gossip";
+pub(crate) const GOSSIP_OBJECT_STORE_NAME: &str = "gossip_store";
+pub(crate) const GOSSIP_SYNC_TIME_KEY: &str = "last_sync_timestamp";
+pub(crate) const NETWORK_GRAPH_KEY: &str = "network_graph";
```
TonyGiorgio, line 20: How does this compare with `ldkstorage.rs` with the `NETWORK_KEY` value there?

benthecarman: Ah yeah I guess I should delete that now because we aren't using LocalStorage for the network graph now

src/gossip.rs
benthecarman: On line 123 we do this, if this function errors we just use a fresh Network Graph and download the whole thing

src/gossip.rs
benthecarman: The transactions are supposed to be atomic so that shouldn't be a concern

PR #337, opened 2023-04-04
Voltage LSP integration

TonyGiorgio:
Connects to LSP, if one is configured, gets the pubkey, adds to whitelist, checks when channel is opened to see if 0 conf or not, and auto generates wrapped invoice if lsp is configured.

### DIFF
--- a/src/lspclient.rs
+++ b/src/lspclient.rs
@@ -1,0 +1,122 @@
+use std::str::FromStr;
+
+use bitcoin::secp256k1::PublicKey;
+use reqwest::Client;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug)]
+pub(crate) struct LspClient {
+    pub pubkey: PublicKey,
+    pub connection_string: String,
+    pub url: String,
+    pub http_client: Client,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub(crate) struct GetInfoResponse {
+    pub pubkey: String,
+    pub connection_methods: Vec<GetInfoAddress>,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub(crate) struct GetInfoAddress {
+    #[serde(rename = "type")]
+    pub item_type: GetInfoAddressType,
+    pub port: u16,
+    pub address: String,
+}
+
+/// Type of connection
+#[derive(Copy, Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub(crate) enum GetInfoAddressType {
+    DNS,
+    IPV4,
+    IPV6,
+    TORV2,
+    TORV3,
+    WEBSOCKET,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct ProposalRequest {
+    pub bolt11: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub host: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub port: Option<u16>,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct ProposalResponse {
+    pub jit_bolt11: String,
+}
+
+const GET_INFO_PATH: &str = "/api/v1/info";
+const PROPOSAL_PATH: &str = "/api/v1/proposal";
+
+impl LspClient {
+    pub async fn new(url: String) -> anyhow::Result<Self> {
+        let http_client = Client::new();
+        let get_info_response: GetInfoResponse = http_client
+            .get(format!("{}{}", &url, GET_INFO_PATH))
+            .send()
+            .await?
+            .json()
+            .await?;
+
+        let pubkey = PublicKey::from_str(&get_info_response.pubkey)?;
+        let connection_string = get_info_response
+            .connection_methods
+            .iter()
+            .filter(|address| {
+                matches!(
+                    address.item_type,
+                    GetInfoAddressType::IPV4 | GetInfoAddressType::IPV6 | GetInfoAddressType::TORV3
+                )
+            })
+            .min_by_key(|address| match address.item_type {
+                // Prioritize IPV4, then 6, then tor
+                // TODO support websocket one day
+                GetInfoAddressType::IPV4 => 0,
+                GetInfoAddressType::IPV6 => 1,
+                GetInfoAddressType::TORV3 => 2,
+                _ => unreachable!(),
+            })
+            .map(|address| {
+                format!(
+                    "{}@{}:{}",
+                    pubkey.to_string(),
+                    address.address,
+                    address.port
+                )
+            })
+            .ok_or_else(|| anyhow::anyhow!("No suitable connection method found"))?;
+
+        Ok(LspClient {
+            pubkey,
+            url,
+            connection_string,
+            http_client,
+        })
+    }
+
+    pub(crate) async fn get_lsp_invoice(&self, bolt11: String) -> anyhow::Result<String> {
+        let payload = ProposalRequest {
+            bolt11,
+            host: None,
+            port: None,
+        };
+
+        let proposal_response: ProposalResponse = self
+            .http_client
+            .post(format!("{}{}", &self.url, PROPOSAL_PATH))
+            .json(&payload)
+            .send()
+            .await?
+            .json()
+            .await?;
+
+        Ok(proposal_response.jit_bolt11)
+    }
+}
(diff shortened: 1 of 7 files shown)


src/lspclient.rs
```diff
@@ -0,0 +1,12 @@
+use bip32::secp256k1::PublicKey;
```
benthecarman, line 1: might make sense to have this as a separate library

TonyGiorgio: damn nice catch that is not supposed to be in there. Code imports messed up. Tired of every single library redefining their own PublicKey.

benthecarman: oh lol, didn't mean to catch that :joy: 

meant that the lsp http client could be a separate library

TonyGiorgio: Oh yeah totally. I thought about that as I created the file. I'm unsure of how it'll end up as time goes on. Could have a voltage specific one, which is what it'll be originally, but what if we want to change it to be our own type or more along the lines of the LSP spec. Would rather wait to see how it turns out or until someone asks.

src/lspclient.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub connection_methods: Vec<GetinfoAddress>,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub(crate) struct GetinfoAddress {
+    #[serde(rename = "type")]
+    pub item_type: GetinfoAddressType,
+    #[serde(alias = "port")]
+    pub port: u16,
+    #[serde(alias = "address", skip_serializing_if = "Option::is_none")]
```
benthecarman, line 25: these dont need aliases

src/lspclient.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[serde(rename = "ipv4")]
+    IPV4,
+    #[serde(rename = "ipv6")]
+    IPV6,
+    #[serde(rename = "torv2")]
+    TORV2,
+    #[serde(rename = "torv3")]
+    TORV3,
+    #[serde(rename = "websocket")]
+    WEBSOCKET,
```
benthecarman, line 43: i think there is a to_lowercase thing you can do for the whole enum rather than individually to each item

src/node.rs
```diff
@@ -233,13 +240,29 @@ impl Node {
...
         };
 
+        let lsp_client_pubkey = lsp_client.clone().map(|lsp| lsp.pubkey);
+
+        // save connection info peer connection list to auto connect
+        if let Some(lsp) = lsp_client.clone() {
+            match persister.persist_peer_connection_info(
+                lsp_client_pubkey.unwrap().to_string(),
+                lsp.connection_string.clone(),
+            ) {
```
benthecarman, line 250: can probably make `persist_peer_connection_info` take `&str`s so we don't need to do all these clones

src/nodemanager.rs
```diff
@@ -387,6 +390,19 @@ impl NodeManager {
 
         let fee_estimator = Arc::new(MutinyFeeEstimator::default());
 
+        // load lsp client, if any
+        let lsp_client: Option<Arc<LspClient>> = match lsp_url.clone() {
+            Some(lsp_url_ref) if !lsp_url_ref.is_empty() => {
```
benthecarman, line 395: i think you're handling an empty string here? if so should add a comment

src/lspclient.rs
```diff
@@ -0,0 +1,84 @@
...
+                // Prioritize IPV4, then 6, then tor
+                GetinfoAddressType::IPV4 => 0,
+                GetinfoAddressType::IPV6 => 1,
+                GetinfoAddressType::TORV3 => 2,
+                _ => unreachable!(),
+            })
+            .map(|address| {
+                format!(
+                    "{}:{}",
+                    address.address.as_ref().unwrap_or(&url),
```
benthecarman, line 72: this `.unwrap_or(&url)` seems dangerous, I think it'd be better to only get the connection info from the LSP rather than sometimes replacing it

src/lspclient.rs
```diff
@@ -0,0 +1,84 @@
...
+                matches!(
+                    address.item_type,
+                    GetinfoAddressType::IPV4 | GetinfoAddressType::IPV6 | GetinfoAddressType::TORV3
+                )
+            })
+            .min_by_key(|address| match address.item_type {
+                // Prioritize IPV4, then 6, then tor
+                GetinfoAddressType::IPV4 => 0,
+                GetinfoAddressType::IPV6 => 1,
+                GetinfoAddressType::TORV3 => 2,
```
benthecarman, line 66: maybe add a TODO to one day support websocket

src/lspclient.rs
```diff
@@ -0,0 +1,126 @@
...
+pub struct ProposalResponse {
+    pub jit_bolt11: String,
+}
+
+const GET_INFO_PATH: &str = "/api/v1/info";
+const PROPOSAL_PATH: &str = "/api/v1/proposal";
+
+impl LspClient {
+    pub async fn new(url: String) -> anyhow::Result<Self> {
+        let http_client = Client::builder().build()?;
```
benthecarman, line 65: you could add the http_client to `LspClient` that way we dont need to reinitialize it every call

src/lspclient.rs
```diff
@@ -0,0 +1,126 @@
+use std::str::FromStr;
+
+use bitcoin::secp256k1::PublicKey;
+use reqwest::Client;
+use serde::{Deserialize, Serialize};
+
+pub(crate) struct LspClient {
```
benthecarman, line 7: should have this dervice clone and debug

src/node.rs
```diff
@@ -149,6 +154,7 @@ pub(crate) struct Node {
     logger: Arc<MutinyLogger>,
     websocket_proxy_addr: String,
     multi_socket: MultiWsSocketDescriptor,
+    lsp_client: Option<Arc<LspClient>>,
```
benthecarman, line 157: if we have LspClient derive clone it shouldnt need to be in an arc

benthecarman (changes requested):

src/node.rs
```diff
@@ -598,7 +622,12 @@ impl Node {
             0,
         ));
 
-        Ok(invoice)
+        if let Some(lsp) = self.lsp_client.clone() {
+            let invoice = lsp.get_lsp_invoice(invoice.to_string()).await?;
+            Ok(Invoice::from_str(&invoice)?)
```
benthecarman, line 627: we might want to make sure that we are connected to the LSP, otherwise the LSP might be down.

Also if the `get_lsp_invoice` fails, we would just fail here, do you think that is okay? In some cases I can see that being  correct (if we have no channels) in others that would kinda suck (we have enough inbound)

TonyGiorgio: Fixed the first one by doing a connect_peer first, but it should at least be trying to auto connect when I added it to the node initialization.

> Also if the get_lsp_invoice fails, we would just fail here, do you think that is okay?

Yeah I thought about that. If they have an LSP configured, then I would say we should fail it simply because the LSP also helps with privacy. Maybe once we get route blinding it won't matter much and we can do the "channel or not?" check?

src/node.rs
```diff
@@ -209,13 +216,29 @@ impl Node {
...
+        if let Some(lsp) = lsp_client.clone() {
+            match persister.persist_peer_connection_info(
+                lsp_client_pubkey.unwrap().to_string(),
+                &lsp.connection_string,
+            ) {
+                Ok(_) => (),
+                Err(_) => {
+                    error!("could not save connection to lsp");
+                }
+            }
```
benthecarman, line 231: nit: could do `if let Err(_) = ` here because we ignore the `Ok` case

src/lspclient.rs
```diff
@@ -0,0 +1,118 @@
...
+            bolt11,
+            host: None,
+            port: None,
+        };
+
+        let proposal_response: ProposalResponse = self
+            .http_client
+            .post(format!("{}{}", &self.url, PROPOSAL_PATH))
+            .header("Content-Type", "application/json")
+            .json(&payload)
```
benthecarman, line 110: nit: i assume calling `.json` here adds the header

src/lspclient.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub item_type: GetinfoAddressType,
+    pub port: u16,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub address: Option<String>,
+}
+
+/// Type of connection
+#[derive(Copy, Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub(crate) enum GetinfoAddressType {
```
benthecarman, line 33: nit: `GetInfoAddressType`

benthecarman (approved):

PR #338, opened 2023-04-04
Better handle if no gossip data available

benthecarman:
Figured out how to handle if data isn't available in indexeddb so we can return an `Option`

### DIFF
--- a/src/gossip.rs
+++ b/src/gossip.rs
@@ -4,7 +4,7 @@ use anyhow::anyhow;
 use bitcoin::Network;
 use bitcoin_hashes::hex::{FromHex, ToHex};
 use lightning::util::ser::{ReadableArgs, Writeable};
-use log::{debug, info, warn};
+use log::{debug, error, info, warn};
 use reqwest::Client;
 use rexie::{ObjectStore, Rexie, TransactionMode};
 use wasm_bindgen::JsValue;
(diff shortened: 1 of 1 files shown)


src/gossip.rs
```diff
@@ -119,8 +133,12 @@ pub async fn get_gossip_sync(
...
     // if we error out, we just use the default gossip data
     let gossip_data = match get_gossip_data(&rexie, logger.clone()).await {
-        Ok(gossip_data) => gossip_data,
-        Err(_) => Gossip::new(network, logger.clone()),
+        Ok(Some(gossip_data)) => gossip_data,
+        Ok(None) => Gossip::new(network, logger.clone()),
+        Err(e) => {
+            error!("Error getting gossip data from storage: {e}, re-syncing gossip...");
+            Gossip::new(network, logger.clone())
+        }
```
TonyGiorgio, line 141: You could use `unwrap_or_else` twice to get rid of the multiple Ok/Some/None parts. 

```
let gossip_data = get_gossip_data(&rexie, logger.clone())
    .await
    .unwrap_or_else(|e| {
        error!("Error getting gossip data from storage: {e}, re-syncing gossip...");
        None
    })
    .unwrap_or_else(|| Gossip::new(network, logger.clone()));
```

TonyGiorgio (approved):
nit, take it or leave it

src/gossip.rs
benthecarman: yeah I considered that but I think the match is cleaner and more clear on what is happening in each case

PR #339, opened 2023-04-05
Add custom Persister for ProbabilisticScorer and NetworkGraph

### LINKED ISSUE #340: Prune network graph occasionally
We get this log from the background processor: `Not pruning network graph, consider enabling `std` or doing so manually with remove_stale_channels_and_tracking_with_time.`

We should manually call `remove_stale_channels_and_tracking_with_time` from time to time.

benthecarman:
Refactors so we no longer use KVStorePersister and instead use our own impl. This impl currently uses indexedDB for just the `NetworkGraph` and `ProbabilisticScorer`, otherwise is just a copy-paste of the `KVStorePersister` for the channel manager and updates so they are still in LocalStorage.

Closes #340

### DIFF
--- a/src/gossip.rs
+++ b/src/gossip.rs
@@ -261,6 +261,11 @@ pub async fn get_gossip_sync(
 
     let now = utils::now().as_secs();
 
+    // remove stale channels
+    gossip_sync
+        .network_graph()
+        .remove_stale_channels_and_tracking_with_time(now);
+
     // if the last sync was less than 24 hours ago, we don't need to sync
     let time_since_sync = now - gossip_data.last_sync_timestamp as u64;
     if time_since_sync < 86_400 {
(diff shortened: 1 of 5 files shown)


TonyGiorgio (approved):
Looks good. So essentially gossip is going to hold all those related things like the scorer and network graph too? Pretty much a side effect of the persister intertwining all those as well? wish that didn't work like that.

benthecarman:
> So essentially gossip is going to hold all those related things like the scorer and network graph too? Pretty much a side effect of the persister intertwining all those as well? wish that didn't work like that.

Yeah.. I tried to make it so the persister just calls functions from gossip, but not much else we can do. Maybe could make sense to move all of gossip into ldkstorage because it is just all storage stuff.

TonyGiorgio:
> > So essentially gossip is going to hold all those related things like the scorer and network graph too? Pretty much a side effect of the persister intertwining all those as well? wish that didn't work like that.
> 
> Yeah.. I tried to make it so the persister just calls functions from gossip, but not much else we can do. Maybe could make sense to move all of gossip into ldkstorage because it is just all storage stuff.

Not a bad idea, idk if it's necessary or if it's reasonable to put some things back in ldkstorage but I do like that you were able to make them associative functions, about the best we can do.

PR #346, opened 2023-04-09
Properly handle checking signet address

benthecarman:
Signet is stupid and has the same address prefix as testnet, this leads to us parsing signet addresses as having a testnet network. This PR changes it so we allow testnet addresses on signet and vice versa because they are the same.

I also added the network to the `IncorrectNetwork` error type so it'd be easier to debug

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -40,6 +40,16 @@ pub fn currency_from_network(network: Network) -> Currency {
     }
 }
 
+pub fn network_from_currency(currency: Currency) -> Network {
+    match currency {
+        Currency::Bitcoin => Network::Bitcoin,
+        Currency::BitcoinTestnet => Network::Testnet,
+        Currency::Signet => Network::Signet,
+        Currency::Regtest => Network::Regtest,
+        Currency::Simnet => Network::Regtest,
+    }
+}
+
 pub fn now() -> Duration {
     SystemTime::now()
         .duration_since(SystemTime::UNIX_EPOCH)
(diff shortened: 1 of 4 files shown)


src/utils.rs
```diff
@@ -40,6 +40,16 @@ pub fn currency_from_network(network: Network) -> Currency {
     }
 }
 
+pub fn network_from_currency(currency: Currency) -> Network {
+    match currency {
+        Currency::Bitcoin => Network::Bitcoin,
+        Currency::BitcoinTestnet => Network::Testnet,
+        Currency::Signet => Network::Signet,
+        Currency::Regtest => Network::Regtest,
+        Currency::Simnet => Network::Regtest,
```
TonyGiorgio, line 49: is `currency_from_network` not going to break something if there's no `Regtest` to map to `Simnet` or is that the reason for `is_valid_network` everywhere?

src/utils.rs
```diff
@@ -109,3 +119,19 @@ impl<'a, S: Writeable> Writeable for MutexGuard<'a, S> {
         S::write(&**self, writer)
     }
 }
+
+/// Returns true if the given network is valid for the given destination network.
+/// This is used to prevent users from sending funds to the wrong network.
+/// We can't just compare the network directly because signet and testnet
+/// have conflicting address prefixes.
+pub(crate) fn is_valid_network(my_network: Network, dest_network: Network) -> bool {
```
TonyGiorgio, line 127: Is it worth considering defining `eq` / `not eq` for these networks? Then we wouldn't have to change all the `==` / `!=` anywhere.

src/utils.rs
benthecarman: Simnet is just btcd's version of regtest, we don't really need to worry about it.

src/utils.rs
benthecarman: I think this is probably a better approach, there may be scenarios where we want to check testnet != signet

TonyGiorgio: Gotcha. Plus idk if we can easily since we don't own those traits. Hate that rust rule.

TonyGiorgio (approved):

PR #347, opened 2023-04-09
Change default links to mutinynet

benthecarman:
RGS isn't setup yet but making PR for now

### DIFF
--- a/src/wallet.rs
+++ b/src/wallet.rs
@@ -250,7 +250,7 @@ pub(crate) fn get_esplora_url(network: Network, user_provided_url: Option<String
         match network {
             Network::Bitcoin => "https://blockstream.info/api",
             Network::Testnet => "https://blockstream.info/testnet/api",
-            Network::Signet => "https://mempool.space/signet/api",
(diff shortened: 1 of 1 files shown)


src/wallet.rs
```diff
@@ -236,7 +236,7 @@ pub(crate) fn get_esplora_url(network: Network, user_provided_url: Option<String
         match network {
             Network::Bitcoin => "https://blockstream.info/api",
             Network::Testnet => "https://blockstream.info/testnet/api",
-            Network::Signet => "https://mempool.space/signet/api",
+            Network::Signet => "https://mutinynet.com/api",
```
TonyGiorgio, line 239: Starting to get into config hell but hopefully once everything is switched over to signet we can not worry about testnet as much.

TonyGiorgio (approved):
I believe there's a link that generates on the frontend to view the transaction after it is made. If you feel like making the extra step, might want to look for that and do it for signet.

benthecarman:
@TonyGiorgio already done 

https://github.com/MutinyWallet/mutiny-web-poc/pull/36

PR #348, opened 2023-04-09
Verfiy LSP is honest

benthecarman:
We definitely need to verify that the payment hash and network is correct. I also made it verify that node id is correct, probably a good sanity check


TonyGiorgio (approved):
Oh yeah totally, nice catch. I think @futurepaul was validing some of that on the frontend of the voltage flow site. I think he was also validating amount and CLTV delta too but I don't think those matter much. Only thing I would say might matter is to make sure the LSP isn't charging too much of a fee but I don't think I incorporated the fee check endpoint yet.

PR #353, opened 2023-04-11
Allow LSP client to fail on startup

benthecarman:


src/nodemanager.rs
```diff
@@ -398,12 +398,8 @@ impl NodeManager {
...
             // check if string is some and not an empty string
             Some(lsp_url_ref) if !lsp_url_ref.is_empty() => {
-                // TODO what to do if the getinfo call fails?
-                // is erroring out completely okay?
-                let lsp_client = LspClient::new(lsp_url_ref)
-                    .await
-                    .map_err(|_| MutinyError::LspFailure)?;
-                Some(lsp_client)
+                // if it fails, return None so the user can continue without lsp
+                LspClient::new(lsp_url_ref).await.ok()
```
TonyGiorgio, line 402: Still not a fan of this but I get that the node manager should still operate for on chain usage. Something should disable the lightning parts of it though?

Or leave a TODO in the comments and we can figure it out later.

TonyGiorgio (changes requested):

benthecarman:
Decided to do a different strategy

PR #355, opened 2023-04-11
Proxy improvements

TonyGiorgio:
Some better logging and fix disconnect issue.

TODO:
~~Also going to try to handle what happens when we open a new tab~~


TonyGiorgio:
Actually I'll do the rest in a seperate PR, could get a bit more complicated

PR #357, opened 2023-04-11
Change LSP clients to be a Vector, save lsp client per node

benthecarman:


src/node.rs
```diff
@@ -57,6 +57,7 @@ use lightning_invoice::{
     Invoice,
 };
 use log::{debug, error, info, trace};
+use secp256k1::rand;
```
TonyGiorgio, line 60: I don't think we can use this? `getrandom` is what we do for entropy, is this working? Should probably have a test that covers initing a `node` with a vec of LSPs.

src/nodemanager.rs
```diff
@@ -393,18 +395,28 @@ impl NodeManager {
...
             // check if string is some and not an empty string
-            Some(lsp_url_ref) if !lsp_url_ref.is_empty() => {
-                // TODO what to do if the getinfo call fails?
-                // is erroring out completely okay?
-                let lsp_client = LspClient::new(lsp_url_ref)
-                    .await
-                    .map_err(|_| MutinyError::LspFailure)?;
-                Some(lsp_client)
+            Some(lsp_urls) if !lsp_urls.is_empty() => {
+                let urls: Vec<&str> = lsp_urls.split(',').collect();
```
TonyGiorgio, line 402: why not a vec of URLs?

src/nodemanager.rs
```diff
@@ -1175,8 +1203,22 @@ pub(crate) async fn create_new_node_from_node_manager(
 
     // Create and save a new node using the next child index
     let next_node_uuid = Uuid::new_v4().to_string();
+
+    // todo generate random lsp client instead of using first
```
TonyGiorgio, line 1207: Is this `todo` here necessary? It looks like you are doing random.

src/node.rs
benthecarman: Yeah it worked for me, it uses getrandom for randomness if you have the dep

src/nodemanager.rs
benthecarman: because `LspClient` takes a string

TonyGiorgio: Would be cleaner if `lsp_urls` was `Vec<String>` or &str instead. This `split` happens before turning into `LspClient` anyways.

src/nodemanager.rs
```diff
@@ -393,18 +395,28 @@ impl NodeManager {
...
-                let lsp_client = LspClient::new(lsp_url_ref)
-                    .await
-                    .map_err(|_| MutinyError::LspFailure)?;
-                Some(lsp_client)
+            Some(lsp_urls) if !lsp_urls.is_empty() => {
+                let urls: Vec<&str> = lsp_urls.split(',').collect();
+
+                let futs = urls.into_iter().map(|url| LspClient::new(url.trim()));
+
+                let results = futures::future::join_all(futs).await;
```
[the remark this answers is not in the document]
benthecarman: Yes, also if a node has one saved and it fails, then that node will fail to start up.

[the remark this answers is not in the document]
TonyGiorgio: Cool, just wondering. Probably as good of an approach as any to do all the checks at first. On one hand, removes redundant checks and failures at node time, on other hand feels a bit inefficient doing checks for, let's say 5 LSPs (which will call their endpoint to get info) when just one node and one LSP might be used (or already set). Can punt for now if we want to make it more efficient later.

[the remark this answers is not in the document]
benthecarman: Yeah it's really where we want the efficiency. If we checked just in time and had 10 nodes with 1 LSP, we'd be setting up the same LSP 10 times.

TonyGiorgio (approved):

PR #358, opened 2023-04-13
Wasm Bindgen Split

TonyGiorgio:
Moves core back into own sub package

TODO:
- [x] Add another package that is just wasm-bindgen wrapper
- [x] Remove wasm-bindgen wrapper from mutiny-core
- [x] CICD library packages for ~~both the rust library and~~ the typescript library

Follow up:
- Complete wasm strip out will happen in a follow up PR
- CICD for rust library package


benthecarman:
there is a .cargo folder that forces wasm, should move that into the mutiny-core folder

mutiny-core/Cargo.toml
```diff
@@ -0,0 +1,77 @@
...
+serde-wasm-bindgen = "0.5.0"
+bip39 = { version = "1.0.1" }
+bip32 = "0.4.0"
+bitcoin-bech32 = "0.12"
+js-sys = "0.3.60"
+secp256k1 = "0.24.0"
+bitcoin_hashes = { version = "0.11", default-features = false }
+bitcoin = { version = "0.29.2", default-features = false, features = ["serde", "secp-recovery"] }
+# TODO waiting for esplora version 0.4.0
+bdk = { git = "https://github.com/mutinywallet/bdk", branch = "esplora-tmp", default-features = false, features = ["keys-bip39", "esplora", "use-esplora-reqwest", "async-interface"] }
```
benthecarman, line 28: yesterday bdk released 0.28 so we shouldn't need the fork, should be fine to do that here

TonyGiorgio:
> there is a .cargo folder that forces wasm, should move that into the mutiny-core folder

It doesn't work when I do it like that, same `mio` problems pop back up.

mutiny-core/Cargo.toml
TonyGiorgio: I'll leave for a seperate PR, I don't want to break anything

TonyGiorgio:
> At present, when being invoked from a workspace, Cargo does not read config files from crates within the workspace. i.e. if a workspace has two crates in it, named /projects/foo/bar/baz/mylib and /projects/foo/bar/baz/mybin, and there are Cargo configs at /projects/foo/bar/baz/mylib/.cargo/config.toml and /projects/foo/bar/baz/mybin/.cargo/config.toml, Cargo does not read those configuration files if it is invoked from the workspace root (/projects/foo/bar/baz/).

https://doc.rust-lang.org/cargo/reference/config.html

mutiny-core/src/nodemanager.rs
benthecarman, line 1: At this point might make sense to move all of this file into `lib.rs`, what do you think

TonyGiorgio: Yeah I can see that

mutiny-core/Cargo.toml
```diff
@@ -0,0 +1,77 @@
...
+
+[package]
+name = "mutiny-core"
+version = "0.2.4"
+edition = "2021"
+authors = ["Tony Giorgio <tony@mutinywallet.com>", "benthecarman <ben@mutinywallet.com>"]
+# forced-target = "wasm32-unknown-unknown"
+
+#[lib]
+#crate-type = ["cdylib"]
```
benthecarman, line 11: think we can get rid of these

mutiny-core/src/lib.rs
```diff
@@ -28,23 +28,13 @@ mod localstorage;
...
 mod node;
-mod nodemanager;
+pub mod nodemanager;
 mod peermanager;
 mod proxy;
 mod socket;
 mod utils;
 mod wallet;
 
-use log::{debug, Level};
```
benthecarman, line 38: are we not gonna move nodemanager into here?

mutiny-core/src/nodemanager.rs
```diff
@@ -71,33 +71,26 @@ pub(crate) struct NodeIndex {
 
 // This is the NodeIdentity that refer to a specific node
 // Used for public facing identification.
-#[wasm_bindgen]
 pub struct NodeIdentity {
```
benthecarman, line 74: for all these types we could change them to be more type-safe and instead of using String use the actual types

mutiny-wasm/Cargo.toml
```diff
@@ -0,0 +1,52 @@
+cargo-features = ["per-package-target"]
+
+[package]
+name = "mutiny-wasm"
+version = "0.2.4"
+edition = "2021"
+authors = ["Tony Giorgio <tony@mutinywallet.com>", "benthecarman <ben@mutinywallet.com>"]
+# forced-target = "wasm32-unknown-unknown"
```
benthecarman, line 8: i think we want this

mutiny-core/src/nodemanager.rs
benthecarman: same with all the functions, we do things like `invoice_str: String` but could take in an `invoice: Invoice`.

benthecarman: Might be a lot so we could do this in a follow up

benthecarman:
Overall looks good to me, mostly stuff that can be done in a follow up

mutiny-core/src/lib.rs
TonyGiorgio: I dont think that core's node manager needs to be in `lib.rs`. I thought it was a good suggestion for `mutiny-wasm` since it's essentially a wrapper, but `core/nodemanager.rs` is more complex and could use it's own file.

mutiny-core/src/nodemanager.rs
TonyGiorgio: I did change a couple `String` that was a `TxId`. A follow up to identify more of them and be type safe everyone would be nice.

TonyGiorgio: https://github.com/MutinyWallet/mutiny-node/issues/360

mutiny-core/Cargo.toml
TonyGiorgio: Good catch. I left `forced-target` on for now since we still need wasm for core.

TonyGiorgio:
Implemented a couple of the suggested changes and created a follow up issue for one of them. Dont want to do the nodemanager lib change so if that's fine, all things should be resolved.

benthecarman (approved):

PR #359, opened 2023-04-13
Save LN peer metadata to indexedDB

benthecarman:
I also moved the peer connection info to indexedDB.

TODO:

- [x] Add tests
- [x] Test to make sure this works myself
- [x] bring back per-node connections

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -30,6 +30,7 @@ use lightning::chain::chaininterface::{BroadcasterInterface, ConfirmationTarget,
 use lightning::chain::keysinterface::{NodeSigner, Recipient};
 use lightning::chain::Confirm;
 use lightning::ln::channelmanager::{ChannelDetails, PhantomRouteHints};
+use lightning::routing::gossip::NodeId;
 use lightning_invoice::{Invoice, InvoiceDescription};
 use lnurl::lnurl::LnUrl;
 use lnurl::{AsyncClient as LnUrlClient, LnUrlResponse, Response};
@@ -156,7 +157,10 @@ impl From<Invoice> for MutinyInvoice {
 #[derive(Serialize, Deserialize, Clone, Eq, PartialEq)]
 pub struct MutinyPeer {
     pub pubkey: secp256k1::PublicKey,
-    pub connection_string: String,
+    pub connection_string: Option<String>,
+    pub alias: Option<String>,
+    pub color: Option<String>,
+    pub label: Option<String>,
     pub is_connected: bool,
 }
 
@@ -170,6 +174,7 @@ impl Ord for MutinyPeer {
     fn cmp(&self, other: &Self) -> Ordering {
         self.is_connected
             .cmp(&other.is_connected)
+            .then_with(|| self.alias.cmp(&other.alias))
             .then_with(|| self.pubkey.cmp(&other.pubkey))
             .then_with(|| self.connection_string.cmp(&other.connection_string))
     }
@@ -621,10 +626,12 @@ impl NodeManager {
         &self,
         self_node_pubkey: String,
         connection_string: String,
+        label: Option<String>,
     ) -> Result<(), MutinyError> {
         if let Some(node) = self.nodes.lock().await.get(&self_node_pubkey) {
             let connect_info = PubkeyConnectionInfo::new(connection_string.clone())?;
-            let res = node.connect_peer(connect_info).await;
+            let label_opt = label.filter(|s| !s.is_empty()); // filter out empty strings
+            let res = node.connect_peer(connect_info, label_opt).await;
             match res {
                 Ok(_) => {
                     info!("connected to peer: {connection_string}");
@@ -664,13 +671,26 @@ impl NodeManager {
         self_node_pubkey: String,
         peer: String,
     ) -> Result<(), MutinyError> {
+        let pubkey = PublicKey::from_str(&peer).expect("invalid pubkey");
+        let node_id = NodeId::from_pubkey(&pubkey.inner);
+
         if let Some(node) = self.nodes.lock().await.get(&self_node_pubkey) {
-            node.persister.delete_peer_connection_info(peer);
+            gossip::delete_peer_info(&node._uuid, &node_id).await?;
             Ok(())
         } else {
             error!("could not find internal node {self_node_pubkey}");
             Err(MutinyError::WalletOperationFailed.into())
         }
+    }
+
+    pub async fn label_peer(
+        &self,
+        peer: secp256k1::PublicKey,
+        label: Option<String>,
+    ) -> Result<(), MutinyError> {
+        let node_id = NodeId::from_pubkey(&peer);
+        gossip::set_peer_label(&node_id, label).await?;
+        Ok(())
     }
 
     // all values in sats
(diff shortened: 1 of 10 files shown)


src/gossip.rs
```diff
@@ -331,6 +339,184 @@ async fn fetch_updated_gossip(
...
+pub(crate) async fn delete_peer_info(node_id: &NodeId) -> Result<(), MutinyError> {
+    let rexie = build_gossip_database().await?;
+    // Create a new read-write transaction
+    let transaction =
+        rexie.transaction(&[LN_PEER_METADATA_STORE_NAME], TransactionMode::ReadWrite)?;
+    let store = transaction.store(LN_PEER_METADATA_STORE_NAME)?;
+
+    let key = JsValue::from(node_id.to_string());
+
+    store.put(&JsValue::UNDEFINED, Some(&key)).await?;
```
TonyGiorgio, line 483: Why not a proper `delete`?

src/ldkstorage.rs
```diff
@@ -242,31 +242,6 @@ impl MutinyNodePersister {
...
-    }
-
-    // FIXME: Useful to keep around until we use it
-    #[allow(dead_code)]
-    pub(crate) fn delete_peer_connection_info(&self, peer_pubkey: String) {
-        let key = self.get_key(peer_key(peer_pubkey).as_str());
-        MutinyBrowserStorage::delete(key)
-    }
-
     pub(crate) fn list_peer_connection_info(&self) -> Vec<(PublicKey, String)> {
```
TonyGiorgio, line 245: Why is this one still useful?

src/node.rs
```diff
@@ -1,3 +1,4 @@
+use crate::gossip::{read_peer_info, save_peer_connection_info};
```
TonyGiorgio, line 1: Nit: drop into the `crate` section below

src/node.rs
```diff
@@ -25,6 +26,7 @@ use bip39::Mnemonic;
 use bitcoin::hashes::sha256::Hash as Sha256;
 use bitcoin::{hashes::Hash, secp256k1::PublicKey, Network};
 use bitcoin_hashes::hex::ToHex;
+use lightning::routing::gossip::NodeId;
```
TonyGiorgio, line 29: nit: merge here as well

src/ldkstorage.rs
benthecarman: woops, totally missed that

src/gossip.rs
```diff
@@ -331,6 +339,184 @@ async fn fetch_updated_gossip(
...
+            timestamp: primary.timestamp.or(secondary.timestamp),
+        }
+    }
+}
+
+impl From<NodeAnnouncement> for LnPeerMetadata {
+    fn from(value: NodeAnnouncement) -> Self {
+        let alias = NodeAlias(value.contents.alias).to_string();
+        Self {
+            connection_string: None,
```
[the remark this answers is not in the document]
benthecarman: Tried to do this but basically ran into this: https://github.com/lightningdevkit/rust-lightning/issues/2183

Probably fine punting on it for now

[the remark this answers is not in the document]
TonyGiorgio: yeah def fine to skip it for now. Not sure if any other node out there auto does that either.

src/node.rs
benthecarman: Yeah seems our IDEs handle imports differently, yours seems like it does it differently than most other projects. Will change the imports for now but we should try to configure our stuff to work the same (or at least get rustfmt to do it)

TonyGiorgio: Yeah my IDE doesn't do it in one big one either, I do that manually. Perhaps there is a `rustfmt` that does it properly. Or maybe I'm wrong.

src/gossip.rs
```diff
@@ -331,6 +339,184 @@ async fn fetch_updated_gossip(
...
+
+    // If there is already some metadata, we add the connection string to it
+    // Otherwise we create a new metadata with the connection string
+    let new_info = match current {
+        Some(current) => current.with_connection_string(connection_string.to_string()),
+        None => LnPeerMetadata {
+            connection_string: Some(connection_string.to_string()),
+            ..Default::default()
+        },
+    };
```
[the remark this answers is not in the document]
benthecarman: yeah i guess i could do that, time is just a bitch in wasm haha

src/gossip.rs
```diff
@@ -331,6 +339,268 @@ async fn fetch_updated_gossip(
...
+        } else {
+            (other.clone(), self.clone())
+        };
+
+        // combine nodes from both
+        let nodes = primary
+            .nodes
+            .into_iter()
+            .chain(secondary.nodes.into_iter())
+            .collect();
```
benthecarman, line 406: I need to dedup this

TonyGiorgio:
needs rebase

src/peermanager.rs
```diff
@@ -127,3 +139,139 @@ impl PeerManager for PeerManagerImpl {
...
+    ) -> Result<bool, LightningError> {
+        let msg_clone = msg.clone();
+        spawn_local(async move {
+            // We use RGS to sync gossip, but we can save the node's metadata (alias and color)
+            // we should only save it for relevant peers however (i.e. peers we have a channel with)
+            if read_peer_info(&msg_clone.contents.node_id)
+                .await
+                .unwrap_or(None)
+                .is_some()
+            {
```
TonyGiorgio, line 177: I think `if let Some(_)` would be cleaner here instead of `unwrap_or(None).is_some()`

TonyGiorgio (changes requested):
Some of the global peer stuff combined with individual peer connections/lists seem to be intertwined still unless I'm wrong. I wish there was an easier way to test and ensure we have multi node support.

src/peermanager.rs
benthecarman: The `unwrap_or` is for a result, I used `ok()` instead, should be a little cleaner

src/node.rs
```diff
@@ -385,24 +392,31 @@ impl Node {
                     continue;
                 }
 
-                let peer_connections = connect_persister.list_peer_connection_info();
+                let peer_connections = get_all_peers().await.unwrap_or_default();
```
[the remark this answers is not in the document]
benthecarman: yes but on line 401 we filter out ones that our node isn't included in

mutiny-core/src/nodemanager.rs
```diff
@@ -933,19 +953,25 @@ impl NodeManager {
...
+                // node id should be safe here
+                pubkey: secp256k1::PublicKey::from_slice(node_id.as_slice())
+                    .expect("Invalid pubkey"),
+                connection_string: metadata.connection_string.clone(),
+                alias: metadata.alias.clone(),
+                color: metadata.color.clone(),
+                label: metadata.label.clone(),
                 is_connected: false,
             })
             .collect();
```
TonyGiorgio, line 971: Wait this could just be a side effect with a bug (or not intended behavior) with our existing API. We're just doing a normal `list_peers`, but perhaps we should change the API to indicate a specific node and the peers that node should be connected to?

TonyGiorgio (approved):
Seperating node and general peers is difficult and unsure if we should be doing that anyways, but existing behavior so not something for this to solve or get hung up on. Nice work, looks good to me!

TonyGiorgio:
needs rebase

PR #361, opened 2023-04-14
Improve typesafety, cleanup deps

benthecarman:
Follow up to the wasm split, this should improve the type safety for the node manager and found a couple places to do it more internally.

Also cleaned up some dependencies as we didn't need that because they were being re-exported (namely secp256k1 and bitcon_hashes)

Finally, added some small stuff to get CI more reliable :pray:


TonyGiorgio (approved):
Huge refactor, really nice!

PR #368, opened 2023-04-15
Update license for new year + company

benthecarman:


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2022 Bitcoin Dev Shop
+Copyright (c) 2022-2023 Mutiny Wallet Developers
```
TonyGiorgio, line 3: I think it technically should be `Mutiny Wallet Inc.`

TonyGiorgio (changes requested):

TonyGiorgio (approved):

PR #369, opened 2023-04-15
Fix variable names for clippy

benthecarman:
Latest version of wasm-bindgen fixed the variable name casing issue so we can re-enable those warnings.

Fixed the issues where we caused it as well.

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -1,13 +1,6 @@
 #![crate_name = "mutiny_core"]
-// wasm_bindgen uses improper casing and it needs to be turned off:
-// https://github.com/rustwasm/wasm-bindgen/issues/2882
-// wasm is also considered "extra_unused_type_parameters"
-#![allow(
-    incomplete_features,
-    non_snake_case,
-    non_upper_case_globals,
-    clippy::extra_unused_type_parameters
-)]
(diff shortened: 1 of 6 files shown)


TonyGiorgio:
Can't wasm-bindgen actually be removed from `mutiny-core`?

benthecarman:
> Can't wasm-bindgen actually be removed from mutiny-core?

Not yet, we still use it for the local storage and indexeddb stuff

mutiny-core/src/lib.rs
```diff
@@ -1,13 +1,6 @@
...
-// https://github.com/rustwasm/wasm-bindgen/issues/2882
-// wasm is also considered "extra_unused_type_parameters"
-#![allow(
-    incomplete_features,
-    non_snake_case,
-    non_upper_case_globals,
-    clippy::extra_unused_type_parameters
-)]
+// wasm is considered "extra_unused_type_parameters"
+#![allow(incomplete_features, clippy::extra_unused_type_parameters)]
```
TonyGiorgio, line 3: Arent these in `mutiny-wasm/src/lib.rs` too?

benthecarman: good catch

TonyGiorgio (approved):

PR #370, opened 2023-04-17
Crates publish

TonyGiorgio:
I think this CI should work. Already published locally with the same cargo command: https://crates.io/crates/mutiny-core

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,7 +3,37 @@ on:
   release:
     types: [published]
 jobs:
-  build:
+  mutiny-core:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          components: clippy
+          target: wasm32-unknown-unknown
+          override: true
+          profile: minimal
+
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: cargo-${{ runner.os }}-release-${{ hashFiles('**/Cargo.toml') }}
+          restore-keys: |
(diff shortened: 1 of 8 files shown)


benthecarman:
ugh the CI issues you are getting are ugly, I thought I fixed that in master, might need to add more flags to the webdriver.json file

TonyGiorgio:
I think it might be timeout related since not all the tests seem to be running, increased to 60s to test it out

TonyGiorgio:
@benthecarman I increased timeout to 60s and it seems like these 3 tests are not making it in: 

```
test mutiny_core::gossip::test::test_merge_peer_info ... ok	
test mutiny_core::keymanager::tests::derive_pubkey_child_from_seed ... ok	
test mutiny_core::ldkstorage::test::test_persist_payment_info ... ok
```

Is there any significance to them? Were they the ones you were using? I can try commenting them out and seeing if that works.

benthecarman:
> test mutiny_core::gossip::test::test_merge_peer_info ... ok	

should be a really fast test, don't know how that'd not be working. Maybe just need to bump the timeout

TonyGiorgio:
Worked now, I'll start adding some back to see if it was just stalled on one of them.

benthecarman:
@TonyGiorgio maybe try making `test_persist_payment_info` an async fn?

TonyGiorgio:
> @TonyGiorgio maybe try making `test_persist_payment_info` an async fn?

Interesting, I'll try that next. Seems like it is something with `test_persist_payment_info`. I've added the cleanup function at the end of it, maybe that's conflicting.

TonyGiorgio:
Do you think we need a cleanup for indexdb as well?

benthecarman:
> Do you think we need a cleanup for indexdb as well?

Yeah that'd be good to add

benthecarman (approved):

PR #372, opened 2023-04-17
Add waila to mutiny-wasm

### LINKED ISSUE #334: Integrate WAILA
We should add as a utility function for the front end

https://github.com/MutinyWallet/bitcoin-waila

benthecarman:
Closes #334

### DIFF
--- a/mutiny-wasm/Cargo.toml
+++ b/mutiny-wasm/Cargo.toml
@@ -19,6 +19,7 @@ wasm-bindgen-futures = "0.4.33"
 serde = { version = "^1.0", features = ["derive"] }
 serde_json = { version = "^1.0" }
 bitcoin = { version = "0.29.2", default-features = false, features = ["serde", "secp-recovery"] }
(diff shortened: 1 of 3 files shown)


futurepaul (approved):
this looks good to me at a glance, high probability I'll have some needs going forward but gotta play with it to know for sure. one thing I'd like but isn't pressing is npub scanning

TonyGiorgio:
Why just in the wasm package instead of `mutiny-core`?

benthecarman:
> Why just in the wasm package instead of mutiny-core?

Anywhere else we can just call the actual library when we need it, this is just a wasm bindgen wrapper for the library basically.

TonyGiorgio:
Wouldn't it make more sense to have `waila` be a wasm ts package on it's own then?

benthecarman:
> Wouldn't it make more sense to have waila be a wasm ts package on it's own then?

That's a good point

TonyGiorgio:
I can see that being annoying, but if we're not going to use it elsewhere and there's going to be increasing things that only the frontend wants to use for it then I can see keeping it as a dedicated package on it's own. Otherwise, if `mutiny-core` will use it, I can see putting it in there and wrapping it in `mutiny-wasm`.

PR #374, opened 2023-04-17
Add last updated to MutinyInvoice

benthecarman:


futurepaul (commented):
why is this optional, wouldn't it be set at creation?

mutiny-core/src/nodemanager.rs
```diff
@@ -117,6 +118,7 @@ impl From<Invoice> for MutinyInvoice {
             paid: false,
             fees_paid: None,
             is_send: false, // todo this could be bad
+            last_updated: None,
```
TonyGiorgio, line 121: Should you use the timestamp field above?

benthecarman:
> why is this optional, wouldn't it be set at creation?

didn't know how to handle from converting from a normal invoice but @TonyGiorgio showed me the way!

futurepaul (approved):
thank you!

TonyGiorgio (approved):

Issue #236, opened 2022-12-17
Move storage to IndexedDb


benthecarman:
Almost there, just a few left

- [x] #393 
- [ ] Fee estimates
- [ ] Node List

TonyGiorgio:
Partially done, need to do it for the rest of the storage components.

PR #377, opened 2023-04-19
Move rest of ldk storage to indexed db

benthecarman:
This does the in memory HashMap of our data, and writes async to indexed db. I needed to put it in an `Arc<RwLock>`, otherwise we needed LDK to make everything mut. Not too sure on the safety implications of this and still need to do some testing.

Will try bdk in a follow up, but would rather wait until bdk core.

Also moved our test utils to a separate module that we can export. Eventually these current ones will probably be in mutiny-wasm but makes it so we don't have to reimplement the same thing on both ends for now

Part of #236

### DIFF
--- a/mutiny-core/src/event.rs
+++ b/mutiny-core/src/event.rs
@@ -20,20 +20,32 @@ use lightning::{
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
 
-#[derive(Serialize, Deserialize, Clone)]
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub(crate) struct PaymentInfo {
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub preimage: Option<[u8; 32]>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub secret: Option<[u8; 32]>,
     pub status: HTLCStatus,
+    #[serde(skip_serializing_if = "MillisatAmount::is_none")]
     pub amt_msat: MillisatAmount,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub fee_paid_msat: Option<u64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub bolt11: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub payee_pubkey: Option<PublicKey>,
     pub last_update: u64,
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub(crate) struct MillisatAmount(pub Option<u64>);
+
+impl MillisatAmount {
+    pub fn is_none(&self) -> bool {
+        self.0.is_none()
+    }
+}
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub(crate) enum HTLCStatus {
@@ -575,5 +587,45 @@ impl EventHandler {
             }
             Event::HTLCIntercepted { .. } => {}
         }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use crate::event::{HTLCStatus, MillisatAmount, PaymentInfo};
+    use crate::utils;
+    use bitcoin::secp256k1::PublicKey;
+    use lightning::ln::PaymentHash;
+    use std::str::FromStr;
+
+    use wasm_bindgen_test::{wasm_bindgen_test as test, wasm_bindgen_test_configure};
+    wasm_bindgen_test_configure!(run_in_browser);
+
+    #[test]
+    fn test_payment_info_serialization_symmetry() {
+        let preimage = [1; 32];
+        let pubkey = PublicKey::from_str(
+            "02465ed5be53d04fde66c9418ff14a5f2267723810176c9212b722e542dc1afb1b",
+        )
+        .unwrap();
+
+        let payment_info = PaymentInfo {
+            preimage: Some(preimage),
+            status: HTLCStatus::Succeeded,
+            amt_msat: MillisatAmount(Some(420)),
+            fee_paid_msat: None,
+            bolt11: None,
+            payee_pubkey: Some(pubkey),
+            secret: None,
+            last_update: utils::now().as_secs(),
+        };
+
+        let serialized = serde_json::to_string(&payment_info).unwrap();
+        let deserialized: PaymentInfo = serde_json::from_str(&serialized).unwrap();
+        assert_eq!(payment_info, deserialized);
+
+        let serialized = serde_json::to_value(&payment_info).unwrap();
+        let deserialized: PaymentInfo = serde_json::from_value(serialized).unwrap();
+        assert_eq!(payment_info, deserialized);
     }
 }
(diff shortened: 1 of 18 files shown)


benthecarman:
i have no idea why tests are failing

mutiny-core/src/indexed_db.rs
```diff
@@ -0,0 +1,248 @@
...
+                _ => serde_wasm_bindgen::from_value(value)?,
+            };
+            map.insert(key, json);
+        }
+
+        Ok(map)
+    }
+
+    fn needs_encryption(key: &str) -> bool {
+        // todo, what other keys need to be encrypted?
```
TonyGiorgio, line 229: probably just channel monitor? Might be worth asking the LDK team. I dont think channel mananger needs it so much.

But eventually probably everything if we're doing remote storage.

mutiny-core/src/ldkstorage.rs
```diff
@@ -84,7 +84,13 @@ impl MutinyNodePersister {
     // that has the concatenated node_id
     fn read_value(&self, _key: &str) -> Result<Vec<u8>, MutinyError> {
         let key = self.get_key(_key);
-        self.storage.get(key).map_err(MutinyError::read_err)
```
TonyGiorgio, line 87: Did this return an error before if the value did not exist?

mutiny-core/src/ldkstorage.rs
```diff
@@ -93,8 +99,10 @@ impl MutinyNodePersister {
     ) -> Result<Vec<(BlockHash, ChannelMonitor<InMemorySigner>)>, io::Error> {
         // Get all the channel monitor buffers that exist for this node
         let suffix = self.node_id.as_str();
-        let channel_monitor_list: HashMap<String, Vec<u8>> =
-            self.storage.scan(MONITORS_PREFIX_KEY, Some(suffix));
+        let channel_monitor_list: HashMap<String, Vec<u8>> = self
+            .storage
+            .scan(MONITORS_PREFIX_KEY, Some(suffix))
+            .map_err(|_| io::ErrorKind::Other)?;
```
TonyGiorgio, line 105: `io::ErrorKind` or `MutinyStorageError::Other`?

mutiny-core/src/gossip.rs
```diff
@@ -640,7 +640,7 @@ mod test {
     use uuid::Uuid;
     use wasm_bindgen_test::{wasm_bindgen_test as test, wasm_bindgen_test_configure};
 
-    use crate::test::*;
+    use crate::test_utils::*;
```
TonyGiorgio, line 643: I think the tests at the bottom of this file need `crate_test_utils` too. 

```
    #[test]
    // hack to disable this test
    #[cfg(feature = "ignored_tests")]
    async fn test_gossip() {
        crate::test::log!("test RGS sync");
```

mutiny-wasm/Cargo.toml
```diff
@@ -25,6 +25,7 @@ lnurl-rs = { version = "0.2", default-features = false }
 wasm-logger = "0.2.0"
 log = "0.4.17"
 gloo-storage = "0.2.2"
+rexie = "0.4"
```
TonyGiorgio, line 28: Does this have to get pulled in? Probably eventually but not seeing this used yet.

TonyGiorgio (approved):
Just smaller things and questions, overall great work! I guess the next follow up PR is going to be to get rid of the rest?

mutiny-core/src/ldkstorage.rs
benthecarman: Yeah

mutiny-core/src/ldkstorage.rs
benthecarman: this function returns `io::ErrorKind` so think should keep it that way at least for this PR, if we want to change the error types of these functions probably more suited for a follow up

mutiny-wasm/Cargo.toml
benthecarman: oh yeah can remove because I did the test utils thing

mutiny-core/src/gossip.rs
benthecarman: made it so `ignored_tests` depends on `test-utils` so it should be safe

mutiny-core/src/indexed_db.rs
benthecarman: asked

TonyGiorgio: Cool, either way this can be done in a follow up, good to leave that todo in there

mutiny-core/src/gossip.rs
TonyGiorgio: right but shouldn't it be:

```
 crate::test_utils::log!("test RGS sync");
```

benthecarman: oh i see, good catch, fixed

TonyGiorgio:
For some reason this fails locally for me:

```
failures:                                                                                                                                                                                     
                                                                                                                                                                                              
    mutiny_core::nodemanager::tests::created_new_nodes                                                                                                                                        
    mutiny_core::nodemanager::tests::correctly_show_seed                                                                                                                                      
                                                                                                                                                                                              
test result: FAILED. 28 passed; 2 failed; 0 ignored                                                                                                                                           
console.log div contained:                                                                                                                                                                    
    test parse peer info with no port                                                                                                                                                         
    test parse peer info                                                                                                                                                                      
    creating pubkeys from a child seed                                                                                                                                                        
    test proxy to url                                                                                                                                                                         
    creating new nodes                                                                                                                                                                        
    showing seed                                                                                                                                                                              
    creating node manager!                                                                                                                                                                    
                                                                                                                                                                                              
console.log div contained:                                                                                                                                                                    
    panicked at 'node manager should initialize: ReadError { source: Other(Failed to deserialize) }', mutiny-core/src/nodemanager.rs:1196:10                                                  
                                                                                                                                                                                              
    Stack:                                                                                                                                                                                    
                                                                                                                                                                                              
    getImports/imports.wbg.__wbg_new_abda76e883ba8a5f/<@http://127.0.0.1:43893/wasm-bindgen-test:705:21                                                                                       
    logError@http://127.0.0.1:43893/wasm-bindgen-test:245:18   
```

mutiny-core/src/nodemanager.rs
```diff
@@ -1216,6 +1219,7 @@ mod tests {
         }
 
         cleanup_test();
```
TonyGiorgio, line 1221: `cleanup_test()` can be removed

TonyGiorgio:
needs rebase

mutiny-core/src/nodemanager.rs
TonyGiorgio: I think you added this back? I still think `cleanup_test()` can be removed.

TonyGiorgio (approved):
Small things but otherwise really nice work. Like the new gloo api's better too.

mutiny-core/src/event.rs
```diff
@@ -20,21 +20,33 @@ use lightning::{
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
 
-#[derive(Serialize, Deserialize, Clone)]
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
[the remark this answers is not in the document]
benthecarman: I just added them because we can, no real drawback to having them

PR #379, opened 2023-04-19
A ton of API cleanups

benthecarman:
When we started this we were pretty new to rust and just used clone() everywhere. I went through and found everywhere we should be able to references and fixed that. Should make us slightly more efficient and make the APIs a little cleaner to user (less having to clone()).

Sorry for the mondo PR

### DIFF
--- a/mutiny-core/src/event.rs
+++ b/mutiny-core/src/event.rs
@@ -204,7 +204,7 @@ impl EventHandler {
                 };
                 match self
                     .persister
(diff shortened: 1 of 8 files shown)


TonyGiorgio (commented):
Is there any cases where this could bit us at all? Rust pretty much makes this compiler safe, right? Any scenarios where we would want to `.clone()` vs pass by reference?

TonyGiorgio:
famous last words

TonyGiorgio (approved):

PR #393, opened 2023-04-21
bdk core

### LINKED ISSUE #387: BDK Core

benthecarman:
Closes #387 

Still some clean up we can do but this migrates us to bdk core. Was able to test an on-chain receive to opening and closing a channel and it all worked :rocket: 

Second commit I move the wallet out of a mutex and then should be able to clean up a lot more of the code in the future


notmandatory:
FYI, [BDK PR 926 ](https://github.com/bitcoindevkit/bdk/pull/926) will make some large-ish changes in the `1.0.0-alpha.1` release, but we'll also be updating examples and can help you work through what you'll need to change.

mutiny-core/src/nodemanager.rs
```diff
@@ -522,45 +516,51 @@ impl NodeManager {
...
+        // if we found a tx we should try to import it into the wallet
+        if let Some(details) = details_opt.clone() {
+            let mut wallet = self.wallet.wallet.try_write()?;
+
+            wallet
+                .insert_tx(
+                    details.transaction.clone().unwrap(),
+                    details.confirmation_time,
+                )
+                .map_err(|_| MutinyError::ChainAccessFailed)?; // TODO better error
```
TonyGiorgio, line 528: Is there any problem with this being written multiple times when it is not needed? Perhaps this is being called into it confirms, in which case this `insert_tx` will keep being triggered.

mutiny-core/src/wallet.rs
```diff
@@ -69,37 +68,59 @@ impl MutinyWallet {
...
+        wallet.commit()?;
 
         Ok(())
     }
 
-    pub async fn list_utxos(&self) -> Result<Vec<LocalUtxo>, MutinyError> {
-        Ok(self.wallet.lock().await.list_unspent()?)
+    pub fn list_utxos(&self) -> Result<Vec<LocalUtxo>, MutinyError> {
+        Ok(self.wallet.try_read()?.list_unspent())
     }
```
TonyGiorgio, line 105: I see a lot of removals of `async`. Is BDK Core now keeping stuff in memory so it doesn't need to read or write directly from indexdb?

TonyGiorgio (commented):
Wild how much simpler this is, feels too good to be true

mutiny-core/Cargo.toml
```diff
@@ -23,7 +23,8 @@ bip32 = "0.4.0"
 bitcoin-bech32 = "0.12"
 js-sys = "0.3.60"
 bitcoin = { version = "0.29.2", default-features = false, features = ["serde", "secp-recovery"] }
-bdk = { version = "0.28.0", default-features = false, features = ["keys-bip39", "esplora", "use-esplora-reqwest", "async-interface"] }
+bdk = { git = "https://github.com/benthecarman/bdk.git", branch = "mutiny-no-std", default-features = false }
```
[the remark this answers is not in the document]
benthecarman: They haven't merged the no-std stuff but I was impatient

mutiny-core/src/nodemanager.rs
benthecarman: The docs for this function make it sound like it is idempotent. It shouldn't be a big issue either as this doesnt write to disk immediately, just stages it as well (according to the docs)

mutiny-core/src/wallet.rs
benthecarman: Yes exactly

TonyGiorgio (approved):
Huge, good job. Haven't tested it but looking forward to all the improvements here and the things to follow.

PR #395, opened 2023-04-22
Write to indexeddb first

benthecarman:
This is a safety check to make sure that we write to indexeddb first. This should be done so if writing to the in memory map fails or if the program stops in the middle, we make sure we right to indexeddb which is actually persistent. That way if writing to memory fails, then if the user just refreshes, the data is not lost.

Second commit changes it to use `try_read` and `try_write`, we need to use these because the others would block if waiting which could cause a panic for us. This instead will return an Error and can be handled correctly.

### DIFF
--- a/mutiny-core/src/error.rs
+++ b/mutiny-core/src/error.rs
@@ -228,6 +228,12 @@ impl<G> From<std::sync::TryLockError<G>> for MutinyError {
     }
 }
 
+impl<G> From<std::sync::TryLockError<G>> for MutinyStorageError {
+    fn from(_e: std::sync::TryLockError<G>) -> Self {
+        MutinyStorageError::LockError
+    }
+}
+
 impl From<bitcoin::hashes::hex::Error> for MutinyError {
     fn from(_e: bitcoin::hashes::hex::Error) -> Self {
         MutinyError::ReadError {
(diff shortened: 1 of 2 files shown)


TonyGiorgio:
> Second commit changes it to use try_read and try_write, we need to use these because the others would block if waiting which could cause a panic for us. This instead will return an Error and can be handled correctly.

How does this work then? What if two things are trying to write to memory at the same time? One fails?

benthecarman:
> How does this work then? What if two things are trying to write to memory at the same time? One fails? 

Yes we can't have concurrent locks as it would cause us to block. Before it would block and cause a panic, now this will cause an error instead so it should be safer. We don't really have a way around this besides getting bdk and ldk to implement async storage.

TonyGiorgio:
> > How does this work then? What if two things are trying to write to memory at the same time? One fails?
> 
> Yes we can't have concurrent locks as it would cause us to block. Before it would block and cause a panic, now this will cause an error instead, which we can properly handle, so it should be safer. We don't really have a way around this besides getting bdk and ldk to implement async storage.

I still think we can do something `spawn_local` & `channel` related to make sure that writes are happening correctly, but maybe we just wait to see how it performs without that first?

benthecarman:
> I still think we can do something spawn_local & channel related to make sure that writes are happening correctly, but maybe we just wait to see how it performs without that first?

That might be a better long term solution but for now this PR is a strict upgrade.

TonyGiorgio (approved):

PR #396, opened 2023-04-22
Create Custom Key Manager

benthecarman:
This makes it so any receive address generated by LDK will be to our bdk wallet, this should make closes take 1 less on-chain transaction because we shouldn't need to sweep.

One caveat will be that we can't open channels to nodes that don't support taproot, at this point every implementation has supported this for over a year so we should be fine.


TonyGiorgio (approved):
Awesome, love it

TonyGiorgio (approved):

PR #399, opened 2023-04-24
Correctly handle errors when creating bdk wallet

benthecarman:
Before we just had a bunch of unwraps and expects everywhere, this makes it so we actually handle the error.

### DIFF
--- a/mutiny-core/src/keymanager.rs
+++ b/mutiny-core/src/keymanager.rs
@@ -230,13 +230,9 @@ mod tests {
         let db = MutinyStorage::new("".to_string()).await.unwrap();
         let fees = Arc::new(MutinyFeeEstimator::new(db.clone()));
 
-        let wallet = Arc::new(MutinyWallet::new(
-            &mnemonic,
-            db,
-            Network::Testnet,
-            Arc::new(esplora),
-            fees,
-        ));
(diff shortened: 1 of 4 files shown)


mutiny-core/src/keymanager.rs
```diff
@@ -228,12 +228,9 @@ mod tests {
...
         let db = MutinyStorage::new("".to_string()).await.unwrap();
 
-        let wallet = Arc::new(MutinyWallet::new(
-            &mnemonic,
-            db,
-            Network::Testnet,
-            Arc::new(esplora),
-        ));
+        let wallet = Arc::new(
+            MutinyWallet::new(&mnemonic, db, Network::Testnet, Arc::new(esplora)).unwrap(),
```
TonyGiorgio, line 232: unwrap?

TonyGiorgio (changes requested):

mutiny-core/src/keymanager.rs
benthecarman: this is in the test, we can't `?` because it returns `()` not a `Result`

TonyGiorgio: Oh didn't see it was in test

TonyGiorgio (approved):

PR #400, opened 2023-04-24
Add import and export json

benthecarman:


mutiny-wasm/src/lib.rs
```diff
@@ -362,6 +362,16 @@ impl NodeManager {
         Ok(self.inner.get_bitcoin_price().await?)
     }
 
+    pub async fn export_json(&self) -> Result<JsValue, MutinyJsError> {
```
futurepaul, line 365: maybe this should just be a string? I think this will deserialize to a javascript object

futurepaul (changes requested):

benthecarman:
> Is everything in indexeddb now? They just all go into the WALLET_OBJECT_STORE_NAME ? All the nodes and everything?

Yeah, the only stuff that doesn't is the network graph and scorer

Issue #373, opened 2023-04-17
Set fee paid for receives through LSP in PaymentInfo


benthecarman:
In our `PaymentInfo` struct we have a fee paid field for sends. We could use this for receives as well to save how much fees we paid the LSP.

PR #407, opened 2023-04-24
Receiver pays fee by default

### LINKED ISSUE #363: Receiver pays fee for lsp by default

TonyGiorgio:
Fixes #363 & #373

Still need to test with the mutiny-web

### DIFF
--- a/mutiny-core/src/lspclient.rs
+++ b/mutiny-core/src/lspclient.rs
@@ -50,8 +50,20 @@ pub struct ProposalResponse {
     pub jit_bolt11: String,
 }
 
+#[derive(Serialize, Deserialize)]
+pub struct FeeRequest {
+    pub pubkey: String,
+    pub amount_msat: u64,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct FeeResponse {
+    pub fee_amount_msat: u64,
+}
+
 const GET_INFO_PATH: &str = "/api/v1/info";
 const PROPOSAL_PATH: &str = "/api/v1/proposal";
+const FEE_PATH: &str = "/api/v1/fee";
 
 impl LspClient {
     pub async fn new(url: &str) -> anyhow::Result<Self> {
@@ -115,5 +127,18 @@ impl LspClient {
             .await?;
 
         Ok(proposal_response.jit_bolt11)
+    }
+
+    pub(crate) async fn get_lsp_fee_msat(&self, fee_request: FeeRequest) -> anyhow::Result<u64> {
+        let fee_response: FeeResponse = self
+            .http_client
+            .post(format!("{}{}", &self.url, FEE_PATH))
+            .json(&fee_request)
+            .send()
+            .await?
+            .json()
+            .await?;
+
+        Ok(fee_response.fee_amount_msat)
     }
 }
--- a/mutiny-core/src/node.rs
+++ b/mutiny-core/src/node.rs
@@ -570,6 +570,39 @@ impl Node {
         description: String,
         route_hints: Option<Vec<PhantomRouteHints>>,
     ) -> Result<Invoice, MutinyError> {
+        let invoice = self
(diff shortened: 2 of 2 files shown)


mutiny-core/src/node.rs
```diff
@@ -569,6 +569,68 @@ impl Node {
...
+            // LSP requires an amount
+            let amount_sat = amount_sat
+                .ok_or(MutinyError::BadAmountError)
+                .and_then(|amount| {
+                    if amount == 0 {
+                        Err(MutinyError::BadAmountError)
+                    } else {
+                        Ok(amount)
+                    }
+                })?;
```
benthecarman, line 584: can simplify this as `amount_sat.filter(|a| a > 0).ok_or(MutinyError::BadAmountError)`

mutiny-core/src/node.rs
```diff
@@ -569,6 +569,68 @@ impl Node {
...
+            (amount_sat, None)
+        };
+
+        let invoice = self
+            .create_internal_invoice(amount_sat, lsp_fee_sat, description, route_hints)
+            .await?;
+
+        if let Some(lsp) = self.lsp_client.clone() {
+            self.connect_peer(PubkeyConnectionInfo::new(&lsp.connection_string)?, None)
+                .await?;
```
benthecarman, line 607: should check that we aren't connected already, I think we do this at the nodemanager level, not the node level

mutiny-core/src/node.rs
```diff
@@ -569,6 +569,68 @@ impl Node {
...
+                        Err(MutinyError::BadAmountError)
+                    } else {
+                        Ok(amount)
+                    }
+                })?;
+
+            // check the fee from the LSP
+            let lsp_fee = lsp
+                .get_lsp_fee_sat(FeeRequest {
+                    pubkey: self.pubkey.to_string(),
```
benthecarman, line 589: .to_hex() might be better here

mutiny-core/src/node.rs
TonyGiorgio: Not, this calls `connect_peer_if_necessary` so it does that check

mutiny-core/src/lspclient.rs
```diff
@@ -50,8 +50,20 @@ pub struct ProposalResponse {
...
+pub struct FeeRequest {
+    pub pubkey: String,
+    pub amount_msat: u64,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct FeeResponse {
+    pub fee_amount_msat: u64,
+}
+
```
benthecarman, line 63: for these would be good to derive debug and clone too

TonyGiorgio: If and when needed :)

mutiny-core/src/node.rs
```diff
@@ -569,8 +569,65 @@ impl Node {
...
+
+    async fn create_internal_invoice(
+        &self,
+        amount_sat: Option<u64>,
+        fee_amount_sat: Option<u64>,
+        description: String,
+        route_hints: Option<Vec<PhantomRouteHints>>,
     ) -> Result<Invoice, MutinyError> {
         let amount_msat = amount_sat.map(|s| s * 1_000);
+        let fee_amount_msat = fee_amount_sat.map(|s| s * 1_000);
```
benthecarman, line 630: You have it dividing by 1000 when we get the fee from the lsp, just to multiply it by 1000 here, should just have `get_lsp_fee_sat` return in msats instead so we don't lose precision for no reason. Can then rename to `get_lsp_fee_msat`

benthecarman (changes requested):

mutiny-core/src/node.rs
TonyGiorgio: fixed.

benthecarman (approved):

PR #409, opened 2023-04-24
Add value() function for wasm models

benthecarman:
@futurepaul wanted this

### DIFF
--- a/mutiny-wasm/src/models.rs
+++ b/mutiny-wasm/src/models.rs
@@ -1,5 +1,6 @@
 use bitcoin::hashes::hex::ToHex;
 use bitcoin::secp256k1::PublicKey;
+use gloo_utils::format::JsValueSerdeExt;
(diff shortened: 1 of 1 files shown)


mutiny-core/src/gossip.rs
```diff
@@ -227,7 +227,7 @@ async fn write_network_graph_to_store(
 }
 
 async fn write_scorer_to_store(store: &Store, scorer_str: &str) -> Result<(), MutinyError> {
-    let scorer_js = JsValue::from_serde(scorer_str)?;
+    let scorer_js = JsValue::from_serde(&scorer_str)?;
```
TonyGiorgio, line 230: Did this mean to be apart of it?

TonyGiorgio (changes requested):
Needs clippy changes

TonyGiorgio (approved):

PR #416, opened 2023-04-25
Reduced connection-related logging

### LINKED ISSUE #412: Remove WSS error if it's just because other node is offline
This is too noisy, should be able to tone down the other logs too now, @futurepaul  would be thankful.

```
mutiny-core/src/socket.rs:408connection closed due to: CloseEvent { code: 1006, reason: "", was_clean: false }
```

We can do exponential back off too (for #415)

TonyGiorgio:
Fixes #412

### DIFF
--- a/mutiny-core/src/socket.rs
+++ b/mutiny-core/src/socket.rs
@@ -9,7 +9,7 @@ use gloo_net::websocket::Message;
 use lightning::ln::peer_handler;
 use lightning::ln::peer_handler::SocketDescriptor;
 use ln_websocket_proxy::MutinyProxyCommand;
-use log::{debug, error, info, trace};
+use log::{debug, error, trace};
 use std::collections::HashMap;
 use std::hash::Hash;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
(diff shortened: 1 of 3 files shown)


mutiny-core/src/socket.rs
```diff
@@ -404,7 +403,9 @@ pub(crate) fn schedule_descriptor_read(
                             error!("got connection error");
                         }
                         gloo_net::websocket::WebSocketError::ConnectionClose(e) => match e.code {
-                            1000 => debug!("normal connection closure"),
+                            // TODO make connection closes from proxy be 1000
+                            1000 => trace!("normal connection closure"),
+                            1006 => debug!("abnormal connection closure"),
```
benthecarman, line 408: Would be good to make these an enum in the ln-websocket-proxy library eventually

benthecarman (approved):

mutiny-core/src/socket.rs
TonyGiorgio: It's a underlying WS library thing, and that lib doesn't make them `enum`, have to look up the code online.

PR #420, opened 2023-04-25
Better handle syncing transaction from check_address

### LINKED ISSUE #414: `check_address` doesn't work for confirmed addresses?
- If I call `check_address` with a fresh address it returns null.
- If I call it with an address that has been paid I get the tx info I expect.
- But once it's confirmed I get an error: `Failed to conduct chain access operation.`

benthecarman:
Closes #414

When we were try to insert the tx it could can fail if our chain tip isn't caught up, so confirmed txs would fail.

This fixes it so we do it in a spawn_local so it doesn't hold up the api call and will still return correctly even if the sync failed (because the sync is just a small optimization).

Also changed it so if the transaction is confirmed that we properly handle it by either giving bdk the block the tx was included in or if we don't have that, just doing a normal sync so we catch everything correctly.

### DIFF
--- a/mutiny-core/src/wallet.rs
+++ b/mutiny-core/src/wallet.rs
@@ -1,13 +1,14 @@
 use std::str::FromStr;
 use std::sync::{Arc, RwLock};
 
+use bdk::chain::{BlockId, ConfirmationTime};
 use bdk::template::DescriptorTemplateOut;
 use bdk::{FeeRate, LocalUtxo, SignOptions, TransactionDetails, Wallet};
 use bdk_esplora::{esplora_client, EsploraAsyncExt};
 use bdk_macros::maybe_await;
 use bip39::Mnemonic;
 use bitcoin::util::bip32::{ChildNumber, DerivationPath, ExtendedPrivKey};
-use bitcoin::{Address, Network, Script, Txid};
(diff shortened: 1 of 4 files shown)


mutiny-core/src/wallet.rs
```diff
@@ -87,6 +88,37 @@ impl MutinyWallet {
...
+                // we can insert it directly
+                if let Some(block_id) = block_id {
+                    let mut wallet = self.wallet.try_write()?;
+                    wallet.insert_checkpoint(block_id)?;
+                    wallet.insert_tx(tx, position)?;
+                } else {
+                    // if the transaction is confirmed and we don't have the block id,
+                    // we should just sync the wallet otherwise we can get an error
+                    // with the wallet being behind the blockchain
+                    self.sync().await?
```
TonyGiorgio, line 109: I'm guessing the `sync()` function then handles the inserts?

And there's no clashing or problems that we are just inserting directly into it's db?

benthecarman: > I'm guessing the sync() function then handles the inserts?

Yeah it does.

> And there's no clashing or problems that we are just inserting directly into it's db?

Yeah they do a similar in memory caching thing that we do.

TonyGiorgio (approved):

PR #427, opened 2023-04-26
Better handle optional description for invoices

### LINKED ISSUE #426: create_bip21 with an empty string makes a bad bip21
`create_bip21(bigAmount, "")`

results in

```url
bitcoin:tb1p4ltq2dpm3dpmlxrn2nendmz05makpyqt38pvxj9rglwv5p77jcyq3spdc4?amount=0.0001&&lightning=lntbs100u1pjyngklsp5x3hktafqkn6y8rxfgx65pfkqpvsvzkzkdllg5g9dt86z23y68vkqpp5k3lvku9tjt8jfe3c74qruljvz9kv9mxwmpespryqdsurr0wqkhvqdqqnp4qdn2hj8tfknpuvdg6tz9yrf3e27ltrx9y58c24jh89lnm43yjwfc5xqrpwjcqpj9qrsgqn0c6zlxclcz6a8hteet27lm7zkwcwnpglae4qtzgze9xt2juh0aq0ww7ralqwnpj36ng7gpkqgfq4xqs0vezqn92f0g0u7vp3879qggq9jtzkk
```

benthecarman:
Closes #426

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -1326,7 +1326,7 @@ mod tests {
 
         let expected: MutinyInvoice = MutinyInvoice {
             bolt11: Some(invoice),
-            description: Some("".to_string()),
+            description: None,
             payment_hash,
             preimage: Some(preimage.to_hex()),
             payee_pubkey: None,
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -107,7 +107,13 @@ pub struct MutinyInvoice {
 impl From<Invoice> for MutinyInvoice {
     fn from(value: Invoice) -> Self {
         let description = match value.description() {
-            InvoiceDescription::Direct(a) => Some(a.to_string()),
+            InvoiceDescription::Direct(a) => {
+                if a.is_empty() {
+                    None
+                } else {
+                    Some(a.to_string())
+                }
+            }
             InvoiceDescription::Hash(_) => None,
         };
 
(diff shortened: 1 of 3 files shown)


benthecarman:
idk why browser tests are failing now?

mutiny-core/src/nodemanager.rs
```diff
@@ -446,6 +446,9 @@ impl NodeManager {
             return Err(MutinyError::WalletOperationFailed);
         };
 
+        // if empty string, treat as None
+        let description = description.filter(|d| !d.is_empty());
+
         // TODO if there's no description should be something random I guess
         let invoice = self
             .create_invoice(amount, description.clone().unwrap_or_else(|| "".into()))
```
TonyGiorgio, line 454: Isn't this just turning it back into `""` anyways? I feel like the issue is more downstream than here.

benthecarman: Yeah but this will make it so the description on what we return is `None`.

Also I have a todo of `/ TODO if there's no description should be something random I guess`, which we'll want to handle if the description is none

TonyGiorgio: I feel that it is better to fix it closer to the source of the problem than above here. So instead of it failing if every call to `create_invoice` is an empty string, have it treat empty string and `None` as the same inside of `create_invoice`, then any call to `create_invoice` won't have the same fate.

benthecarman: Good call, I like this a lot better

TonyGiorgio (approved):
Awesome looks good!

PR #428, opened 2023-04-26
Use mempool.space's fee rate api 

benthecarman:
Mempool.space's fee rate api should be smarter than the generic esplora one. Hopefully this will cause less issues.

Added some tests and cleaned up how we calculate them as well too

part of #423


TonyGiorgio (approved):
Oh cool, I like this. Have you tested it against our signet?

benthecarman:
> Oh cool, I like this. Have you tested it against our signet?

The test I made calls upon the mutinynet api so we're testing that, haven't tried doing any channels or anything yet

Issue #433, opened 2023-04-27
Could not spend spendable outputs panic


TonyGiorgio:
This is probably related to a channel closure that I did on this wallet. 

```
INFO mutiny-core/src/logging.rs:27
2023-04-27 20:57:04.786 INFO  [mutiny_core::esplora:220] Starting transaction sync.
[mutiny_wasm.js:2268:17](http://localhost:3420/@fs/data/Dev/Mutiny/mutiny-node/mutiny-wasm/pkg/mutiny_wasm.js)
DEBUG mutiny-core/src/logging.rs:26
2023-04-27 20:57:05.438 DEBUG [lightning::chain::chainmonitor:612] New best block 0000010118f1840b4559ce14ca2ef491e0a734e14b30a852f5de406b237e9217 at height 36878 provided via best_block_updated
[mutiny_wasm.js:2271:17](http://localhost:3420/@fs/data/Dev/Mutiny/mutiny-node/mutiny-wasm/pkg/mutiny_wasm.js)
DEBUG mutiny-core/src/logging.rs:26
2023-04-27 20:57:05.439 DEBUG [lightning::chain::onchaintx:707] Updating claims view at height 36878 with 0 claim requests
[mutiny_wasm.js:2271:17](http://localhost:3420/@fs/data/Dev/Mutiny/mutiny-node/mutiny-wasm/pkg/mutiny_wasm.js)
DEBUG mutiny-core/src/logging.rs:26
2023-04-27 20:57:05.439 DEBUG [lightning::chain::onchaintx:820] Updating claims view at height 36878 with 0 matched transactions in block 36878
[mutiny_wasm.js:2271:17](http://localhost:3420/@fs/data/Dev/Mutiny/mutiny-node/mutiny-wasm/pkg/mutiny_wasm.js)
DEBUG mutiny-core/src/logging.rs:26
2023-04-27 20:57:05.439 DEBUG [lightning::chain::channelmonitor:3244] Descriptor StaticOutput 532ff37cb6c3cb28b5f3ec5f96a0373c1765ab757a7343a40737cd5f50186110:0 marked for spending has got enough confirmations to be passed upstream
[mutiny_wasm.js:2271:17](http://localhost:3420/@fs/data/Dev/Mutiny/mutiny-node/mutiny-wasm/pkg/mutiny_wasm.js)
DEBUG mutiny-core/src/logging.rs:26
[...]

PR #435, opened 2023-04-28
Filter out StaticOutput spendable outputs

benthecarman:
Part of #433

We can filter out the `StaticOutput`s because these are just the utxos that already went to our BDK wallet, there's no reason to spend them back into our BDK wallet.

### DIFF
--- a/mutiny-core/src/event.rs
+++ b/mutiny-core/src/event.rs
@@ -7,6 +7,7 @@ use crate::wallet::MutinyWallet;
 use bitcoin::hashes::hex::ToHex;
 use bitcoin::secp256k1::PublicKey;
 use bitcoin::secp256k1::Secp256k1;
+use lightning::chain::keysinterface::SpendableOutputDescriptor;
 use lightning::events::{Event, PaymentPurpose};
 use lightning::{
     chain::chaininterface::{ConfirmationTarget, FeeEstimator},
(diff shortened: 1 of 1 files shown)


TonyGiorgio (changes requested):
There's another reason for the panics as well, we are passing an empty vec of outputs.

```
                let spending_tx = self
                    .keys_manager
                    .spend_spendable_outputs(
                        output_descriptors,
                        Vec::new(),
                        tx_feerate,
                        &Secp256k1::new(),
                    )
                    .expect("could not spend spendable outputs");
```

benthecarman:
> There's another reason for the panics as well, we are passing an empty vec of outputs.

That is just any extra outputs we want to add on (like if we wanted to do a payment or something). Since we put an empty vec everything will just be swept to our change address.

TonyGiorgio:
> > There's another reason for the panics as well, we are passing an empty vec of outputs.
> 
> That is just any extra outputs we want to add on (like if we wanted to do a payment or something). Since we put an empty vec everything will just be swept to our change address.

Is that because we provided our change addresses in a `keys_manager` function?

benthecarman:
> Is that because we provided our change addresses in a keys_manager function?

Yes

TonyGiorgio (approved):
Looks good!

PR #436, opened 2023-04-28
Graceful shutdown

TonyGiorgio:
Handles most of the stopping behavior. 

Still need to:
- [x] Test
- [x] Drop the nodes
- [x] Handle starts again
- [x] Stop reading from sockets
- [x] Stop the database connection

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -1,5 +1,5 @@
 use anyhow::anyhow;
-use std::cmp::Ordering;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::{collections::HashMap, ops::Deref, sync::Arc};
 
 use crate::auth::{AuthManager, AuthProfile};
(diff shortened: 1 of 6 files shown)


TonyGiorgio:
Test keeps failing/timing out but I don't think it's from the code, running fine locally.

mutiny-core/src/nodemanager.rs
```diff
@@ -454,6 +455,36 @@ impl NodeManager {
...
+                }
+            }
+        });
+        debug!("stopping all nodes");
+        join_all(node_futures).await;
+        nodes.clear();
+        debug!("stopped all nodes");
+
+        // stop the indexeddb object to close db connection
+        if self.storage.clone().connected().is_ok_and(|b| b) {
```
benthecarman, line 479: Instead of is_ok_and, could do unwrap_or(false)

benthecarman (approved):
I like creating the config object, had a similar idea.

PR #439, opened 2023-04-29
Add ability to archive node

benthecarman:
Will add a `archived` flag to the node storage. if this is on the node won't be started up when we create the node manager.

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -1,3 +1,4 @@
+use anyhow::anyhow;
 use std::cmp::Ordering;
 use std::{collections::HashMap, ops::Deref, sync::Arc};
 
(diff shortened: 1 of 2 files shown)


mutiny-core/src/nodemanager.rs
```diff
@@ -356,9 +364,16 @@ impl NodeManager {
 
         let node_storage = storage.get_nodes()?;
 
+        // Remove the archived nodes, we don't need to start them up.
+        let unarchived_nodes = node_storage
+            .clone()
+            .nodes
+            .into_iter()
+            .filter(|(_, n)| !n.is_archived());
```
TonyGiorgio, line 372: We may want to apply this filter on all of the calls we do in node manager where we look through all the nodes.

TonyGiorgio: Wait, nvm, we only check the node list so as long as unarchived nodes are not added to the node list, then it is fine.

TonyGiorgio (approved):
Not sure if this is good post-hackathon so I'll defer to you for a merge.

PR #440, opened 2023-04-30
Add ability to sweep to a new channel

benthecarman:
Needed for redshifts. We need the ability to sweep a utxo (or set of utxos) to a given channel. LDK makes this difficult where we need to calculate the channel size beforehand by calculating the fee and subtracting from the total value of the selected outputs, then we need to save it so the event manager can fetch it.


TonyGiorgio (approved):

benthecarman:
tested in the redshifts branch, merging

PR #443, opened 2023-04-30
Expose redshift init in wasm

TonyGiorgio:


mutiny-wasm/src/lib.rs
```diff
@@ -485,6 +486,47 @@ impl NodeManager {
...
+    /// Initiates a redshift
+    #[wasm_bindgen]
+    pub async fn init_redshift(
+        &self,
+        outpoint: String,
+        lightning_recipient_pubkey: Option<String>,
+        lightning_recipient_connection_string: Option<String>,
+        onchain_recipient: Option<String>,
+    ) -> Result<JsValue /* Redshift */, MutinyJsError> {
+        let outpoint: OutPoint = OutPoint::from_str(outpoint.as_str()).expect("invalid outpoint");
```
benthecarman, line 498: instead of as_str can just use `&`

PR #449, opened 2023-05-01
Redshifts

benthecarman:
Todos:

- [ ] tests where we can
- [x] successfully do one on signet
- [x] redo storage in not such an ugly way
- [x] better logging
- [x] figure out channel reserve bs
- [ ] handle shutdown while redshifts is attempting payments
- [ ] archive nodes afterwards

### DIFF
--- a/mutiny-core/src/redshift.rs
+++ b/mutiny-core/src/redshift.rs
@@ -7,8 +7,8 @@ use anyhow::anyhow;
 use bitcoin::hashes::hex::ToHex;
 use bitcoin::secp256k1::PublicKey;
 use bitcoin::{Address, OutPoint};
-use lightning::util::logger::Logger;
 use lightning::{log_debug, log_error, log_info};
+use lightning::{log_warn, util::logger::Logger};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
@@ -64,7 +64,9 @@ pub struct Redshift {
     pub recipient: RedshiftRecipient,
     pub output_utxo: Option<OutPoint>,
     pub introduction_channel: Option<OutPoint>,
-    pub output_channel: Option<OutPoint>,
+    /// output_channel will be None if being kept in lightning, this is only relevant when the
+    /// channels will be closed into an on chain address
+    pub output_channel: Option<Vec<OutPoint>>,
     pub introduction_node: PublicKey,
     pub amount_sats: u64,
     pub sats_sent: u64,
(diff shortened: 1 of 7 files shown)


TonyGiorgio:
This commit can be taken out: https://github.com/MutinyWallet/mutiny-node/pull/449/commits/380118dd4fd1846910cbb8135a7c579b4f36ed06

TonyGiorgio:
Can we also remove https://github.com/MutinyWallet/mutiny-node/pull/449/commits/f8ec46bb5c17a3868b272bd477dffd66990dc1a9 & https://github.com/MutinyWallet/mutiny-node/pull/449/commits/3972b581e83faf1efb6855fe21a9659843d84c09 ?

benthecarman:
Removed no longer needed commits

mutiny-core/src/redshift.rs
```diff
@@ -280,6 +280,28 @@ impl RedshiftManager for NodeManager {
...
+                        c.unspendable_punishment_reserve
+                    } else {
+                        None
+                    }
+                })
+                .unwrap_or(0),
+        };
+
+        // original utxo value - opening tx fee - channel reserve
+        let max_sats = rs.amount_sats - rs.fees_paid - reserve;
```
TonyGiorgio, line 301: Are these unsigned? May need to do saturating subtract since your unwraps set it to 0.  How possible is it for those 0's to trigger? Shouldn't we always have this number?

TonyGiorgio (changes requested):

mutiny-core/src/redshift.rs
TonyGiorgio: ah yeah you're right, i was thinking about the `amount_sats` part coming through as `0` but you're right that wouldn't happen.

mutiny-core/src/redshift.rs
```diff
@@ -425,12 +427,6 @@ impl RedshiftManager for NodeManager {
             rs.status
         );
 
-        // fixme, this could close the wrong channel if the receiving_node already had channels
```
benthecarman: ah shit @TonyGiorgio this is still the case after your PR

TonyGiorgio: No, the close channel flow only happens if it's an on chain redshift. So there is no chance that a receiving node already had channels.

benthecarman: Ah I see, so we'll never set that channel outpoint then. Probably okay for now

TonyGiorgio: Yeah, I added this comment for the outpoint itself: 

```
    /// output_channel will be None if being kept in lightning, this is only relevant when the
    /// channels will be closed into an on chain address
```

Realized it isn't really something that matters for the lightning option (and as we determined by matt, a gibberish thing for now that's not possible to determine the channel payments came down)

benthecarman:
Rebased

benthecarman:
Pushed up a couple commits on some cleanups I saw. Otherwise looks good to me

benthecarman:
Something I realized as well, instead of doing the channel reserve calculations we could just use the channel's `outbound_capacity_msat`. The docs indicate it is not exact but it would work better.

https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/ln/channelmanager.rs#L1267

TonyGiorgio (approved):
Added more todo's. Looks good to merge IMO. Will squash and merge when tests pass.

PR #455, opened 2023-05-02
Turn off upfront shutdown pubkey for channels

benthecarman:
When this setting is `true` it generates an address and assigns it to the channel when it is opened. This is not ideal for us when we will be wanting to do things like label every address and could have complications with redshifts.


TonyGiorgio:
So does it get it from our BDK addresses anyways some other way? Via the close procedure or one of the event manager events?

benthecarman:
> So does it get it from our BDK addresses anyways some other way? Via the close procedure or one of the event manager events?

Yeah instead of asking the keymanager for the bdk address on channel open and saving it, it will ask for it on channel close JIT

TonyGiorgio (approved):

PR #457, opened 2023-05-02
Skip phantom invoices if LSP is involved

TonyGiorgio:
LSP helps with privacy and will handle the direct JIT channel open so it's not really necessary and fixes the TODO we had in the redshift PR.


benthecarman (approved):

benthecarman:
We may just want to remove the phantom stuff altogether

TonyGiorgio:
> We may just want to remove the phantom stuff altogether

Yeah, might need to think about that one. The Phantom KeysManager vs normal one aren't interchangable last I remembered. Phantom KeysManager can do things the normal one can but not vice versa, so it depends on if we ever care about phantom, which we might not.

PR #458, opened 2023-05-03
Loop through pays until it succeeds or too small

TonyGiorgio:
This will loop through until the maximum amount of the channel was paid or the amount is too small.

### DIFF
--- a/mutiny-core/src/redshift.rs
+++ b/mutiny-core/src/redshift.rs
@@ -262,7 +262,7 @@ impl RedshiftManager for NodeManager {
     async fn attempt_payments(&self, mut rs: Redshift) -> Result<(), MutinyError> {
         // TODO find the max channel reserve
         let max_sats = (rs.amount_sats as f64 * 0.99) as u64;
-        let _min_sats = 10_000;
+        let min_sats = 10_000;
 
         // get the node making the payment
         let nodes = self.nodes.lock().await;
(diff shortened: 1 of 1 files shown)


mutiny-core/src/redshift.rs
```diff
@@ -273,57 +273,65 @@ impl RedshiftManager for NodeManager {
...
+                Ok(i) => {
+                    if i.paid {
+                        debug!("paid the redshift invoice");
+                        rs.payment_attempted(local_max_sats);
+
+                        // check if the max amount was sent on first try
+                        if local_max_sats == max_sats {
+                            rs.status = RedshiftStatus::Completed;
+                            break;
+                        }
```
benthecarman, line 315: also need an exit condition for if the first fails but a later attempt succeeds

benthecarman: With the strategy you're doing, probably fine to just exit if any payment works

TonyGiorgio: It should now consider the previous amount paid when determining if a successful payment has now maxed out the channel. Otherwise, the amount will keep going down until it hits the min amount.

mutiny-core/src/redshift.rs
```diff
@@ -273,57 +273,70 @@ impl RedshiftManager for NodeManager {
...
+                    if i.paid {
+                        debug!("paid the redshift invoice");
+                        let prev_amount_paid = match rs.status {
+                            RedshiftStatus::AttemptingPayments(x) => x,
+                            _ => 0,
+                        };
+
+                        rs.payment_attempted(local_max_sats);
+
+                        // check if the max amount was sent on first try
```
benthecarman, line 316: this comment is wrong now

benthecarman (approved):
otherwise looks good to me for now

mutiny-core/src/redshift.rs
TonyGiorgio: Fixed, and also changed things to a `saturating_sub` since there was a chance for that to go negative.

benthecarman (approved):

PR #459, opened 2023-05-03
Broadcast tx over nostr

benthecarman:
Implements https://github.com/nostr-protocol/nips/pull/476

For now we still broadcast to esplora as well because I am pretty sure I am the only one running a server.

Need to make the nostr-sdk more wasm safe

### DIFF
--- a/mutiny-core/Cargo.toml
+++ b/mutiny-core/Cargo.toml
@@ -15,7 +15,7 @@ repository = "https://github.com/mutinywallet/mutiny-node"
 [dependencies]
 ln-websocket-proxy = "0.3.0"
 lnurl-rs = { version = "0.2.2", default-features = false, features = ["async", "async-https"] }
-
+nostr-sdk = { version = "0.21.0", default-features = false }
 cfg-if = "1.0.0"
 wasm-bindgen = "0.2.84"
 bip39 = { version = "2.0.0" }
--- a/mutiny-core/src/chain.rs
+++ b/mutiny-core/src/chain.rs
@@ -49,7 +49,7 @@ impl MutinyChain {
         client.connect().await;
 
         let tag = Tag::Generic("magic".into(), vec![network_magic.to_hex()]);
-        let base64_tx = base64::encode(&tx.encode());
+        let base64_tx = base64::encode(tx.encode());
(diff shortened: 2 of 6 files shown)


mutiny-core/Cargo.toml
```diff
@@ -15,7 +15,7 @@ repository = "https://github.com/mutinywallet/mutiny-node"
 [dependencies]
 ln-websocket-proxy = "0.3.0"
 lnurl-rs = { version = "0.2.2", default-features = false, features = ["async", "async-https"] }
-
+nostr-sdk = { version = "0.21.0", default-features = false }
```
TonyGiorgio, line 18: This uses `tokio`, are you sure it's browser safe? Might be wasm safe but idk about `tokio::sync` https://github.com/rust-nostr/nostr/blob/master/crates/nostr-sdk/Cargo.toml#L41

benthecarman: I tested and was able to broadcast a transaction, will look into where tokio is needed

benthecarman: looks like it is used for channels

TonyGiorgio: I didn't think tokio channels were wasm safe. I've just been going with futures.

TonyGiorgio: Nothing I'm seeing online convinces me that it's truly wasm safe. From documentation, I see that even `tokio::sync::mpsc` will block if there's multiple sends that it needs to do. https://github.com/tokio-rs/tokio/blob/c89406965ffb4a64936d781c556a2c48855dfbdb/tokio/src/sync/mpsc/bounded.rs#L448-L451

benthecarman: Okay will convert to draft and we can make PRs into rust-nostr

TonyGiorgio: I could also be wrong, it's just so hard to tell

benthecarman: We could just implement our own client. All we need is a websocket, won't be very hard.

TonyGiorgio:
Let's go without this

PR #460, opened 2023-05-03
Create MutinyManager that holds a node manager

benthecarman:
For the hackathon we were doing weird stuff having `NodeManager::new` return an `Arc` of itself so we can start the redshifts thread. This refactors so we have a `MutinyManager` that holds an `Arc<NodeManager>`, this gives us a little bit better of an abstraction so we can have `NodeManager` functions reference itself without having to do too many weird things.

I image we can one day eventually refactor so `MutinyManager` handles things like the mnemonic, lnurl auth, nostr, etc and keep the `NodeManager to purely just bitcoin and lightning.

### DIFF
--- a/mutiny-wasm/src/lib.rs
+++ b/mutiny-wasm/src/lib.rs
@@ -22,15 +22,14 @@ use lnurl::lnurl::LnUrl;
 use mutiny_core::redshift::RedshiftManager;
 use mutiny_core::{nodemanager, redshift::RedshiftRecipient};
 use std::str::FromStr;
-use std::sync::Arc;
 use wasm_bindgen::prelude::*;
 
 #[wasm_bindgen]
-pub struct NodeManager {
-    inner: Arc<nodemanager::NodeManager>,
+pub struct MutinyManager {
+    inner: mutiny_core::MutinyManager,
 }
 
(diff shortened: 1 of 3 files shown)


mutiny-wasm/src/lib.rs
```diff
@@ -38,8 +37,8 @@ pub struct NodeManager {
 /// It can be configured to use all different custom backend services, or to use the default
 /// services provided by Mutiny.
 #[wasm_bindgen]
-impl NodeManager {
-    /// Creates a new [NodeManager] with the given parameters.
+impl MutinyManager {
```
TonyGiorgio, line 40: Comment above on line 33 still has `NodeManager`, just double check for all of the comments.

TonyGiorgio (approved):
Definitely a lot of interesting ways we can go about it over time, probably just going to have to play it by ear and keep everything as clean as possible to allow for abstractions, merges, interactions, etc. going forward. I do think this is in the right direction, though we should probably bump to 0.3 on the next release and put it in the release notes.

mutiny-wasm/src/lib.rs
benthecarman: could catch, didn't find anymore

benthecarman: nvm found a test that i renamed

PR #462, opened 2023-05-03
Address labeling

### LINKED ISSUE #312: Add address labeling
Would be also good to add BIP 329 support

benthecarman:
Closes #312

Start of address labeling. Labels are assigned to addresses and when we spend we take the labels of the inputs and assign them to our outputs.

Contacts are labels with extra info. There is a way to convert a label into a contact or we also have the ability to create a fresh contract where we will generate a uuid as the label.

TODOs

- [x] use vec of labels instead of comma separated strings everywhere
- [x] assign new label to tx inputs
- [x] tests
- [x] expose a way to view and create labels/contacts to mutiny-wasm

### DIFF
--- a/mutiny-core/src/labels.rs
+++ b/mutiny-core/src/labels.rs
@@ -1,0 +1,700 @@
+use crate::error::MutinyError;
+use crate::indexed_db::MutinyStorage;
+use crate::nodemanager::NodeManager;
+use bitcoin::{Address, XOnlyPublicKey};
+use lightning_invoice::Invoice;
+use lnurl::lightning_address::LightningAddress;
+use lnurl::lnurl::LnUrl;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use uuid::Uuid;
+
+const ADDRESS_LABELS_MAP_KEY: &str = "address_labels";
+const INVOICE_LABELS_MAP_KEY: &str = "invoice_labels";
+const LABEL_PREFIX: &str = "label/";
+const CONTACT_PREFIX: &str = "contact/";
+
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, Ord, PartialEq, PartialOrd, Hash, Default)]
+pub struct LabelItem {
+    /// List of addresses that have this label
+    pub addresses: Vec<Address>,
+    /// List of invoices that have this label
+    pub invoices: Vec<String>, // fixme: use Invoice type after https://github.com/lightningdevkit/rust-lightning/pull/2279
+    /// Epoch time in seconds when this label was last used
+    pub last_used_time: u64,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, Ord, PartialEq, PartialOrd, Hash)]
+pub struct Contact {
+    pub name: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub npub: Option<XOnlyPublicKey>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub ln_address: Option<LightningAddress>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub lnurl: Option<LnUrl>,
+    pub last_used: u64,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, Ord, PartialEq, PartialOrd, Hash)]
+pub enum TagItem {
+    Label((String, LabelItem)),
+    Contact((String, Contact)),
+}
+
+fn get_label_item_key(label: impl AsRef<str>) -> String {
+    format!("{}{}", LABEL_PREFIX, label.as_ref())
+}
+
+fn get_contact_key(label: impl AsRef<str>) -> String {
+    format!("{}{}", CONTACT_PREFIX, label.as_ref())
+}
+
+pub trait LabelStorage {
+    /// Get a map of addresses to labels. This can be used to get all the labels for an address
+    fn get_address_labels(&self) -> Result<HashMap<Address, Vec<String>>, MutinyError>;
+    /// Get a map of invoices to labels. This can be used to get all the labels for an invoice
+    fn get_invoice_labels(&self) -> Result<HashMap<Invoice, Vec<String>>, MutinyError>;
+    /// Get all the existing labels
+    fn get_labels(&self) -> Result<HashMap<String, LabelItem>, MutinyError>;
+    /// Get information about a label
+    fn get_label(&self, label: impl AsRef<str>) -> Result<Option<LabelItem>, MutinyError>;
+    /// Set the labels for an address, replacing any existing labels
+    /// If you do not want to replace any existing labels, use `get_address_labels` to get the existing labels,
+    /// add the new labels, and then use `set_address_labels` to set the new labels
+    fn set_address_labels(&self, address: Address, labels: Vec<String>) -> Result<(), MutinyError>;
+    /// Set the labels for an invoice, replacing any existing labels
+    /// If you do not want to replace any existing labels, use `get_invoice_labels` to get the existing labels,
+    /// add the new labels, and then use `set_invoice_labels` to set the new labels
+    fn set_invoice_labels(&self, invoice: Invoice, labels: Vec<String>) -> Result<(), MutinyError>;
+    /// Get all the existing contacts
+    fn get_contacts(&self) -> Result<HashMap<String, Contact>, MutinyError>;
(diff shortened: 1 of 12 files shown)


mutiny-wasm/src/models.rs
```diff
@@ -338,3 +342,90 @@ impl AuthProfile {
         JsValue::from_serde(&serde_json::to_value(&self.used_services).unwrap()).unwrap()
     }
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, Eq, PartialEq, Hash)]
+#[wasm_bindgen]
+pub enum TagKing {
```
futurepaul, line 348: TagKind? or the king of the tags

benthecarman: I also misspell kind as king, I have no idea why.

mutiny-wasm/src/lib.rs
```diff
@@ -542,6 +551,70 @@ impl MutinyWallet {
...
 
+    pub fn get_address_labels(
+        &self,
+    ) -> Result<JsValue /* Map<Address, Vec<String>> */, MutinyJsError> {
+        Ok(JsValue::from_serde(
+            &self.inner.node_manager.get_address_labels()?,
+        )?)
+    }
+
+    pub fn label_address(
```
futurepaul, line 562: unclear from this name if this method overwrites the address labels or extends? maybe `set_address_labels` ?

mutiny-wasm/src/models.rs
```diff
@@ -338,3 +342,90 @@ impl AuthProfile {
...
+    Label,
+    Contact,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
+#[wasm_bindgen]
+pub struct TagItem {
+    id: String,
+    pub kind: TagKing,
```
futurepaul, line 357: fwiw I really appreciate this kind of enum in js land... a property of a big object with a bunch of options, rather than multiple object types. I know it makes more work in rust land but it's easier for me to work with

futurepaul (commented):
I won't comment on all the under the hood stuff, but just to be clear I understand how this will work from my perspective:

I can list out all the `TagItem`s, I can create a a `Contact`, I can convert a `Label` into a `Contact`... where do I create just a `Label`? is that in `label_address`?

mutiny-wasm/src/models.rs
benthecarman: Yeah that was my goal, made the nice enum in our internal stuff but struct with lots of `Option`s for the wasm crate.

benthecarman:
> I can list out all the TagItems, I can create a a Contact, I can convert a Label into a Contact... where do I create just a Label? is that in label_address?

So my thinking is that there really isn't anywhere where you would create a label, you would only be creating them when sending or receiving, so you pass them in when calling those functions. I guess you could also create them by calling `label_address` but my think would be that would only be for editing a label of an address.

mutiny-wasm/src/lib.rs
benthecarman: Good call, changed name and added docs.

TonyGiorgio (changes requested):
The tight coupling of contacts and labels have definitely confused me as to what is going on. IMO it would be ideal if they were seperated and one did not have the reliance on the other.

IE, Contacts can and probably should be a seperate thing on their own, referenced by a UUID, with no labeling logic around them. Labels can select a contact ID to use for a label, or a human readable string, and can apply to addresses, invoices, or channels.

As is, this seems like it would be hard to keep evolving contacts while it is intertwined with labeling.

benthecarman:
@TonyGiorgio My thinking was to have contacts be independent of labeling in all ways but their id. This way all of our labeling logic handles contacts the same way. Otherwise we need to have something like `label_psbt` for contacts as well and any other label like system we have. This lets it be treated all the same and contacts are just labels with extra info. Imo this is the best way to do it because we want to treat contacts as labels so users tag their transactions appropriately. If we separate the two it will make us have to duplicate a lot of code and logic as well as remove the main benefit from users using the contacts system (tricking them into properly labeling).

mutiny-core/src/labels.rs
```diff
@@ -0,0 +1,492 @@
...
+        Ok(())
+    }
+
+    /// Create a new contact and return the identifying label
+    fn create_new_contact(&self, contact: Contact) -> Result<String, MutinyError> {
+        // generate a uuid, this will be the "label" that we use to store the contact
+        let id = Uuid::new_v4().to_string();
+        let key = get_contact_key(&id);
+        self.set(key, contact)?;
+        Ok(id)
```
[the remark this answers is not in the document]
benthecarman: Both, labels are just a String that a user assigns to a tx, address, etc. For contacts we aren't going to be displaying the label and instead displaying the contact info so we use a `uuid` to guarantee uniqueness.

mutiny-core/src/labels.rs
```diff
@@ -0,0 +1,492 @@
...
+use std::collections::HashMap;
+use uuid::Uuid;
+
+const LABELS_MAP_KEY: &str = "labels";
+const LABEL_PREFIX: &str = "label/";
+const CONTACT_PREFIX: &str = "contact/";
+
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, Ord, PartialEq, PartialOrd, Hash)]
+pub struct LabelItem {
+    pub addresses: Vec<Address>,
```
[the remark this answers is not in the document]
benthecarman: Yeah I only did on-chain for this PR because that will be the hardest. With invoices we don't have to worry about labels transfering from a previous invoice to the next so figured that can be a followup.

mutiny-core/src/labels.rs
```diff
@@ -0,0 +1,492 @@
...
+
+    fn set_address_labels(&self, address: Address, labels: Vec<String>) -> Result<(), MutinyError> {
+        // update the labels map
+        let mut address_labels = self.get_address_labels()?;
+        address_labels.insert(address.clone(), labels.clone());
+        self.set(LABELS_MAP_KEY, address_labels)?;
+
+        // update the label items
+        let now = crate::utils::now().as_secs();
+        for label in labels {
```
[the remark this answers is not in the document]
benthecarman: This is just the nature of a key-value store, open to suggestions to make this better. We want two things, a way to get a label and see all the addresses associated with it, a way to get an address and get all the labels associated with it. If we were using sql we could use joins to make this pretty simple and efficient but we don't have that luxury here. So instead I made a Map of addressees to labels so we can easily check and lookup a set of labels for a given address or transaction. Then also added to storage a rows for individual addresses and their list of associated addresses. That way we can do the inverse and lookup a label without having to scan through all the addresses. I also have a fear about the flaklyness of this but not really sure a better model to fix it that still lets us have fast lookups.

All of this has nothing to do with contacts, contacts are just a mapped to a label that has extra info. Nothing in here should touch contacts.

mutiny-core/src/labels.rs
```diff
@@ -0,0 +1,492 @@
...
+                .into_iter()
+                .map(|(label, label_item)| TagItem::Label((label, label_item))),
+        );
+
+        // Convert the contacts into tag items
+        tag_items.extend(
+            contacts
+                .into_iter()
+                .map(|(id, c)| TagItem::Contact((id, c))),
+        );
```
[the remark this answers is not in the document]
benthecarman: >  How can a label, which is only supposed to reference an on chain address, also be a contact

Like I said below but a contact is only useful if is a label so it can be referenced throughout a list of transactions so users can see where funds are coming from/going to. Contacts use labels as their identifiers so we can get all the nice features of labels without having to re-implement it.

> How is a contact an on chain address?

That never happens.

mutiny-core/src/labels.rs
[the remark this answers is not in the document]
TonyGiorgio: Yeah key values make things tricky. Still need to be able to handle the case where lists are overwritten and it go back and update the references, and probably write a test checking that as well.

TonyGiorgio:
Only thing I would like to see is a fix / test around the two lists being different if there's an overwrite but feel free to fix in a follow up PR if it's blocking progress on labeling stuff.

benthecarman:
> Only thing I would like to see is a fix / test around the two lists being different if there's an overwrite but feel free to fix in a follow up PR if it's blocking progress on labeling stuff.

Added better tests and made it more robust

TonyGiorgio:
Great tests! Hard to vet it all but looks good to me!

TonyGiorgio (approved):

PR #464, opened 2023-05-04
Reconnection improvements

### LINKED ISSUE #461: P2P messages producing errors
Been getting lots of errors in my logs, I think it has to do with multinode stuff?

I get lots of peer sent invalid data logs

TonyGiorgio:
Fixes #461 

Two connections to the LSP were being initiated at once. First from the hardcoded init and then from the reconnection thread. This just makes the reconnection thread wait 5 seconds before handling.

It's going to be hard to make sure races do not happen while a `connect_peer` has been made and when it is finished. LDK doesn't consider it finished until all the handshaking has occured. So multiple handshakes can happen at once.

### DIFF
--- a/mutiny-core/src/node.rs
+++ b/mutiny-core/src/node.rs
@@ -1043,6 +1043,33 @@ async fn start_reconnection_handling(
     uuid: String,
     lsp_client: &Option<LspClient>,
 ) -> Result<(), MutinyError> {
+    // Attempt connection to LSP first
+    if let Some(lsp) = lsp_client.clone() {
+        let node_id = NodeId::from_pubkey(&lsp.pubkey);
+
+        match connect_peer_if_necessary(
+            multi_socket.clone(),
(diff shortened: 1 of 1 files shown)


mutiny-core/src/node.rs
```diff
@@ -1148,6 +1034,139 @@ impl Node {
...
+                        sleep(5 * 1000).await;
+                        continue;
+                    }
+                };
+            }
+            sleep(5 * 1000).await;
+        }
+    });
+
+    // Attempt connection to LSP first
```
benthecarman, line 1067: you have here we are connecting to the LSP first but we're actually connecting to the m2m proxy first. Would probably make sense to actually do the LSP first so that is ready as soon as possible.

TonyGiorgio: Moved to the top

benthecarman (approved):

PR #465, opened 2023-05-04
Fix deadlock issues and close channel properly

TonyGiorgio:


benthecarman (approved):
Ah, clippy has warned me about this somewhere else before but didn't here so figured it was okay. Good catch!

PR #468, opened 2023-05-05
Add utility function `get_node`

benthecarman:
This should help prevent things like #465, making it easier to get a node without having to be as worried about the locks.

I think this fixes something where with lightning payments we were locking the nodes the entire timeout duration.


TonyGiorgio (approved):
Good idea

PR #469, opened 2023-05-05
Refactor redshift storage

benthecarman:
Refactors storage to have the key be the id instead of having the utxo be the key. This will allow multi-utxo redshifts and we can do some nice optimizations. In the `start_redshifts()` thread before we were looking at every redshift and then checking if the channel is ready, now we can just look at our couple active channels and see if any are a redshift, this should be more efficient I think.

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -456,7 +456,6 @@ impl NodeManager {
     }
 
     pub(crate) fn start_redshifts(nm: Arc<NodeManager>) {
-        let node_manager = nm.clone();
         spawn_local(async move {
             loop {
                 // find redshifts with channels ready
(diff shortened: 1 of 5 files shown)


benthecarman:
Going to add tests now that I redid the storage

mutiny-core/src/nodemanager.rs
```diff
@@ -455,39 +455,41 @@ impl NodeManager {
         Ok(nm)
     }
 
-    pub(crate) fn start_redshifts(nm: Arc<NodeManager>) {
-        let node_manager = nm.clone();
```
benthecarman, line 459: very confused on how i was able to remove this, will do a test redshift to make sure it still works

TonyGiorgio (approved):
Good idea on the ID change

benthecarman:
Found bug, do not merge

benthecarman:
Pushed up some fixes:

- before if a payment failed, it would handle it like it was pending, this ended up blowing up my db with a ton of invoices. Made it so we instead throw an error in pay_invoice_with_timeout when the payment fails.
- properly accounted for routing fees in lightning payments
- set status to completed when it is done without the full payment amount.

mutiny-core/src/redshift.rs
```diff
@@ -349,6 +339,9 @@ impl RedshiftManager for NodeManager {
                         local_max_sats,
                     );
                     rs.fail("exhausted all payment attempt amounts".to_string());
+                } else {
+                    // todo save amount paid somewhere
+                    rs.status = RedshiftStatus::Completed;
```
TonyGiorgio, line 344: Could we simply just do `RedshiftStatus::Completed(u64)` here?

mutiny-core/src/redshift.rs
```diff
@@ -427,13 +421,21 @@ impl RedshiftManager for NodeManager {
...
         }
 
+        log_debug!(
+            &self.logger,
+            "Redshift {} completed with status: {:?}",
+            rs.id.to_hex(),
+            rs.status
+        );
+
+        // fixme, this could close the wrong channel if the receiving_node already had channels
```
TonyGiorgio, line 431: ah shit nice one, forgot it's not always a spun up node

TonyGiorgio (approved):
Just a small comment on a todo, can ignore it for a full up pr if you want, looks good otherwise

PR #472, opened 2023-05-06
Cleanup logging

benthecarman:
Before we were manually creating the log record and just hardcoding some of the params like the line it occurred on and what exact file. The ldk macro will do all that for us and makes the code much cleaner.


TonyGiorgio (approved):
ah my eyes are so much happier. do we do this everywhere now?

benthecarman:
cleaned up for the rest of the repo, only places it still exists is in the background processor and socket. Those are probably fine for now.

TonyGiorgio (approved):
This is great, thank you! Should make it easier to transition the wasm logger out of this when the time comes.

PR #473, opened 2023-05-08
Close channels if there's multiple from a redshift

TonyGiorgio:


mutiny-core/src/redshift.rs
```diff
@@ -443,17 +439,22 @@ impl RedshiftManager for NodeManager {
...
+                // close all the channels that were opened on the receiving node, if the receiving
+                // node was only created temporarily for the purpose of being thrown away
+                // record all of the channel outpoints.
+                let mut channel_outpoints: Vec<OutPoint> = vec![];
+                for c in receiving_node.channel_manager.list_channels() {
+                    if let Some(funding_txo) = c.funding_txo {
+                        let channel_outpoint = funding_txo.into_bitcoin_outpoint();
+                        self.close_channel(&channel_outpoint).await?;
+                        channel_outpoints.push(channel_outpoint);
+                    }
```
benthecarman, line 455: could join all these futures

benthecarman: we should also keep the log if there's no channel

benthecarman (approved):

PR #479, opened 2023-05-09
Log to storage and clean up tests

TonyGiorgio:
Stores the last 1k lines.

### DIFF
--- a/mutiny-core/src/keymanager.rs
+++ b/mutiny-core/src/keymanager.rs
@@ -257,7 +257,6 @@ mod tests {
         let second_pubkey_again = pubkey_from_keys_manager(&km);
 
         assert_eq!(second_pubkey, second_pubkey_again);
-
-        cleanup_test();
+        cleanup_all().await;
     }
 }
--- a/mutiny-core/src/logging.rs
+++ b/mutiny-core/src/logging.rs
@@ -1,9 +1,102 @@
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
+
+use crate::utils::Mutex;
+use crate::{error::MutinyError, utils::sleep};
 use chrono::Utc;
+use gloo_utils::format::JsValueSerdeExt;
 use lightning::util::logger::{Level, Logger, Record};
 use log::*;
+use rexie::{ObjectStore, Rexie, TransactionMode};
+use wasm_bindgen::JsValue;
+use wasm_bindgen_futures::spawn_local;
 
-#[derive(Default, Debug, Eq, PartialEq, Copy, Clone)]
-pub struct MutinyLogger {}
+pub(crate) const LOGGING_DATABASE_NAME: &str = "logging";
+pub(crate) const LOGGING_OBJECT_STORE_NAME: &str = "log_store";
+pub(crate) const LOGGING_KEY: &str = "logs";
+
+const MAX_LOG_ITEMS: usize = 1_000;
+
+#[derive(Clone)]
+pub struct MutinyLogger {
+    should_write_to_storage: bool,
+    memory_logs: Arc<Mutex<Vec<String>>>,
+}
+
+impl MutinyLogger {
+    pub(crate) fn with_writer(stop: Arc<AtomicBool>) -> Self {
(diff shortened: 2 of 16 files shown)


mutiny-core/src/keymanager.rs
```diff
@@ -257,7 +257,5 @@ mod tests {
         let second_pubkey_again = pubkey_from_keys_manager(&km);
 
         assert_eq!(second_pubkey, second_pubkey_again);
-
-        cleanup_test();
```
benthecarman, line 261: this should probably be `cleanup_all`

mutiny-core/src/logging.rs
```diff
@@ -1,9 +1,90 @@
...
+
+            loop {
+                // wait up to 5s, checking graceful shutdown check each 1s.
+                for _ in 0..5 {
+                    if stop.load(Ordering::Relaxed) {
+                        logging_db.close();
+                        return;
+                    }
+                    sleep(1_000).await;
+                }
```
TonyGiorgio, line 49: I'll probably go back and do this everywhere in a follow up PR. Slims down on the stops taking too long.

mutiny-core/src/test_utils.rs
```diff
@@ -10,19 +10,25 @@ macro_rules! log {
...
 }
 
 pub async fn cleanup_wallet_test() {
-    cleanup_test();
     MutinyStorage::clear().await.unwrap();
 }
+
+pub async fn cleanup_logging_test() {
+    logging::clear().await.unwrap();
+}
```
benthecarman, line 26: should make these private so we make sure we always call `cleanup_all` just to be safe

mutiny-core/src/logging.rs
```diff
@@ -30,3 +119,156 @@ impl Logger for MutinyLogger {
...
+
+        #[cfg(feature = "ignored_tests")]
+        assert!(logger
+            .get_logs()
+            .await
+            .unwrap()
+            .unwrap()
+            .first()
+            .unwrap()
+            .contains(log_str));
```
benthecarman, line 269: instead of using `first()` why not use a `find` so even the other tests won't break it.

mutiny-core/src/logging.rs
```diff
@@ -1,9 +1,90 @@
...
+use wasm_bindgen::JsValue;
+use wasm_bindgen_futures::spawn_local;
+
+pub(crate) const LOGGING_DATABASE_NAME: &str = "logging";
+pub(crate) const LOGGING_OBJECT_STORE_NAME: &str = "log_store";
+pub(crate) const LOGGING_KEY: &str = "logs";
+
+#[derive(Default, Clone)]
+pub struct MutinyLogger {
+    should_write: bool,
```
benthecarman, line 21: `should_write` is a little ambiguous, something like `should_write_to_storage` would be more clear

mutiny-core/src/logging.rs
```diff
@@ -30,3 +119,156 @@ impl Logger for MutinyLogger {
...
+
+    let mut store = transaction.store(LOGGING_OBJECT_STORE_NAME)?;
+
+    // get the existing data so we can append to it, trimming if needed
+    let mut existing_logs = get_logging_data_with_transaction_store(&mut store)
+        .await?
+        .unwrap_or_default();
+    existing_logs.append(recent_logs);
+    if existing_logs.len() > 1_000 {
+        let start_index = existing_logs.len() - 1000;
```
benthecarman, line 194: should make `1000` a const at the top so it is easy to reference and gives the number more context

TonyGiorgio:
Updated with fixes

mutiny-core/src/logging.rs
```diff
@@ -1,9 +1,92 @@
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc, Mutex,
```
benthecarman, line 3: this is using the syncronous mutex, shouldn't it use the `crate::utils::Mutex`?

TonyGiorgio: there's so many `Mutex`'s I never know which one to use

benthecarman (approved):

PR #483, opened 2023-05-09
Use ldk's get_secure_random_bytes() instead of getrandom()

benthecarman:
I trust ldk's implementation more than the `getrandom` library and it cleans up the code a little.


benthecarman:
No, we just call the default key manager's version

https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/sign/mod.rs#L777

PR #485, opened 2023-05-09
Segregate testing databases for concurrency runs

TonyGiorgio:
This allows a prefix to be added to the DB builders with the enforcement of setting a prefix under `test` instances of the node manager / mutiny wallet. Does not allow setting a prefix when not under test.

Allowed to remove all the cleanup code and fixed the specific logging test I was doing. Ran a few times locally, wondering how it behaves in CI. I am going to run it a few times as part of this PR.

### DIFF
--- a/mutiny-core/src/auth.rs
+++ b/mutiny-core/src/auth.rs
@@ -216,7 +216,7 @@ impl AuthManager {
 #[cfg(test)]
 mod test {
     use crate::keymanager::generate_seed;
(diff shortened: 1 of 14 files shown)


benthecarman (commented):
Not really the biggest fan of this, moving storage to be runtime agnostic would fix this better more likely.

However, we may want this for testing stuff in the wasm side even after we do that, so I guess ACK

benthecarman (approved):

PR #491, opened 2023-05-10
Put labels in MutinyInvoice and transactionDetails

benthecarman:


mutiny-core/src/nodemanager.rs
```diff
@@ -739,15 +797,65 @@ impl NodeManager {
...
+                            address_labels.get(&addr).cloned()
+                        } else {
+                            None
+                        }
+                    })
+                    .unwrap_or_default();
+
+                TransactionDetails {
+                    labels,
+                    ..tx.into()
```
TonyGiorgio, line 823: This should be extracted so that both `list_onchain` and `get_transaction` use it. Identical code.

mutiny-core/src/nodemanager.rs
```diff
@@ -739,15 +797,65 @@ impl NodeManager {
...
+        let mut txs = self.wallet.list_transactions(true)?;
         txs.sort();
+        let address_labels = self.get_address_labels()?;
+        let txs = txs
+            .into_iter()
+            .map(|tx| {
+                let labels = tx
+                    .transaction
+                    .clone()
+                    .unwrap()
```
TonyGiorgio, line 809: A comment on this `unwrap` would be nice, knowing that it's only safe when `list_transactions` is true.

mutiny-core/src/nodemanager.rs
```diff
@@ -1109,7 +1221,9 @@ impl NodeManager {
                 let msats = amount_sats * 1000;
                 let invoice = self.lnurl_client.get_invoice(&pay, msats).await?;
 
-                self.pay_invoice(from_node, &invoice.invoice(), None).await
+                let label = "LNURL Pay".to_string();
```
TonyGiorgio, line 1224: Might want to expose `label` for `lnurl_pay` instead so we label the invoice like any other.

mutiny-core/src/nodemanager.rs
```diff
@@ -1757,12 +1876,17 @@ mod tests {
...
-        let actual =
-            MutinyInvoice::from(payment_info, PaymentHash(payment_hash.into_inner()), false)
-                .unwrap();
+        let actual = MutinyInvoice::from(
+            payment_info,
+            PaymentHash(payment_hash.into_inner()),
+            false,
+            vec![],
+        )
+        .unwrap();
```
TonyGiorgio, line 1889: Adding another test with some labels here would be nice.

mutiny-core/src/redshift.rs
```diff
@@ -379,7 +379,7 @@ impl RedshiftManager for NodeManager {
 
             // make attempts to pay it
             match sending_node
-                .pay_invoice_with_timeout(&invoice, None, None)
+                .pay_invoice_with_timeout(&invoice, None, None, vec![])
```
TonyGiorgio, line 382: Could use a todo comment here mentioning that we should add labels for redshift.

TonyGiorgio (changes requested):

mutiny-core/src/nodemanager.rs
```diff
@@ -736,18 +795,57 @@ impl NodeManager {
...
+    fn add_onchain_labels(
+        &self,
+        address_labels: HashMap<Address, Vec<String>>,
+        tx: bdk::TransactionDetails,
+    ) -> TransactionDetails {
+        // find the first output address that has a label
+        let labels = tx
+            .transaction
+            .clone()
+            .unwrap() // safe because we call with list_transactions(true)
```
TonyGiorgio, line 807: Ugh, well now this isn't always going to be the case... A comment on the function itself might be necessary, or safely handle this and don't add any labels.

TonyGiorgio (changes requested):

TonyGiorgio (approved):

PR #494, opened 2023-05-10
Peer reconnection improvements

TonyGiorgio:
- Refactors the sleeping and stopping logic
- Exponential backoffs
- Checks if wss is actually connected before returning successfully

### DIFF
--- a/mutiny-core/src/node.rs
+++ b/mutiny-core/src/node.rs
@@ -31,6 +31,7 @@ use bitcoin::hashes::{hex::ToHex, sha256::Hash as Sha256};
 use bitcoin::secp256k1::rand;
 use bitcoin::{hashes::Hash, secp256k1::PublicKey, Network, OutPoint};
 use core::time::Duration;
+use futures::lock::Mutex;
 use lightning::chain::chaininterface::{ConfirmationTarget, FeeEstimator};
 use lightning::ln::channelmanager::RecipientOnionFields;
 use lightning::{
@@ -159,8 +160,9 @@ pub(crate) struct Node {
     wallet: Arc<OnChainWallet>,
     logger: Arc<MutinyLogger>,
     websocket_proxy_addr: String,
-    multi_socket: MultiWsSocketDescriptor,
+    multi_socket: Arc<Mutex<MultiWsSocketDescriptor>>,
     pub(crate) lsp_client: Option<LspClient>,
+    stop: Arc<AtomicBool>,
 }
 
 impl Node {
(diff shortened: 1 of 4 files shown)


mutiny-core/src/proxy.rs
```diff
@@ -46,6 +46,31 @@ impl WsProxy {
...
+            State::Open => {}
+            _ => return Err(MutinyError::ConnectionFailed),
+        }
+
+        // TODO wait until we get an OK response from websocket.
+        // A connection to the proxy for connections just means that
+        // it just connected to the proxy. It does not mean the proxy
+        // successfully connected out to the other end. They may be
+        // offline and shortly cut off from the WS but that happens
+        // outside of the connect flow. This will falsely return success.
```
TonyGiorgio, line 72: I will do this in a follow up when I have some more time. Essentially I need to update the ln-websocket-proxy and add a new `Command` type that returns `OK` once it successfully connected out to the other peer.

TonyGiorgio:
Valid failure I'll try to fix tomorrow. I'll probably have to pass stop all the way through this too given the sleep loop.

mutiny-core/src/node.rs
```diff
@@ -409,19 +411,37 @@ impl Node {
             connection_type: ConnectionType::Mutiny(websocket_proxy_addr.to_string()),
             original_connection_string: format!("mutiny:{pubkey}@{websocket_proxy_addr}"),
         };
-        let main_proxy = WsProxy::new(
+        // TODO this works but I want to move this to the background due to proxy connection could
+        // take a little bit to actually start up, and I think I can get rid of some of the unwraps
+        // in the MultiWsSocketDescriptor too
+        let ws_proxy_result = WsProxy::new(
```
TonyGiorgio, line 417: Do not merge, I'm going to handle this tomorrow.

TonyGiorgio:
Finally ready for review @benthecarman. 

The follow up / TODO I have above is still something I would rather punt on for now, but would be a minor improvement. It appears things restart / recover just fine if the other peer isn't actually there, but yeah ideally it "fails" connection peer logic instead of fails very shortly on.

mutiny-core/src/node.rs
```diff
@@ -1105,133 +1102,170 @@ async fn start_reconnection_handling(
...
-                    Err(_) => {
-                        sleep(5 * 1000).await;
-                        continue;
+                    Err(e) => {
+                        log_error!(
+                            reconnection_logger,
+                            "could not create new multi socket proxy: {e}",
+                        );
+                        // double the delay if reconnection fails, but do not exceed 60 seconds
+                        delay = (delay * 2).min(60);
```
benthecarman, line 1224: should make `5` and `60` `const`s for easier reference, these values are used in a bunch of places, would add more context and make them easier to update

benthecarman (commented):
I've never really understood our networking code so can't review this too much. Just one small comment otherwise, didn't see any other code smells so lgtm

mutiny-core/src/node.rs
TonyGiorgio: Good suggestion, added.

benthecarman: think you forgot to push

benthecarman (approved):

PR #495, opened 2023-05-10
Add wasm constructor for Contact

TonyGiorgio:


mutiny-wasm/src/models.rs
```diff
@@ -562,6 +565,30 @@ pub struct Contact {
 
 #[wasm_bindgen]
 impl Contact {
+    #[wasm_bindgen(constructor)]
+    pub fn new(
+        name: String,
+        npub: Option<String>,
+        ln_address: Option<String>,
+        lnurl: Option<String>,
+        last_used: u64,
```
benthecarman, line 574: We should exclude last used and just call utils::now() for them

benthecarman (approved):

PR #496, opened 2023-05-10
Add test for labeling on-chain tx

benthecarman:


TonyGiorgio (approved):

mutiny-core/src/nodemanager.rs
```diff
@@ -812,7 +821,7 @@ impl NodeManager {
             .output
             .iter()
             .find_map(|o| {
-                if let Ok(addr) = Address::from_script(&o.script_pubkey, self.network) {
+                if let Ok(addr) = Address::from_script(&o.script_pubkey, used_network) {
```
TonyGiorgio, line 824: This happens in two more places in `onchain.rs`, is that needed there?

benthecarman: No, those are correct. This is only a problem here because they are coming from parsing json. In `onchain.rs` we want to use the proper network so we can compare.

benthecarman: wait actually we might want to do it

benthecarman: I am just gonna make address labels use a string

TonyGiorgio (approved):

PR #498, opened 2023-05-10
Use TagItem struct everywhere

### LINKED ISSUE #497: Create nice structs for `TagItem` in mutiny-wasm
Right now we just write our enum into json which doesn't really play well with javascript. We can convert this into an easier to work with struct.

benthecarman:
Closes #497

turns out i already had the struct, just missed a few places

### DIFF
--- a/mutiny-core/src/labels.rs
+++ b/mutiny-core/src/labels.rs
@@ -267,6 +267,13 @@ impl LabelStorage for MutinyStorage {
         let id = Uuid::new_v4().to_string();
         let key = get_contact_key(&id);
         self.set(key, contact)?;
+
+        let key = get_label_item_key(&id);
+        let label_item = LabelItem {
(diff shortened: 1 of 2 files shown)


futurepaul (approved):
worked on my machine

TonyGiorgio (approved):

mutiny-core/src/labels.rs
```diff
@@ -267,19 +267,24 @@ impl LabelStorage for MutinyStorage {
...
         let mut tag_items = vec![];
 
         // Get all the contacts
-        let mut contacts = self.get_contacts()?;
+        let contacts = self.get_contacts()?;
         // Get all the labels
         let mut labels = self.get_labels()?;
 
-        // filter out contacts that have a label
-        contacts.retain(|label, _| labels.get(label).is_none());
```
benthecarman, line 282: if you revert this the test will now fail

PR #500, opened 2023-05-11
Label paid invoices

benthecarman:


futurepaul (approved):
worked for me. we can handle keysends in a bit

mutiny-core/src/node.rs
```diff
@@ -795,6 +796,10 @@ impl Node {
             )
         };
 
+        self.persister
+            .storage
+            .set_invoice_labels(invoice.clone(), labels)?;
```
TonyGiorgio, line 801: Failing to save labels should not fail the payment that has already been attempted.

It would be nice if labels were just saved to the invoice db but probably too late at this point.

TonyGiorgio (changes requested):

TonyGiorgio (approved):

Issue #501, opened 2023-05-11
Fee estimation


futurepaul:
For on-chain and lightning it would be nice to have a fee estimate, but especially rn for on-chain. @TonyGiorgio said it might require reworking send flow a bit to have a psbt I can review.

blocking https://github.com/MutinyWallet/mutiny-web/issues/57

benthecarman:
Onchain fee estimation done, lightning is harder

PR #504, opened 2023-05-13
Add ability to estimate on-chain tx fee

benthecarman:
Part of #501

Needed to move it so we don't label when creating the PSBT so we don't label just for estimating a psbt. Otherwise was pretty straight forward. Something that I am unsure about is how coin selection is done by bdk, if it is random then the fee estimation could be wrong, probably fine to punt on for now.

### DIFF
--- a/mutiny-core/src/onchain.rs
+++ b/mutiny-core/src/onchain.rs
@@ -3,6 +3,7 @@ use std::str::FromStr;
 use std::sync::{Arc, RwLock};
 
 use bdk::chain::{BlockId, ConfirmationTime};
+use bdk::psbt::PsbtUtils;
 use bdk::template::DescriptorTemplateOut;
 use bdk::{FeeRate, LocalUtxo, SignOptions, TransactionDetails, Wallet};
 use bdk_esplora::{esplora_client, EsploraAsyncExt};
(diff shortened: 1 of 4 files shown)


mutiny-core/src/onchain.rs
```diff
@@ -349,9 +346,20 @@ impl OnChainWallet {
         log_debug!(self.logger, "Unsigned PSBT: {psbt}");
         let finalized = wallet.sign(&mut psbt, SignOptions::default())?;
         log_debug!(self.logger, "finalized: {finalized}");
-        self.label_psbt(&psbt, labels)?;
```
TonyGiorgio, line 352: I see why it should change for estimating but why for `spend_utxos_to_output`? Shouldn't that behave just like `sweep` would?

benthecarman: Just moved the labeling logic for fee estimation, no other functionality changes.

TonyGiorgio: But labeling when spending was removed from `spend_utxos_to_output`? Meanwhile labeling when spending was added to `sweep`. None of that seems to rely on any of the fee estimation calls. Just seems a bit inconsistent and unrelated.

benthecarman: My goal was to make it so calls that just make a PSBT don't label, and the calls that create and broadcast the transaction do. So for `spend_utxos_to_output` the labeling happens in `event.rs` because that is where the transaction is finalized and broadcast. Labeling is done in `sweep` now because that's where the final transcation is broadcast as well.

TonyGiorgio: I see, perhaps the name `spend_utxos_to_output` should be changed then as it does not spend.

benthecarman: updated name to `create_sweep_psbt_to_output`

TonyGiorgio (approved):

Issue #507, opened 2023-05-15
Check & Update Channel Details on reload


TonyGiorgio:
Whenever the channel manager loads, we should check and update the channel details for all the channels. Since this gets saved to disk, whenever we need to change any of the defaults (such as the dust exposure config, which we should update as part of this), we need to persist that. 

https://docs.rs/lightning/latest/lightning/ln/channelmanager/struct.ChannelManager.html#method.update_channel_config

PR #511, opened 2023-05-16
New channel config defaults

TonyGiorgio:
Fixes https://github.com/MutinyWallet/mutiny-node/issues/507

### DIFF
--- a/mutiny-core/src/node.rs
+++ b/mutiny-core/src/node.rs
@@ -32,8 +32,11 @@ use bitcoin::secp256k1::rand;
 use bitcoin::{hashes::Hash, secp256k1::PublicKey, Network, OutPoint};
 use core::time::Duration;
 use futures::lock::Mutex;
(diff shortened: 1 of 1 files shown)


mutiny-core/src/node.rs
```diff
@@ -242,6 +245,36 @@ impl Node {
         let channel_manager: Arc<PhantomChannelManager> =
             Arc::new(read_channel_manager.channel_manager);
 
+        // Check all existing channels against default configs.
+        // If we have default config changes, those should apply
+        // to all existing and new channels.
+        let default_config = default_user_config().channel_config;
+        for channel in channel_manager.list_channels() {
+            // unwrap is safe after LDK.0.0.109
+            if channel.config.unwrap() != default_config {
```
benthecarman, line 254: nit: Could do `if channel.config != Some(default_config)` just to be safe but not a big deal

benthecarman (approved):
Maybe eventually we should tweak this based on our fee estimates

PR #515, opened 2023-05-17
Nostr Wallet Connect support

benthecarman:
https://github.com/nostr-protocol/nips/blob/master/47.md

There is a PR to the rust-nostr library for adding better support. Once that is released we can clean up this code a lot.

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -25,6 +25,7 @@ pub mod logging;
 mod lspclient;
 mod node;
 pub mod nodemanager;
+mod nostr;
 mod onchain;
 mod peermanager;
 mod proxy;
@@ -39,11 +40,21 @@ pub use crate::keymanager::generate_seed;
 
 use crate::error::MutinyError;
 use crate::nodemanager::NodeManager;
+use crate::nostr::NostrManager;
 use crate::storage::MutinyStorage;
+use ::nostr::Kind;
 pub use auth::AuthProfile;
 use bip39::Mnemonic;
+use bitcoin::secp256k1::PublicKey;
+use bitcoin::util::bip32::ExtendedPrivKey;
 use bitcoin::Network;
+use futures::{pin_mut, select, FutureExt};
+use lightning::util::logger::Logger;
+use lightning::{log_error, log_warn};
+use nostr_sdk::{Client, RelayPoolNotification};
+use std::sync::atomic::Ordering;
 use std::sync::Arc;
+use wasm_bindgen_futures::spawn_local;
 
 #[derive(Clone)]
 pub struct MutinyWalletConfig {
@@ -84,6 +95,7 @@ pub struct MutinyWallet<S: MutinyStorage> {
     config: MutinyWalletConfig,
     storage: S,
     pub node_manager: Arc<NodeManager<S>>,
+    pub nostr: Arc<NostrManager>,
 }
 
 impl<S: MutinyStorage> MutinyWallet<S> {
@@ -108,10 +120,17 @@ impl<S: MutinyStorage> MutinyWallet<S> {
 
         let node_manager = Arc::new(NodeManager::new(config.clone(), storage.clone()).await?);
 
+        // create nostr manager
+        let seed = node_manager.show_seed().to_seed("");
+        let xprivkey = ExtendedPrivKey::new_master(node_manager.get_network(), &seed)?;
+        let relays = vec!["wss://relay.damus.io".to_string()]; // todo make configurable
+        let nostr = Arc::new(NostrManager::from_mnemonic(xprivkey, relays)?);
+
         Ok(Self {
             config,
             storage,
             node_manager,
+            nostr,
         })
     }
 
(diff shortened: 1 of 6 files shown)


mutiny-core/src/lib.rs
```diff
@@ -124,10 +141,53 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+                                    log_warn!(nm.logger, "Error sending NWC event: {e}");
+                                }
+                            }
+                            Ok(None) => {} // no response
+                            Err(e) => {
+                                log_error!(nm.logger, "Error handling NWC request: {e}");
+                            }
+                        }
+                    }
+                }
```
TonyGiorgio, line 182: Should this be wrapped in a loop? Otherwise if `notifications.recv()` throw an error (assuming connection error), then it'll stop listening to connections?

mutiny-core/src/lib.rs
```diff
@@ -124,10 +141,53 @@ impl<S: MutinyStorage> MutinyWallet<S> {
         Ok(())
     }
 
+    /// Starts a background process that will watch for nostr wallet connect events
+    pub async fn start_nostr_wallet_connect(&self, from_node: PublicKey) {
+        let nostr = self.nostr.clone();
+        let nm = self.node_manager.clone();
+        spawn_local(async move {
```
TonyGiorgio, line 148: It shouldn't actually be hard to stop here, just throw the atomic bool checker in the loop.

mutiny-core/src/lib.rs
```diff
@@ -108,10 +118,17 @@ impl<S: MutinyStorage> MutinyWallet<S> {
 
         let node_manager = Arc::new(NodeManager::new(config.clone(), storage.clone()).await?);
 
+        // create nostr manager
+        let seed = node_manager.show_seed().to_seed("");
+        let xprivkey = ExtendedPrivKey::new_master(node_manager.get_network(), &seed)?;
+        let relays = vec!["wss://relay.damus.io".to_string()]; // todo make configurable
```
TonyGiorgio, line 124: Should we at least do a second just in case damus goes down?

mutiny-core/src/nostr/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+        xprivkey: ExtendedPrivKey,
+        relays: Vec<String>,
+    ) -> Result<Self, MutinyError> {
+        if relays.is_empty() {
+            return Err(MutinyError::Other(anyhow!("No relays provided")));
+        }
+
+        let primary_key = Self::derive_nostr_key(xprivkey, 0)?;
+        let nwc_client_key = Self::derive_nostr_key(xprivkey, 1)?;
+        let nwc_server_key = Self::derive_nostr_key(xprivkey, 2)?;
```
TonyGiorgio, line 61: I wonder if this should use the auth stuff at all? Otherwise these wont ever change or support multiple. What if they leak?

TonyGiorgio (changes requested):

mutiny-core/src/lib.rs
benthecarman: For nostr wallet connect you can only have one so just hard coded this one for now

mutiny-core/src/nostr/mod.rs
benthecarman: Yeah I tried doing this originally but the lnurl derivation stuff is too different from the nostr derivation stuff. Maybe can unify one day but this was easier for now.

benthecarman:
Put in a loop and added support for `stop`

mutiny-core/src/lib.rs
```diff
@@ -124,10 +142,70 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+                    }
+                }
+
+                // handle NWC requests
+                let mut notifications = client.notifications();
+                while let Ok(notification) = notifications.recv().await {
+                    if nm.stop.load(Ordering::Relaxed) {
+                        utils::sleep(1_000).await;
+                        continue;
+                    };
```
TonyGiorgio, line 183: Actually there's a better way to handle this if you want to reference some socket code: 

```
            loop {
                let mut read_fut = conn_copy.read().fuse();
                let delay_fut = Box::pin(utils::sleep(1_000)).fuse();
                pin_mut!(delay_fut);
                select! {
                    msg_option = read_fut => {
                        // ...
                    }
                    _ = delay_fut => {
                        if stop_copy.load(Ordering::Relaxed) {
                            break;
                        }
                    }
                }

```

TonyGiorgio (changes requested):

mutiny-core/src/lib.rs
TonyGiorgio: Oh that's dumb that you can only have one

mutiny-core/src/lib.rs
benthecarman: Switched it to use that, not entirely sure if I did it correctly.

TonyGiorgio (approved):

TonyGiorgio (approved):

PR #520, opened 2023-05-18
Await funding transaction on open channel

benthecarman:
Similar to #375 but for channels.

Was able to take this logic from redshifts and make the `open_channel` functions better. Now they will await the funding transaction being created, this should give better feedback to the frontend if the channel open actually worked.


benthecarman:
Ready for re-review

TonyGiorgio (approved):

PR #521, opened 2023-05-19
Handle restarts for redshifts properly

benthecarman:
Before if someone closed mutiny while redshifts were attempting payments they would never be restarted. This fixes by adding a new status and making the `start_redshifts` function a little smarter.

First, added a new statuses called `ChannelOpened` and `ClosingChannels` this is set by the event handler when we get a `ChannelReady` event. This signals to the `start_redshifts` that it can start attempting payments. This also allows us to change it so we can enable handle restarts.

With adding the new statuses we can now handle restarts by only starting to attempt payments on redshifts that have the `ChannelOpened` when polling. We can then only check for redshifts with the status of `AttemptingPayments` at the very beginning to restart them. This makes it so every 10 seconds we are only checking for new redshifts.

### DIFF
--- a/mutiny-core/src/event.rs
+++ b/mutiny-core/src/event.rs
@@ -3,6 +3,7 @@ use crate::keymanager::PhantomKeysManager;
 use crate::ldkstorage::{MutinyNodePersister, PhantomChannelManager};
 use crate::logging::MutinyLogger;
 use crate::onchain::OnChainWallet;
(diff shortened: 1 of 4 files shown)


TonyGiorgio (approved):
Hard to fully remember this flow but looks like it should work, will be happy to start testing all of this again, nice work.

PR #523, opened 2023-05-22
Remove RGS workarounds

benthecarman:
We don't need to call `remove_stale_channels_and_tracking_with_time` anymore because syncing RGS will now do that as of ldk 0.0.115.

We also do not need to check if the last sync was 24 hours ago because the RGS server will send us an empty diff if there is nothing to sync.

You can test by syncing RGS and then restarting and seeing you synced again


TonyGiorgio (approved):
Works for me

PR #524, opened 2023-05-22
Default to opening channel to LSP

TonyGiorgio:


mutiny-core/src/nodemanager.rs
```diff
@@ -1335,12 +1335,24 @@ impl<S: MutinyStorage> NodeManager<S> {
...
 
+        let to_pubkey = match to_pubkey {
+            Some(pubkey) => pubkey,
+            None => {
+                if self.lsp_clients.is_empty() {
+                    return Err(MutinyError::PubkeyInvalid);
+                }
+                // pick a random lsp from the list to open a channel to
+                let rand = rand::random::<usize>() % self.lsp_clients.len();
+                self.lsp_clients[rand].pubkey
```
benthecarman, line 1352: the node has a defined LSP, we should use the same one here

benthecarman (changes requested):

benthecarman (approved):

Issue #366, opened 2023-04-15
rip out wasm from mutiny-core


TonyGiorgio:

benthecarman:
I think all that is remaining is the websocket stuff!

PR #534, opened 2023-05-24
Remove some unused deps

benthecarman:
Part of #366 

removes some unused stuff from when we were using local storage.

### DIFF
--- a/mutiny-core/Cargo.toml
+++ b/mutiny-core/Cargo.toml
@@ -20,7 +20,7 @@ cfg-if = "1.0.0"
 wasm-bindgen = "0.2.84"
 bip39 = { version = "2.0.0" }
 bitcoin-bech32 = "0.12"
-js-sys = "0.3.60"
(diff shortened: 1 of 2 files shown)


TonyGiorgio:
Conflicts

mutiny-core/Cargo.toml
```diff
@@ -20,15 +20,14 @@ cfg-if = "1.0.0"
 wasm-bindgen = "0.2.84"
 bip39 = { version = "2.0.0" }
 bitcoin-bech32 = "0.12"
-js-sys = "0.3.60"
+js-sys = { version = "0.3.60", optional = true }
```
TonyGiorgio, line 23: should this be removed from up here if it's specified for `wasm32` below?

benthecarman: it is set as optional, not really sure how all these target dependent flags work tbh

TonyGiorgio (approved):

PR #540, opened 2023-05-26
Request NWC events every 10 seconds

benthecarman:
Made it so we request nostr wallet connect events every 10 seconds, this makes it so we don't have to make it so blastr sends events to concurrent subscriptions. This also will help with failed invoices, we will reattempt them on the next loop.

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -51,7 +51,7 @@ use bitcoin::Network;
 use futures::{pin_mut, select, FutureExt};
 use lightning::util::logger::Logger;
 use lightning::{log_error, log_warn};
-use nostr_sdk::{Client, RelayPoolNotification};
+use nostr_sdk::{Client, RelayMessage, RelayPoolNotification};
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
 
@@ -150,9 +150,19 @@ impl<S: MutinyStorage> MutinyWallet<S> {
             let mut broadcasted_info = false;
             loop {
                 if nm.stop.load(Ordering::Relaxed) {
+                    break;
+                };
+
+                // check we have lightning channels ready
+                if nm
+                    .get_node(&from_node)
+                    .await
+                    .map(|n| n.channel_manager.list_usable_channels().is_empty())
+                    .unwrap_or(true)
+                {
                     utils::sleep(1_000).await;
                     continue;
-                };
+                }
 
                 let client = Client::new(&nostr.primary_key);
                 client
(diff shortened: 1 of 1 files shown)


mutiny-core/src/lib.rs
```diff
@@ -184,20 +184,32 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
-                                        }
                                     }
                                 }
+                                Ok(RelayPoolNotification::Message(_url, RelayMessage::EndOfStoredEvents(_))) => {
+                                    // after we process all events, sleep for 5 seconds
+                                    // and then we will reconnect to the relay
+                                    if let Err(e) = client.disconnect().await {
+                                        log_warn!(nm.logger, "Error disconnecting from nostr relay: {e}");
+                                    }
+                                    utils::sleep(10_000).await;
```
TonyGiorgio, line 209: Ideally this is stop compatible like this: 

```
                // wait up to 5s, checking graceful shutdown check each 1s.
                for _ in 0..5 {
                    if stop.load(Ordering::Relaxed) {
                        logging_db.stop();
                        return;
                    }
                    sleep(1_000).await;
                }
```

mutiny-core/src/lib.rs
```diff
@@ -184,20 +184,32 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
                                             }
                                         }
-                                        Ok(None) => {} // no response
-                                        Err(e) => {
-                                            log_error!(nm.logger, "Error handling NWC request: {e}");
-                                        }
                                     }
                                 }
+                                Ok(RelayPoolNotification::Message(_url, RelayMessage::EndOfStoredEvents(_))) => {
+                                    // after we process all events, sleep for 5 seconds
```
TonyGiorgio, line 204:
```suggestion
                                    // after we process all events, sleep for 10 seconds
```

TonyGiorgio (changes requested):
Blastr will delete events after they have been sent to the client right? I think this will work to continuously request events. But if starting from an off state, then we should either: 
- add a 5 or 10 second sleep before requesting events
- introduce some wait till connected to LSP
- add some retry handling to events pulled down from the relay

benthecarman:
> Blastr will delete events after they have been sent to the client right? 

It deletes after the client sends a response event

TonyGiorgio:
> > Blastr will delete events after they have been sent to the client right?
> 
> It deletes after the client sends a response event

Okay that makes sense actually. I think after the minor changes above it should be good to go then.

benthecarman:
Changed so it waits for it to have usable channels and now the 10 second sleep between runs works with stops

mutiny-core/src/lib.rs
```diff
@@ -154,6 +154,17 @@ impl<S: MutinyStorage> MutinyWallet<S> {
                     continue;
```
TonyGiorgio, line 154: This should be a break not a continue.

TonyGiorgio (changes requested):

TonyGiorgio (approved):

PR #544, opened 2023-05-30
Unify activity into ActivityItem struct

### LINKED ISSUE #541: Synthetic `Activity` list in mutiny-wasm
I could delete fairly nasty part of the frontend logic if I could get an `Activity` list straight from `mutiny_wallet`

Here's the relevant code rn:

https://github.com/MutinyWallet/mutiny-web/blob/master/src/components/Activity.tsx#L115

Now that we have an activity list we like I think it would be a good time to put this logic in `mutiny-wasm`

I believe the list I need would look like:

```rs
struct ActivityItem {
  kind: "onchain" | "lightning"
  id: string (either txid for onchain or a payment hash to lookup an invoice)
  contacts?: Vec<Contact>
  labels?: Vec<string>
  amount: bigint
  positive: bool
  last_updated: bigint 
}

fn list_activity() ->  Vec<ActivityItem> {}
```

list should be sorted by `last_updated`, newest first.

Then, for the details view I will use `get_transaction` / `get_invoice_by_hash` instead of using data straight from the list, which means the details view won't disappear when the list updates anymore. Also should be way cheaper for the frontend to ask for new versions of the activity list more often, and we can potentially do pagination as well down the line.

benthecarman:
Closes #541

### DIFF
--- a/mutiny-wasm/src/lib.rs
+++ b/mutiny-wasm/src/lib.rs
@@ -663,6 +663,29 @@ impl MutinyWallet {
         )?)
     }
 
+    /// Returns all the on-chain and lightning activity from the wallet.
+    #[wasm_bindgen]
+    pub async fn get_activity(&self) -> Result<JsValue /* Vec<ActivityItem> */, MutinyJsError> {
+        // get activity from the node manager
+        let activity = self.inner.node_manager.get_activity().await?;
+        let mut activity: Vec<ActivityItem> = activity.into_iter().map(|a| a.into()).collect();
+
+        // add contacts to the activity
+        let contacts = self.inner.node_manager.get_contacts()?;
+        for a in activity.iter_mut() {
+            // find labels that have a contact and add them to the item
+            for label in a.labels.iter() {
+                if let Some(contact) = contacts.get(label) {
+                    a.contacts.push(Contact::from(contact.clone()));
+                }
+            }
+            // remove labels that have a contact to prevent duplicates
+            a.labels.retain(|l| !contacts.contains_key(l));
+        }
+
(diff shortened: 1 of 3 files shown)


mutiny-wasm/src/models.rs
```diff
@@ -14,6 +14,92 @@ use wasm_bindgen::prelude::*;
...
+#[wasm_bindgen]
+pub enum ActivityType {
+    OnChain,
+    Lightning,
+    ChannelOpen,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[wasm_bindgen]
+pub struct ActivityItem {
```
benthecarman, line 27: @futurepaul does this struct look good to you?

mutiny-wasm/src/lib.rs
```diff
@@ -663,6 +663,29 @@ impl MutinyWallet {
...
+        for a in activity.iter_mut() {
+            // find labels that have a contact and add them to the item
+            for label in a.labels.iter() {
+                if let Some(contact) = contacts.get(label) {
+                    a.contacts.push(Contact::from(contact.clone()));
+                }
+            }
+            // remove labels that have a contact to prevent duplicates
+            a.labels.retain(|l| !contacts.contains_key(l));
+        }
```
TonyGiorgio, line 684: Why is this only in `mutiny-wasm`?

benthecarman: Didn't really know how to fit it into the enum. Felt more ergonomic to put here

benthecarman: added

TonyGiorgio (approved):

PR #548, opened 2023-05-31
Write channel manager and monitor to LocalStorage

benthecarman:
This is so we can have better assurances these are saved before LDK has async storage. This should hopefully help with force close issues.

Part of #536

### DIFF
--- a/mutiny-wasm/src/indexed_db.rs
+++ b/mutiny-wasm/src/indexed_db.rs
@@ -1,7 +1,8 @@
 use anyhow::anyhow;
+use gloo_storage::{LocalStorage, Storage};
 use gloo_utils::format::JsValueSerdeExt;
-use lightning::log_error;
 use lightning::util::logger::Logger;
+use lightning::{log_debug, log_error};
 use log::error;
 use mutiny_core::error::{MutinyError, MutinyStorageError};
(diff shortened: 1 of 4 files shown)


TonyGiorgio:
Let me give this a shot with channels first.

mutiny-wasm/src/indexed_db.rs
```diff
@@ -212,6 +232,18 @@ fn used_once(key: &str) -> bool {
     )
 }
 
+/// To help prevent force closes we save to local storage as well as indexed db.
+/// This is because indexed db is not always reliable.
+///
+/// We need to do this for the channel manager and channel monitors.
+fn write_to_local_storage(key: &str) -> bool {
+    match key {
+        CHANNEL_MANAGER_KEY => true,
```
TonyGiorgio, line 241: this should also be starts_with, because it looks like: `manager_c324484a-6de1-48fb-91c8-cf506db6ac79`

TonyGiorgio (changes requested):

mutiny-wasm/src/indexed_db.rs
benthecarman: good catch

PR #549, opened 2023-06-01
Networking wasm split

TonyGiorgio:
- [x] Move connecting logic
- [x] Hide all proxy logic behind wasm flag
- [x] Get alternative builds compiling properly
- [x] Implement lightning-net-tokio for non wasm code


TonyGiorgio:
This is ready for review though Im going to keep working on the github actions for it.

benthecarman:
should delete this file

https://github.com/MutinyWallet/mutiny-node/blob/master/.cargo/config

mutiny-core/Cargo.toml
```diff
@@ -5,35 +5,28 @@ name = "mutiny-core"
...
 
 cfg-if = "1.0.0"
-wasm-bindgen = "0.2.84"
 bip39 = { version = "2.0.0" }
-js-sys = { version = "0.3.60", optional = true }
 bitcoin = { version = "0.29.2", default-features = false, features = ["serde", "secp-recovery"] }
 bdk = { git = "https://github.com/MutinyWallet/bdk.git", branch = "mutiny-no-std", default-features = false }
 bdk_esplora = { git = "https://github.com/MutinyWallet/bdk.git", branch = "mutiny-no-std", default-features = false, features = ["async-https"] }
 bdk-macros = "0.6.0"
-getrandom = { version = "0.2", features = ["js"] }
```
benthecarman, line 27: can just remove this here and just in the wasm config add the feature

TonyGiorgio:
> should delete this file
> 
> https://github.com/MutinyWallet/mutiny-node/blob/master/.cargo/config

It causes problems with my IDE otherwise. I think we still need some default target runtime target and since mutiny-wasm is also here, I think we should keep it set to wasm.

mutiny-core/src/node.rs
```diff
@@ -449,24 +427,24 @@ impl<S: MutinyStorage> Node<S> {
...
+            uuid.clone(),
+            &lsp_client,
             stop.clone(),
-            logger.clone(),
-        )));
+            stopped_components.clone(),
+        )
+        .await?;
+
+        #[cfg(not(target_arch = "wasm32"))]
```
benthecarman, line 444: here too, the only diff is the `websocket_proxy_addr` I think you can just put that behind a build config

mutiny-core/src/node.rs
```diff
@@ -1484,16 +1382,26 @@ async fn start_reconnection_handling(
...
                     connect_logger.clone(),
                     connect_peer_man.clone(),
                     stop.clone(),
                 )
-                .await
-                {
+                .await;
+
+                #[cfg(not(target_arch = "wasm32"))]
+                let connect_res = connect_peer_if_necessary(
```
benthecarman, line 1396: same here

benthecarman (changes requested):
LFG, some small code smells I saw, will tests and give more review tomorrow

mutiny-core/src/nodemanager.rs
```diff
@@ -509,7 +512,26 @@ impl<S: MutinyStorage> NodeManager<S> {
                 fee_estimator.clone(),
                 wallet.clone(),
                 network,
+                esplora.clone(),
+                &lsp_clients,
+                logger.clone(),
                 websocket_proxy_addr.clone(),
```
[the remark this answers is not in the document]
TonyGiorgio: I feel like I sometimes had problems doing that, I'll try again.

mutiny-core/src/utils.rs
```diff
@@ -21,14 +20,20 @@ pub async fn sleep(millis: i32) {
...
-    SystemTime::now()
-        .duration_since(SystemTime::UNIX_EPOCH)
-        .unwrap()
+    #[cfg(target_arch = "wasm32")]
+    return instant::SystemTime::now()
+        .duration_since(instant::SystemTime::UNIX_EPOCH)
+        .unwrap();
+
+    #[cfg(not(target_arch = "wasm32"))]
+    return std::time::SystemTime::now()
```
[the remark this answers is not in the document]
TonyGiorgio: My thought was to just cut out a dependency if it wasn't needed, but I suppose it's fine?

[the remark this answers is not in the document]
benthecarman: Yeah good point that's probably better, less risk

mutiny-core/src/nodemanager.rs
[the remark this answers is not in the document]
TonyGiorgio: Good catches, I also took the oppportunity to combine the `connect_peer` functions.

mutiny-core/src/networking/proxy.rs
```diff
@@ -123,18 +119,14 @@ pub fn tcp_proxy_to_url(proxy_url: &str, peer_addr: &str) -> Result<String, Muti
...
     #[cfg(feature = "ignored_tests")]
-    use crate::proxy::*;
+    use crate::networking::proxy::*;
 
     use crate::test_utils::*;
 
-    use crate::proxy::{mutiny_conn_proxy_to_url, tcp_proxy_to_url};
+    use crate::networking::proxy::tcp_proxy_to_url;
 
     use wasm_bindgen_test::{wasm_bindgen_test as test, wasm_bindgen_test_configure};
```
benthecarman, line 132: should put `PEER_PUBKEY` behind `#[cfg(feature = "ignored_tests")]` to get rid of the warning.

mutiny-core/src/utils.rs
```diff
@@ -101,7 +106,7 @@ where
 {
     #[cfg(not(target_arch = "wasm32"))]
     {
-        tokio::task::spawn(future);
+        tokio::task::spawn_local(future);
```
benthecarman, line 109: this didn't work for me when trying to create a mutiny-server

switching to this did 

`tokio::task::LocalSet::new().spawn_local(future);`

benthecarman (commented):
`just test` just runs the browser tests, should add to it that it runs the normal tests too.
However, `cargo test` only works for me if we delete the `.cargo` folder. We also could move a lot of our tests to not need to run in the browser, maybe we just do that in a follow up

benthecarman:
Otherwise looks good to me! Awesome work!

TonyGiorgio:
Good catches. 

`just test` and `just clippy` should work for you now, without any need to delete the cargo config file.

.github/workflows/test.yml
```diff
@@ -106,11 +121,113 @@ jobs:
...
-      - name: Check formatting
-        run: cargo fmt --check
-
-      - name: Check clippy mutiny-core
-        run: cargo clippy --all-features --tests --package mutiny-core -- -D warnings
-
       - name: Check clippy mutiny-wasm
         run: cargo clippy --all-features --tests --package mutiny-wasm -- -D warnings
+
+  core_tests_linux:
```
benthecarman, line 127: last thing, should add caching to these CI jobs

benthecarman (approved):

PR #551, opened 2023-06-01
sort activity newest first and don't include unpaid invoices

futurepaul:


mutiny-core/src/nodemanager.rs
```diff
@@ -926,13 +926,19 @@ impl<S: MutinyStorage> NodeManager<S> {
...
+            if ln.paid {
+                activity.push(ActivityItem::Lightning(Box::new(ln)));
+            }
         }
         for on in onchain {
             activity.push(ActivityItem::OnChain(on));
         }
 
         activity.sort();
+        // Newest first
```
TonyGiorgio, line 939:
```
activity.sort_by(|a, b| b.cmp(a));
```

mutiny-core/src/nodemanager.rs
```diff
@@ -926,13 +926,19 @@ impl<S: MutinyStorage> NodeManager<S> {
         let onchain = self.list_onchain()?;
         let mut activity = Vec::with_capacity(lightning.len() + onchain.len());
         for ln in lightning {
-            activity.push(ActivityItem::Lightning(Box::new(ln)));
+            // Only show paid invoices
+            if ln.paid {
+                activity.push(ActivityItem::Lightning(Box::new(ln)));
+            }
```
TonyGiorgio, line 932: I think it would be valid to show pending invoices that have not been paid yet.

TonyGiorgio (changes requested):

mutiny-core/src/nodemanager.rs
futurepaul: It would be nice but I'm not sure how to do that. Rn what happens is you create an invoice and an on-chain addy. If it gets paid on-chain you have the invoice in the activity list as well.

mutiny-core/src/nodemanager.rs
benthecarman: We could also change it so the `impl ord` sorts in descending order.

futurepaul: I'm fine with tony's solution

mutiny-core/src/nodemanager.rs
futurepaul: Can work toward this in a followup? Def would be good to solve if we can find a nice solution

TonyGiorgio: Yeah can be another feature request if we really want it.

benthecarman (approved):

PR #553, opened 2023-06-01
Avoid 0 sat invoices or fees that go negative

TonyGiorgio:


mutiny-core/src/node.rs
```diff
@@ -615,9 +615,22 @@ impl<S: MutinyStorage> Node<S> {
...
-                .checked_sub(lsp_fee_msat / 1000)
-                .ok_or(MutinyError::BadAmountError)?;
+
+            // Convert the fee from msat to sat for comparison and subtraction
+            let lsp_fee_sat = lsp_fee_msat / 1000;
+
+            // Ensure that the fee is less than the amount being requested.
+            // If it isn't, we don't subtract it.
+            // This prevents amount from being subtracted down to 0.
+            // This will mean that the LSP fee will be paid by the payer instead.
```
benthecarman, line 625: Is this what we want?

TonyGiorgio: 1 sat invoices go down to 0 otherwise, and those are unpayable for obvious reasons.

TonyGiorgio: I think the alternative is just not allowing someone to receive 1 sat invoices if the LSP is charging a base fee of 1 sat.

TonyGiorgio: No, the above logic checks that: 

```
// Needs any amount over 0 if channel exists
            // Needs amount over 10k if no channel
            let has_usable_channel = self
                .channel_manager
                .list_channels_with_counterparty(&lsp.pubkey)
                .iter()
                .any(|c| c.inbound_capacity_msat >= amount_sat * 1000);
            let min_amount_sat = if has_usable_channel { 1 } else { 10_000 };
            if amount_sat < min_amount_sat {
                return Err(MutinyError::BadAmountError);
            }
```

benthecarman: I think this approach is fine as long as it doesn't effect the channel open fee.

TonyGiorgio: But there's this hardcoded 10k... so hmm.

benthecarman (approved):
I think we need to change that 10k number for Mainnet but that is separate to this.

mutiny-core/src/node.rs
TonyGiorgio: https://github.com/MutinyWallet/mutiny-node/issues/554

PR #561, opened 2023-06-05
Fix activity item sort for unconfirmed txs

benthecarman:
Unconfirmed txs were showing up at the bottom of the list because rust by default makes None < Some, but here we want the opposite so they are at the top of our list.

Added a test to make sure we don't have this again


futurepaul (approved):
works on my machine

PR #566, opened 2023-06-05
Optimize `get_invoice_by_hash`

benthecarman:


TonyGiorgio (approved):

mutiny-core/src/node.rs
```diff
@@ -27,7 +27,7 @@ use crate::{lspclient::FeeRequest, storage::MutinyStorage};
 use anyhow::{anyhow, Context};
 use bdk_esplora::esplora_client::AsyncClient;
 use bip39::Mnemonic;
-use bitcoin::hashes::{hex::ToHex, sha256::Hash as Sha256};
+use bitcoin::hashes::{hex::ToHex, sha256, sha256::Hash as Sha256};
```
benthecarman, line 30: Oh I could make this cleaner

benthecarman: was importing `sha256` just to do `sha256::Hash`, replaced that with just using the already imported `Sha256`

PR #571, opened 2023-06-07
After successful broadcast, insert transcation into bdk 

### LINKED ISSUE #569: On chain balance persists after new swap until sync occurs

benthecarman:
Closes #569

When opening a channel or sweeping a force close the transaction wouldn't be reflected until we synced because we didn't have BDK finalize the transaction. 

Here we make it so when we broadcast the transaction we insert it into bdk so it is reflected immediately.

### DIFF
--- a/mutiny-core/src/chain.rs
+++ b/mutiny-core/src/chain.rs
@@ -1,31 +1,38 @@
 use std::sync::Arc;
 
 use crate::esplora::EsploraSyncClient;
-use bdk_macros::maybe_await;
 use bitcoin::{Script, Transaction, Txid};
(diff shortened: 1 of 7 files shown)


benthecarman:
do not merge, found a bug i think

TonyGiorgio:
oh fuck ya thank you, was just going to make an issue because i'm missing millions of sats from my own chain wallet every time i open it

benthecarman:
Seems to have found a bdk bug with this :/

benthecarman:
Added a fix to get around the bug

TonyGiorgio (approved):

PR #582, opened 2023-06-10
Persist logs every 3 seconds

benthecarman:
we keep missing some logs it seems like, hopefully this will help


TonyGiorgio:
NGL, I'm a bit worried about reading and writing a 1MB file over and over again so much.

benthecarman:
> NGL, I'm a bit worried about reading and writing a 1MB file over and over again so much.

true.. that is a whole host of new problems

maybe we could change it to write each X second batch to a new key, then concat them all,

PR #583, opened 2023-06-10
Better handle bdk syncing errors

benthecarman:
Before it was kind of a black box if bdk was failing while syncing. This should hopefully let us know better what is wrong. Apparently bdk-alpha1.0 will fix a lot of the potential issues as well.

### DIFF
--- a/mutiny-core/src/onchain.rs
+++ b/mutiny-core/src/onchain.rs
@@ -107,10 +107,23 @@ impl<S: MutinyStorage> OnChainWallet<S> {
                 5,
             )
             .await?;
-        // get new wallet lock for writing
-        let mut wallet = self.wallet.try_write()?;
-        wallet.apply_update(update)?;
-        wallet.commit()?;
+
+        // get new wallet lock for writing and apply the update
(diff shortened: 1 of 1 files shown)


mutiny-core/src/onchain.rs
```diff
@@ -107,10 +107,23 @@ impl<S: MutinyStorage> OnChainWallet<S> {
...
             .await?;
-        // get new wallet lock for writing
-        let mut wallet = self.wallet.try_write()?;
-        wallet.apply_update(update)?;
-        wallet.commit()?;
+
+        // get new wallet lock for writing and apply the update
+        match self.wallet.try_write() {
+            Ok(mut wallet) => match wallet.apply_update(update) {
+                Ok(_) => wallet.commit()?,
```
TonyGiorgio, line 114: should `commit` be matched too?

benthecarman: No, that's just writing to storage which will return a proper error

TonyGiorgio (approved):

PR #585, opened 2023-06-12
Wait for our fee estimates before connecting to peers

benthecarman:
Recommendation from @tnull

### DIFF
--- a/mutiny-core/src/node.rs
+++ b/mutiny-core/src/node.rs
@@ -439,6 +439,7 @@ impl<S: MutinyStorage> Node<S> {
             &lsp_client,
             stop.clone(),
             stopped_components.clone(),
+            network == Network::Regtest,
         )
         .await?;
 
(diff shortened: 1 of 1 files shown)


TonyGiorgio:
>  just getting a fee rate estimation might be sufficient to fix the fee issues

Yeah let's try this instead if it's not too much of an issue. Paying invoices might start to really slow down if we're waiting on too many things. Especially syncing, which is a lot slower on mainnet.

benthecarman:
this seems to break tests :/

benthecarman:
Okay made it so we don't do the connection retrying on regtest

TonyGiorgio:
> Okay made it so we don't do the connection retrying on regtest

Err, not sure about that. Can you just make it not do fee checking on regtest instead?

TonyGiorgio (approved):

PR #592, opened 2023-06-14
Remove 10s reconnect for nwc

benthecarman:
Need this first: https://github.com/MutinyWallet/blastr/pull/23


TonyGiorgio:
This works great! I'm noticing sometimes I refresh and get sent the same events I had once already failed on but it doesn't hurt it. Probably just because cloudflare isn't always instant with KV updates. But all new events seem to come in within the 10s so nice work.

TonyGiorgio (approved):

PR #601, opened 2023-06-16
Handle specific errors from the LSP upon invoice

TonyGiorgio:
This will give us more indication of what went wrong during invoice creation. The common reasons are because the user is not connected to the LSP or the LSP has ran out of confirmed funds. 

@futurepaul you should review the API logic here. 

The proposed changes I would like `mutiny-web` to do is to check the error on `create_bip21` for an Lsp related error and fall back to `get_new_address` if that's the case. This is so that you may show the user the connection error or lsp funding error, but still allow them to proceed with receiving on chain funds with that warning in mind.

Relies on https://github.com/voltagecloud/lsp-server/pull/138

Related to https://github.com/MutinyWallet/mutiny-web/issues/249

### DIFF
--- a/mutiny-core/src/lspclient.rs
+++ b/mutiny-core/src/lspclient.rs
@@ -63,21 +63,27 @@ pub struct FeeResponse {
     pub fee_amount_msat: u64,
 }
 
+#[derive(Deserialize, Debug)]
+struct ErrorResponse {
+    error: String,
+    message: String,
+}
+
 const GET_INFO_PATH: &str = "/api/v1/info";
 const PROPOSAL_PATH: &str = "/api/v1/proposal";
 const FEE_PATH: &str = "/api/v1/fee";
 
 impl LspClient {
-    pub async fn new(url: &str) -> anyhow::Result<Self> {
+    pub async fn new(url: &str) -> Result<Self, MutinyError> {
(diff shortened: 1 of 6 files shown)


futurepaul (approved):
this will be a huge improvement thank you, especially appreciate the docs on what potential errors I can get so I can match on them

mutiny-core/src/lspclient.rs
```diff
@@ -114,38 +120,64 @@ impl LspClient {
...
             .await
-            .map_err(|_| MutinyError::LspFailure)?
-            .json()
-            .await
-            .map_err(|_| MutinyError::LspFailure)?;
-
-        Ok(proposal_response.jit_bolt11)
+            .map_err(|_| MutinyError::LspGenericError)?;
+
+        // if the status is 500, copy the response body to a string and try to parse as ErrorResponse
```
benthecarman, line 138: I think it'd be better checking for 200 and parsing the error on any non-successful request

benthecarman (commented):
One comment otherwise looks great

mutiny-core/src/lspclient.rs
TonyGiorgio: Changed

benthecarman (approved):

PR #602, opened 2023-06-17
Insert broadcasted channels on pending channel event

benthecarman:
We originally added calling `insert_tx` when we would broadcast to fix it when we'd open a channel and our on-chain balance wouldn't update immedaitely. However, this is causing issues with other txs being inserted into bdk on broadcast. So instead we change it to not broadcast on broadcast while this issue exists in bdk and use the channel open params to save the tx and insert it on the channel pending event. With this we can also wait until the channel pending event happens by waiting for the channel opening params are deleted.

### DIFF
--- a/mutiny-core/src/event.rs
+++ b/mutiny-core/src/event.rs
@@ -112,34 +112,39 @@ impl<S: MutinyStorage> EventHandler<S> {
                 };
 
                 let psbt_result = match &params_opt {
-                    None => self.wallet.create_signed_psbt_to_spk(
-                        output_script,
-                        channel_value_satoshis,
-                        None,
-                    ),
-                    Some(params) => {
-                        log_debug!(self.logger, "Opening channel with params: {params:?}");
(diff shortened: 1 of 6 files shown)


benthecarman:
need to test still, should test opening a channel and making sure that the on-chain balance is immediately reflected

benthecarman:
found a bug

TonyGiorgio:
This appears after a channel close, I should not have 15,000, I should only have 7500 on chain.

Screenshot from 2023-06-19 12-47-51

Then when I refresh, I don't see it at first: 
Screenshot from 2023-06-19 12-50-25


And refreshing over and over I just get different results. Either 0 sats, 7500 sats, or 15,000 sats and sometimes it changes after a few seconds of syncing.

TonyGiorgio (approved):
Works for me locally and no code smells. We'll live with some on chain inconsistencies until the next BDK alpha release.

PR #615, opened 2023-06-21
Add option to force close channel or abandon channel

benthecarman:
This should help with issues like @shyfire131 had where he has a channel that will never open.

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -1752,7 +1752,24 @@ impl<S: MutinyStorage> NodeManager<S> {
     }
 
     /// Closes a channel with the given outpoint.
-    pub async fn close_channel(&self, outpoint: &OutPoint, force: bool) -> Result<(), MutinyError> {
+    ///
+    /// If force is true, the channel will be force closed.
+    ///
+    /// If abandon is true, the channel will be abandoned.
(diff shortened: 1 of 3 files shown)


mutiny-core/src/nodemanager.rs
```diff
@@ -1765,17 +1765,34 @@ impl<S: MutinyStorage> NodeManager<S> {
...
-                            self.logger,
-                            "had an error closing channel {} with node {} : {e:?}",
-                            &channel.channel_id.to_hex(),
-                            &channel.counterparty.node_id.to_hex()
-                        );
-                        MutinyError::ChannelClosingFailed
-                    })?;
+                if force {
+                    node.channel_manager
+                        .force_close_broadcasting_latest_txn(
```
TonyGiorgio, line 1770: THe issue where the channel opening will never confirm, should that be by force closing without broadcasting latest txn?

benthecarman: hmm true, maybe add an `abandon_channel` function that does that

benthecarman: added

TonyGiorgio: What do you think of having `force` and `abandon` as a parameter here? The code is pretty similar.

TonyGiorgio (approved):

PR #616, opened 2023-06-21
Exporting log & state does not rely on nodemanager

TonyGiorgio:
Have not tested it yet, hoping @futurepaul does


futurepaul (approved):
seems to be working

PR #626, opened 2023-06-25
Bump to ldk 0.0.116-alpha1

TonyGiorgio:
TODO: Need to use main release once https://github.com/lightningdevkit/rust-lightning/pull/2376 merges or there's a way around it.

### DIFF
--- a/mutiny-core/Cargo.toml
+++ b/mutiny-core/Cargo.toml
@@ -12,7 +12,7 @@ homepage = "https://mutinywallet.com"
 repository = "https://github.com/mutinywallet/mutiny-node"
 
 [dependencies]
-lnurl-rs = { version = "=0.2.4", default-features = false, features = ["async", "async-https"] }
+lnurl-rs = { git = "https://github.com/TonyGiorgio/lnurl-rs.git", branch = "ldk-116-alpha-pub-sync", default-features = false, features = ["async", "async-https"] }
 
 cfg-if = "1.0.0"
 bip39 = { version = "2.0.0" }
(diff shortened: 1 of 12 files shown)


mutiny-core/Cargo.toml
```diff
@@ -12,7 +12,7 @@ homepage = "https://mutinywallet.com"
 repository = "https://github.com/mutinywallet/mutiny-node"
 
 [dependencies]
-lnurl-rs = { version = "0.2.4", default-features = false, features = ["async", "async-https"] }
+lnurl-rs = { git = "https://github.com/benthecarman/lnurl-rs.git", default-features = false, features = ["async", "async-https"] }
```
benthecarman, line 15: i made a release for this

mutiny-core/src/chain.rs
```diff
@@ -43,14 +43,16 @@ impl<S: MutinyStorage> Filter for MutinyChain<S> {
...
+        for tx in txs {
+            let tx_clone = tx.clone();
+            let wallet = self.wallet.clone();
+            let logger = self.logger.clone();
+            utils::spawn(async move {
+                if let Err(e) = wallet.broadcast_transaction(tx_clone).await {
+                    log_warn!(logger, "Error broadcasting transaction: {e}")
+                }
+            });
+        }
```
benthecarman, line 56: would be better to have the for inside the spawn

benthecarman:
you should be able to change this to use `update_channel_from_announcement_no_lookup` and delete the dummy utxo lookup impl i made

https://github.com/MutinyWallet/mutiny-node/blob/master/mutiny-core/src/peermanager.rs#L227

benthecarman:
can also remove this whole match statement and replace it with `bal.claimable_amount_satoshis`

https://github.com/MutinyWallet/mutiny-node/blob/f1c7ea0400c2a10d3a8a4fc96743a1d1536a498b/mutiny-core/src/nodemanager.rs#L1268

PR #629, opened 2023-06-26
Restore mnemonic

TonyGiorgio:


mutiny-core/src/lib.rs
```diff
@@ -258,6 +258,17 @@ impl<S: MutinyStorage> MutinyWallet<S> {
 
         Ok(())
     }
+
+    /// Restore's the mnemonic after deleting the previous state.
+    ///
+    /// Backup the state beforehand. Does not restore lightning data.
+    /// Should refresh or restart afterwards. Wallet should be stopped.
+    pub async fn restore_mnemonic(storage: S, m: String) -> Result<(), MutinyError> {
```
benthecarman, line 266: this can take in a `Mnemonic` instead of `String` and do the parsing on the wasm side

benthecarman (approved):

mutiny-core/src/lib.rs
TonyGiorgio: Good catch.

PR #634, opened 2023-06-29
Allow for approval and denial of nwc events

benthecarman:
Lets us approve or deny invoices that we get from NWC

### DIFF
--- a/mutiny-core/src/nostr/mod.rs
+++ b/mutiny-core/src/nostr/mod.rs
@@ -1,11 +1,15 @@
 use crate::error::MutinyError;
 use crate::nodemanager::NodeManager;
-use crate::nostr::nwc::{NostrWalletConnect, NwcProfile, Profile};
+use crate::nostr::nwc::{
+    NostrWalletConnect, NwcProfile, PendingNwcInvoice, Profile, PENDING_NWC_EVENTS_KEY,
+};
 use crate::storage::MutinyStorage;
+use bitcoin::hashes::sha256;
 use bitcoin::secp256k1::{PublicKey, Secp256k1, Signing};
 use bitcoin::util::bip32::{ChildNumber, DerivationPath, ExtendedPrivKey};
 use nostr::key::SecretKey;
-use nostr::{Event, Filter, Keys};
+use nostr::prelude::encrypt;
+use nostr::{Event, EventBuilder, EventId, Filter, Keys, Kind, Tag};
 use nostr_sdk::Client;
 use std::str::FromStr;
 use std::sync::{Arc, RwLock};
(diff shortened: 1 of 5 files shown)


benthecarman:
Need to add tests still

mutiny-core/src/nostr/mod.rs
```diff
@@ -173,6 +178,113 @@ impl<S: MutinyStorage> NostrManager<S> {
...
+            &nwc.client_pubkey(),
+            resp.as_json(),
+        )
+        .unwrap();
+
+        let p_tag = Tag::PubKey(inv.pubkey, None);
+        let e_tag = Tag::Event(inv.event_id, None, None);
+        let response = EventBuilder::new(Kind::WalletConnectResponse, encrypted, &[p_tag, e_tag])
+            .to_event(&nwc.server_key)
+            .unwrap();
```
TonyGiorgio, line 241: Unwrapping on nostr key stuff makes me nervous.

mutiny-core/src/nostr/mod.rs
```diff
@@ -173,6 +178,113 @@ impl<S: MutinyStorage> NostrManager<S> {
...
+
+        let event_id = client
+            .send_event(response)
+            .await
+            .map_err(|e| MutinyError::Other(anyhow::anyhow!("Failed to send info event: {e:?}")))?;
+
+        // remove from storage
+        pending.retain(|x| x.invoice.payment_hash() != &hash);
+        self.storage.set_data(PENDING_NWC_EVENTS_KEY, pending)?;
```
TonyGiorgio, line 250: There's a race here where more pending NWCs have been added since the last time. We need a lock here to grab all of the values and then take our paid invoice out and set it.  This will be the case where any of `set_data` is being called here.

mutiny-core/src/nostr/nwc.rs
```diff
@@ -117,62 +158,66 @@ impl NostrWalletConnect {
...
+                    .storage
+                    .set_data(PENDING_NWC_EVENTS_KEY, current)?;
+
+                return Ok(None);
             } else {
-                log_warn!(
-                    node_manager.logger,
-                    "Invoice amount too high: {msats} msats"
-                );
+                let msats = invoice.amount_milli_satoshis().unwrap_or(0);
```
TonyGiorgio, line 189: We can't auto pay 0 amount invoices.

TonyGiorgio (changes requested):

mutiny-core/src/nostr/mod.rs
benthecarman: The error would only happen if we didn't have the private key, changed to map the error tho

mutiny-core/src/nostr/mod.rs
TonyGiorgio: A RwLock or mutex will have to be introduced to not run into time races, and there's other areas where `set_data` is being called. 

This will especially be a problem when someone does an approval all or there's multiple that come in while one is being paid.

benthecarman: Added a mutex around it

TonyGiorgio (approved):

PR #635, opened 2023-06-30
Add mutiny auth service

TonyGiorgio:
Tested it already locally, but removed the check. Right now it doesn't have another purpose until we add something to make an authenticated request to.

### DIFF
--- a/mutiny-core/src/auth.rs
+++ b/mutiny-core/src/auth.rs
@@ -1,0 +1,216 @@
+#![allow(dead_code)]
+use crate::{
+    error::MutinyError,
+    lnurlauth::{make_lnurl_auth_connection, AuthManager},
+    logging::MutinyLogger,
+    networking::websocket::{SimpleWebSocket, WebSocketImpl},
+    storage::MutinyStorage,
+};
+use jwt_compact::UntrustedToken;
+use lightning::util::logger::*;
+use lightning::{log_error, log_info};
+use lnurl::{lnurl::LnUrl, AsyncClient as LnUrlClient};
+use reqwest::Client;
+use reqwest::{Method, StatusCode, Url};
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use std::str::FromStr;
+use std::sync::{Arc, RwLock};
+
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+struct CustomClaims {
+    sub: String,
+}
+
+pub(crate) struct MutinyAuthClient<S: MutinyStorage> {
+    auth: AuthManager<S>,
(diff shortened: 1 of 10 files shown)


mutiny-core/src/auth.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    async fn authenticated_request(
+        &self,
+        method: Method,
+        url: Url,
+        body: Option<Value>,
+    ) -> Result<reqwest::Response, MutinyError> {
+        let mut request = self.http_client.request(method, url);
+
+        if self.jwt.try_read().unwrap().is_none() {
```
benthecarman, line 89: should use `is_authenticated` here

benthecarman: maybe should have it return `Option<String>` of the token and can use that for the request

mutiny-core/src/auth.rs
```diff
@@ -0,0 +1,219 @@
...
+            self.retrieve_new_jwt().await?;
+        }
+        request = request.bearer_auth(
+            self.jwt
+                .try_read()?
+                .as_ref()
+                .expect("just retrieved or had already")
+                .as_str(),
+        );
```
benthecarman, line 98: this is kinda ugly, should be able to use the token we check above

mutiny-core/src/auth.rs
```diff
@@ -0,0 +1,219 @@
...
+        let mut ws = WebSocketImpl::new(url.to_string()).await.map_err(|e| {
+            log_error!(self.logger, "Error starting up auth ws: {e}");
+            MutinyError::LnUrlFailure
+        })?;
+
+        let lnurl_auth_str = ws.recv().await.map_err(|e| {
+            log_error!(self.logger, "Error receiving LNURL from ws: {e}");
+            MutinyError::LnUrlFailure
+        })?;
+        let lnurl = match LnUrl::decode(lnurl_auth_str.clone()) {
```
benthecarman, line 130: you should be able to just do `&lnurl_auth_str` instead of clone

mutiny-core/src/auth.rs
```diff
@@ -0,0 +1,215 @@
...
+        self.retrieve_new_jwt().await?;
+        Ok(())
+    }
+
+    pub fn is_authenticated(&self) -> (bool, Option<String>) {
+        if let Some(ref jwt) = *self.jwt.try_read().unwrap() {
+            return (true, Some(jwt.to_string())); // TODO parse and make sure still valid
+        }
+        (false, None)
+    }
```
benthecarman, line 61: don't think we need the bool here, is implicit from the Option being defined or not

benthecarman (approved):

PR #636, opened 2023-07-01
Gifting

benthecarman:
Took a stab at the gifting feature. Turns out is is fairly simple, tricky part is making sure the NWC is single use. Need to test and spend some time on that.

### DIFF
--- a/mutiny-core/src/nostr/nwc.rs
+++ b/mutiny-core/src/nostr/nwc.rs
@@ -21,20 +21,37 @@ use std::str::FromStr;
 pub(crate) const PENDING_NWC_EVENTS_KEY: &str = "pending_nwc_events";
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct SingleUseSpendingConditions {
+    pub spent: bool,
+    pub amount_sats: u64,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum SpendingConditions {
+    SingleUse(SingleUseSpendingConditions),
+    /// Require approval before sending a payment
+    RequireApproval,
+}
+
+impl Default for SpendingConditions {
+    fn default() -> Self {
+        Self::RequireApproval
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub(crate) struct Profile {
     pub name: String,
     pub index: u32,
-    /// Maximum amount of sats that can be sent in a single payment
-    pub max_single_amt_sats: u64,
     pub relay: String,
     pub enabled: bool,
     /// Require approval before sending a payment
     #[serde(default)]
-    pub require_approval: bool,
+    pub spending_conditions: SpendingConditions,
 }
 
 impl PartialOrd for Profile {
-    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
         self.index.partial_cmp(&other.index)
     }
 }
@@ -107,7 +124,6 @@ impl NostrWalletConnect {
         from_node: &PublicKey,
         invoice: &Bolt11Invoice,
     ) -> Result<Response, MutinyError> {
-        // todo we could get the author of the event we zapping and use that as the label
         let labels = vec![self.profile.name.clone()];
         match node_manager
             .pay_invoice(from_node, invoice, None, labels)
@@ -129,15 +145,18 @@ impl NostrWalletConnect {
         }
     }
 
-    /// Handle a Nostr Wallet Connect request, returns a response event if one is needed
+    /// Handle a Nostr Wallet Connect request
+    ///
+    /// Returns a response event if one is needed and if the profile needs to be saved to disk
     pub async fn handle_nwc_request<S: MutinyStorage>(
-        &self,
+        &mut self,
         event: Event,
         node_manager: &NodeManager<S>,
         from_node: &PublicKey,
         pending_nwc_lock: &Mutex<()>,
-    ) -> anyhow::Result<Option<Event>> {
+    ) -> anyhow::Result<(Option<Event>, bool)> {
         let client_pubkey = self.client_key.public_key();
+        let mut needs_save = false;
         if self.profile.enabled
             && event.kind == Kind::WalletConnectRequest
             && event.pubkey == client_pubkey
@@ -149,7 +168,7 @@ impl NostrWalletConnect {
 
             // only respond to pay invoice requests
             if req.method != Method::PayInvoice {
-                return Ok(None);
+                return Ok((None, needs_save));
             }
 
             let invoice = Bolt11Invoice::from_str(&req.params.invoice)
@@ -157,7 +176,7 @@ impl NostrWalletConnect {
 
             // if the invoice has expired, skip it
             if invoice.would_expire(utils::now()) {
-                return Ok(None);
+                return Ok((None, needs_save));
             }
 
             // if the invoice has no amount, we cannot pay it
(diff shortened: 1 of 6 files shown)


mutiny-core/src/nostr/nwc.rs
```diff
@@ -155,94 +175,111 @@ impl NostrWalletConnect {
...
+                        return Ok((None, needs_save));
+                    }
 
-                return Ok(Some(response));
+                    let msats = invoice.amount_milli_satoshis().unwrap();
+
+                    // verify amount is under our limit
+                    let content = if msats <= single_use.amount_sats * 1_000 {
+                        let response = self
+                            .pay_nwc_invoice(node_manager, from_node, &invoice)
```
benthecarman, line 211: we could have a race here where we get multiple requests at the same time and pay all of them before we update that this nwc is spent.

TonyGiorgio:
needs rebase

TonyGiorgio (commented):
Some initial comments, haven't gone through it all yet.

mutiny-core/src/nostr/nwc.rs
benthecarman: this is actually fine because we don't process in parallel

mutiny-core/src/nostr/nwc.rs
```diff
@@ -166,99 +185,123 @@ impl NostrWalletConnect {
...
+                        match self
+                            .pay_nwc_invoice(node_manager, from_node, &invoice)
+                            .await
+                        {
+                            Ok(resp) => {
+                                // mark as spent and disable profile
+                                single_use.spent = true;
+                                self.profile.spending_conditions =
+                                    SpendingConditions::SingleUse(single_use);
+                                self.profile.enabled = false;
```
TonyGiorgio, line 227: thoughts on deleting the profile instead? Don't need many profiles sitting around if this is a popular thing.

mutiny-core/src/nostr/nwc.rs
```diff
@@ -166,99 +185,123 @@ impl NostrWalletConnect {
...
-                        .await
-                    {
-                        Ok(resp) => resp,
-                        Err(e) => Response {
+            match self.profile.spending_conditions.clone() {
+                SpendingConditions::SingleUse(mut single_use) => {
+                    // check if we have already spent
+                    if single_use.spent {
+                        return Ok((None, needs_save));
+                    }
```
TonyGiorgio, line 212: how do we make sure this is atomic? Should we introduce a lock here? Wouldn't want multiple requests coming in trying to activate this in a race. We already a duplicate problem.

mutiny-core/src/nostr/nwc.rs
```diff
@@ -166,99 +185,123 @@ impl NostrWalletConnect {
...
+                            .pay_nwc_invoice(node_manager, from_node, &invoice)
+                            .await
+                        {
+                            Ok(resp) => {
+                                // mark as spent and disable profile
+                                single_use.spent = true;
+                                self.profile.spending_conditions =
+                                    SpendingConditions::SingleUse(single_use);
+                                self.profile.enabled = false;
+                                needs_save = true;
```
TonyGiorgio, line 228: is this the only time `needs_save` is set? Would be nice if this was a bit cleaner and we weren't passing around `needs_save` everywhere but maybe it's the best approach.

mutiny-core/src/nostr/nwc.rs
```diff
@@ -166,99 +185,123 @@ impl NostrWalletConnect {
...
+                        .unwrap_or_default();
+
+                    current.push(pending);
+
+                    node_manager
+                        .storage
+                        .set_data(PENDING_NWC_EVENTS_KEY, current, None)?;
+
+                    return Ok((None, needs_save));
+                }
```
TonyGiorgio, line 290: should this not be saved? or is the saving happening here instead of from the result of the call here?

mutiny-core/src/nostr/nwc.rs
benthecarman: If we delete then keys will be reused, maybe we should add another flag like `archived` so it doesn't show up in the list

mutiny-core/src/nostr/nwc.rs
benthecarman: We only handle 1 nwc request at a time so it should be safe, but yeah, maybe not the safest to rely on that

mutiny-core/src/nostr/nwc.rs
benthecarman: yeah, couldn't figure out a cleaner way

mutiny-core/src/nostr/nwc.rs
benthecarman: we dont need to update the nwc profile here, just save the invoice

mutiny-core/src/nostr/nwc.rs
TonyGiorgio: Damn, well maybe we need to solve the key reuse thing as part of this too. Maybe we blow away or rewrite all profiles and communicate to user's that they need to relink them?

mutiny-core/src/nostr/nwc.rs
TonyGiorgio: That's different than paying/approving them right? Because I think we can pay many at the same time. Probably different flow not related to this but yeah maybe just being sure would be good.

benthecarman: Yeah, that is separate

mutiny-core/src/nostr/nwc.rs
```diff
@@ -166,99 +185,123 @@ impl NostrWalletConnect {
...
 
             // if we have already paid this invoice, skip it
             let node = node_manager.get_node(from_node).await?;
             if node.get_invoice(&invoice).is_ok_and(|i| i.paid) {
-                return Ok(None);
+                return Ok((None, needs_save));
             }
             drop(node);
 
+            if invoice.amount_milli_satoshis().is_none() {
```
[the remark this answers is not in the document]
benthecarman: woops had it repeated

mutiny-wasm/src/models.rs
```diff
@@ -854,17 +856,44 @@ impl NwcProfile {
...
+    #[wasm_bindgen(getter)]
+    pub fn sharable_url(&self) -> Option<String> {
+        match self.spending_conditions {
+            SpendingConditions::SingleUse(ref single_use) => {
+                let encoded = urlencoding::encode(&self.nwc_uri);
+                // todo figure out how to change the url
+                Some(format!(
+                    "https://app.mutinywallet.com/gift?amount={}&nwc_uri={}",
+                    single_use.amount_sats, encoded
+                ))
```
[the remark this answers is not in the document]
benthecarman: added `url_prefix`, not sure how else we'd handle this

mutiny-core/src/nostr/nwc.rs
benthecarman: If we want to change how we derive keys for nwc, probably suitable for a follow up (would be a pretty large change)

TonyGiorgio: Good point. Then set an archive flag with a TODO comment? Not sure, users have asked to remove some NWCs anyways not just gifting ones. Up to you.

mutiny-wasm/src/models.rs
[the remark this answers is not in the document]
TonyGiorgio: yeah i think that makes sense, though the `todo` comment can be removed

mutiny-core/src/nostr/nwc.rs
benthecarman: Okay added the archiving

mutiny-wasm/src/models.rs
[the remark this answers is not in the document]
benthecarman: oh woops, fixed

mutiny-core/src/nostr/nwc.rs
```diff
@@ -268,24 +311,20 @@ impl NostrWalletConnect {
...
     pub index: u32,
-    /// Maximum amount of sats that can be sent in a single payment
-    pub max_single_amt_sats: u64,
     pub relay: String,
     pub enabled: bool,
-    /// Require approval before sending a payment
-    pub require_approval: bool,
     pub nwc_uri: String,
+    pub spending_conditions: SpendingConditions,
 }
```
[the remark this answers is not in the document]
TonyGiorgio: Wait did this have an answer at all?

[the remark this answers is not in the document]
benthecarman: I added `#[serde(default)]` so it'll use the default (RequireApproval) if it is not there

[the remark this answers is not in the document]
TonyGiorgio: Okay I think that'll work then

benthecarman:
Found one issue that I fixed, otherwise only issue seemed to be general nwc stuff

https://github.com/MutinyWallet/mutiny-node/pull/636

PR #640, opened 2023-07-03
Fix pending nwc model for paul

benthecarman:
We can't use the nice functions on the model because we have to serialize to json, just add all the data to a json object instead


futurepaul (approved):
this is working on my machine

TonyGiorgio (approved):

PR #641, opened 2023-07-03
Await when clearing the storage

TonyGiorgio:


benthecarman (approved):
whoever made this bug is fired

PR #643, opened 2023-07-06
Add subscription client

TonyGiorgio:
This adds 4 APIs relating to the subscription client: 

```rust
    /// Checks whether or not the user is subscribed to Mutiny+.
    ///
    /// Returns None if there's no subscription at all.
    /// Returns Some(u64) for their unix expiration timestamp, which may be in the
    /// past or in the future, depending on whether or not it is currently active.
    #[wasm_bindgen]
    pub async fn check_subscribed(&self) -> Result<Option<u64>, MutinyJsError> {
        Ok(self.inner.node_manager.check_subscribed().await?)
    }

    /// Gets the subscription plans for Mutiny+ subscriptions
    #[wasm_bindgen]
    pub async fn get_subscription_plans(&self) -> Result<JsValue /* Vec<Plan> */, MutinyJsError> {
        let plans = self.inner.node_manager.get_subscription_plans().await?;

        Ok(JsValue::from_serde(&plans)?)
    }

    /// Subscribes to a Mutiny+ plan with a specific plan id.
    ///
    /// Returns a lightning invoice so that the plan can be paid for to start it.
    #[wasm_bindgen]
    pub async fn subscribe_to_plan(&self, id: u8) -> Result<MutinyInvoice, MutinyJsError> {
        Ok(self.inner.node_manager.subscribe_to_plan(id).await?.into())
    }

    /// Pay the subscription invoice. This will post a NWC automatically afterwards.
    pub async fn pay_subscription_invoice(&self, inv: &Invoice) -> Result<(), MutinyJsError> {
        Ok(self.inner.pay_subscription_invoice(inv).await?.into())
    }
```

Depends on https://github.com/MutinyWallet/subscriptions-api/pull/6

Only tested the plans API so far but the rest should work. 

Requires a new `subscription_url` to be passed in from the frontend for `node_manager`.

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -2,7 +2,6 @@ use anyhow::anyhow;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::{collections::HashMap, ops::Deref, sync::Arc};
 
-use crate::labels::LabelStorage;
 use crate::logging::LOGGING_KEY;
 use crate::redshift::{RedshiftManager, RedshiftStatus, RedshiftStorage};
 use crate::scb::{
@@ -29,6 +28,7 @@ use crate::{
     event::{HTLCStatus, PaymentInfo},
     lnurlauth::make_lnurl_auth_connection,
 };
+use crate::{labels::LabelStorage, subscription::MutinySubscriptionClient};
 use crate::{
     lnurlauth::{AuthManager, AuthProfile},
     MutinyWalletConfig,
@@ -483,6 +483,17 @@ pub struct LnUrlParams {
     pub tag: String,
 }
 
+/// Plan is a subscription plan for Mutiny+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct Plan {
+    /// The ID of the internal plan.
+    /// Used for subscribing to specific one.
+    pub id: u8,
+
+    /// The amount in sats for the plan.
+    pub amount_sat: u64,
+}
+
 /// The [NodeManager] is the main entry point for interacting with the Mutiny Wallet.
 /// It is responsible for managing the on-chain wallet and the lightning nodes.
 ///
@@ -508,8 +519,7 @@ pub struct NodeManager<S: MutinyStorage> {
     auth: AuthManager<S>,
     lnurl_client: Arc<LnUrlClient>,
     pub(crate) lsp_clients: Vec<LspClient>,
-    #[allow(dead_code)]
-    pub(crate) auth_client: Option<Arc<MutinyAuthClient<S>>>,
+    pub(crate) subscription_client: Option<Arc<MutinySubscriptionClient<S>>>,
     pub(crate) logger: Arc<MutinyLogger>,
     bitcoin_price_cache: Arc<Mutex<Option<(f32, Duration)>>>,
     do_not_connect_peers: bool,
@@ -685,6 +695,21 @@ impl<S: MutinyStorage> NodeManager<S> {
             None
         };
 
+        let subscription_client = if let Some(auth_client) = auth_client {
+            if let Some(subscription_url) = c.subscription_url {
+                let s = Arc::new(MutinySubscriptionClient::new(
+                    auth_client,
+                    subscription_url,
+                    logger.clone(),
+                ));
+                Some(s)
+            } else {
+                None
+            }
+        } else {
+            None
+        };
+
         let nm = NodeManager {
             stop,
             mnemonic,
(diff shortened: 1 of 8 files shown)


mutiny-wasm/src/lib.rs
```diff
@@ -1049,6 +1051,37 @@ impl MutinyWallet {
...
+    /// Subscribes to a Mutiny+ plan with a specific plan id.
+    ///
+    /// Returns a lightning invoice so that the plan can be paid for to start it.
+    #[wasm_bindgen]
+    pub async fn subscribe_to_plan(&self, id: u8) -> Result<MutinyInvoice, MutinyJsError> {
+        Ok(self.inner.node_manager.subscribe_to_plan(id).await?.into())
+    }
+
+    /// Pay the subscription invoice. This will post a NWC automatically afterwards.
+    pub async fn pay_subscription_invoice(&self, inv: &Invoice) -> Result<(), MutinyJsError> {
```
futurepaul, line 1081: I think this should take a bolt11 string arg, I don't believe I can construct an `Invoice` type from the js side

mutiny-wasm/src/models.rs
```diff
@@ -889,3 +889,39 @@ impl From<nostr::nwc::PendingNwcInvoice> for PendingNwcInvoice {
         }
     }
 }
+
+// This is the NodeIdentity that refer to a specific node
```
futurepaul, line 893: looks like a copy paste miss

futurepaul (changes requested):

mutiny-core/src/nodemanager.rs
```diff
@@ -685,6 +690,32 @@ impl<S: MutinyStorage> NodeManager<S> {
             None
         };
 
+        let subscription_client = if let Some(auth_client) = auth_client.clone() {
+            if let Some(subscription_url) = c.subscription_url {
+                let s = Arc::new(MutinySubscriptionClient::new(
+                    auth_client,
+                    subscription_url,
+                    logger.clone(),
+                )); // TODO url
```
benthecarman, line 699: think you can remove this todo

mutiny-core/src/nodemanager.rs
```diff
@@ -685,6 +690,32 @@ impl<S: MutinyStorage> NodeManager<S> {
...
+
+                // TODO remove
+                match s.get_plans().await {
+                    Ok(p) => {
+                        log_debug!(logger, "got plans: {p:?}");
+                    }
+                    Err(e) => {
+                        log_error!(logger, "failed to get plans: {e}")
+                    }
+                }
```
benthecarman, line 709: can remove this now?

mutiny-wasm/src/models.rs
```diff
@@ -889,3 +889,39 @@ impl From<nostr::nwc::PendingNwcInvoice> for PendingNwcInvoice {
...
+
+    #[wasm_bindgen(getter)]
+    pub fn id(&self) -> u8 {
+        self.id.clone()
+    }
+
+    #[wasm_bindgen(getter)]
+    pub fn amount_sat(&self) -> u64 {
+        self.amount_sat
+    }
```
benthecarman, line 917: you don't need these and can just make the values `pub`, we only need to do this for Strings

TonyGiorgio:
Resolved, I'll fix the tests in a bit.

mutiny-core/src/nostr/mod.rs
```diff
@@ -117,11 +135,27 @@ impl<S: MutinyStorage> NostrManager<S> {
...
+            .iter()
+            .filter(|&nwc| nwc.profile.index >= USER_NWC_PROFILE_START_INDEX)
+            .count() as u32;
+
+        let (name, index) = match profile_type {
+            ProfileType::Reserved(reserved_profile) => {
+                let (name, index) = reserved_profile.info();
+                (name.to_string(), index)
+            }
+            // Ensure normal profiles start from 100
```
benthecarman, line 153: think this is supposed to be 1000

mutiny-core/src/nostr/mod.rs
```diff
@@ -117,11 +135,27 @@ impl<S: MutinyStorage> NostrManager<S> {
...
+            ProfileType::Normal { name } => {
+                (name, normal_profiles_count + USER_NWC_PROFILE_START_INDEX)
```
benthecarman, line 155: should put a check here that the index isn't over 1000, not sure what we should do in the case that happens

benthecarman (commented):
I really like the changes for doing >1k indexes for reserved, just a couple things

mutiny-core/src/nostr/mod.rs
benthecarman: maybe we could do negative numbers instead

TonyGiorgio: I now set a panic where we convert `ReservedProfile` to an index. So it'll never have an index that is greater than 1000.

TonyGiorgio: And normal profile counts *can* be over 1000. Not sure if that's what you were referring to now that I read your comment again.

benthecarman (approved):

futurepaul (approved):

PR #645, opened 2023-07-06
VSS client

benthecarman:


mutiny-core/src/auth.rs
```diff
@@ -76,7 +77,11 @@ impl<S: MutinyStorage> MutinyAuthClient<S> {
                 self.retrieve_new_jwt().await?;
                 self.authenticated_request(method, url, body).await
             }
-            _ => Ok(res),
+            StatusCode::OK | StatusCode::ACCEPTED | StatusCode::CREATED => Ok(res),
+            code => Err(MutinyError::Other(anyhow!(
```
TonyGiorgio, line 81: I don't think we need to anyhow this. We have connection errors.

mutiny-core/src/vss.rs
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+
+    pub async fn put_objects(&self, items: Vec<VssKeyValueItem>) -> Result<(), MutinyError> {
+        let url = Url::parse(&format!("{}/putObjects", self.url)).map_err(|e| {
+            log_error!(self.logger, "Error parsing put objects url: {e}");
+            MutinyError::Other(anyhow!("Error parsing put objects url: {e}"))
+        })?;
+
+        // todo do we need global version here?
```
TonyGiorgio, line 44: Yeah we may need to add something to the auth_client to get the linking_key?

benthecarman: this is a global version number, nothing to do with the linking key. I am not really sure the point of the global version number

mutiny-core/src/auth.rs
benthecarman: This gives more info, was useful for debugging

TonyGiorgio: You could log the error here and then return the connection error.

TonyGiorgio (approved):
Clean and straitforward. Just a nit about not using `anyhow` otherwise LGTM.

mutiny-core/src/nodemanager.rs
```diff
@@ -671,47 +653,33 @@ impl<S: MutinyStorage> NodeManager<S> {
...
+                (Some(s), auth)
             } else {
-                None
+                (None, auth_client.auth.clone())
             }
         } else {
-            None
+            let auth_manager = AuthManager::new(c.xprivkey)?;
+            (None, auth_manager)
         };
```
TonyGiorgio, line 677: Is this just so that the auth client can be passed in from `mutiny-wasm`? Maybe if that's required right now, we can just enforce that both auth manager and auth client be loaded in from there. Just don't like taking the inner of another object to use for something else.

mutiny-core/src/storage.rs
```diff
@@ -67,12 +67,22 @@ pub trait MutinyStorage: Clone + Sized + 'static {
     fn cipher(&self) -> Option<Cipher>;
 
     /// Set a value in the storage, the value will already be encrypted if needed
-    fn set<T>(&self, key: impl AsRef<str>, value: T) -> Result<(), MutinyError>
+    fn set<T>(
+        &self,
+        key: impl AsRef<str>,
+        value: T,
+        version: Option<u32>,
+    ) -> Result<(), MutinyError>
```
TonyGiorgio, line 75: Doesn't have to be part of this, but are we really calling both `set` and `set_data` in different places? Might want to refactor this later or make the names / comments more clear when to use what.

mutiny-core/src/vss.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub value: Option<Value>,
+    pub version: u32,
+}
+
+impl VssKeyValueItem {
+    /// Encrypts the value of the item using the encryption key
+    /// and returns an encrypted version of the item
+    pub(crate) fn encrypt(self, encryption_key: &SecretKey) -> EncryptedVssKeyValueItem {
+        // should we handle this unwrap better?
+        let bytes = self.value.unwrap().to_string().into_bytes();
```
TonyGiorgio, line 40: Why an option for `value`? Should we just enforce this?

mutiny-core/src/vss.rs
```diff
@@ -0,0 +1,273 @@
...
+        let iv: [u8; 16] = secp256k1::rand::random();
+
+        let cipher = Aes256CbcEnc::new(&encryption_key.secret_bytes().into(), &iv.into());
+        let mut encrypted: Vec<u8> = cipher.encrypt_padded_vec_mut::<Pkcs7>(&bytes);
+        encrypted.extend(iv);
+        let encrypted_value = base64::encode(encrypted);
```
TonyGiorgio, line 46: I think we're now using this encryption mechanism for here and SCB. Can we put this in `encrypt.rs`?

mutiny-core/src/vss.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    pub async fn list_key_versions(
+        &self,
+        key_prefix: Option<String>,
+    ) -> Result<Vec<VssKeyValueItem>, MutinyError> {
+        let url = Url::parse(&format!("{}/listKeyVersions", self.url)).map_err(|e| {
+            log_error!(self.logger, "Error parsing list key versions url: {e}");
+            MutinyError::Other(anyhow!("Error parsing list key versions url: {e}"))
+        })?;
```
TonyGiorgio, line 156: If this is just for keys and version numbers, should we create just a type for this then? That would solve the optional value problem too. Seems like it'll only be an option based on the results here.

TonyGiorgio (changes requested):
Some minor things but this looks as elegant as you could make it, at least for now. Very nice work, looking forward to testing this out!

TonyGiorgio:
Wait where is the channel manager being saved? Only saw some call outs for monitor.

mutiny-core/src/nodemanager.rs
benthecarman: Yeah we could, felt weird to have to two optional copies of it so I did it like this, could do it the other way too

mutiny-core/src/storage.rs
benthecarman: Yeah, `set_data` is what we use internally and `set` is what the user implements

mutiny-core/src/vss.rs
benthecarman: Because we reuse it for scanning keys, can probably just make its own struct for that function

benthecarman:
> Wait where is the channel manager being saved? Only saw some call outs for monitor.

Didn't do that because it didn't have an easy version number. We'll have to implement our own versioning with it.

mutiny-core/src/vss.rs
benthecarman: done, might be able to unify them between that and the password encryption too

TonyGiorgio:
> > Wait where is the channel manager being saved? Only saw some call outs for monitor.
> 
> Didn't do that because it didn't have an easy version number. We'll have to implement our own versioning with it.

I think we'll have to do this though in order for this remote cloud feature to be useful. Do you think that's attainable as part of this PR?

benthecarman:
> I think we'll have to do this though in order for this remote cloud feature to be useful. Do you think that's attainable as part of this PR?

done

mutiny-core/src/encrypt.rs
```diff
@@ -86,9 +95,36 @@ fn argon2() -> Argon2<'static> {
...
+    // split last 16 bytes off as iv
+    let iv = &bytes[bytes.len() - 16..];
+    let bytes = &bytes[..bytes.len() - 16];
+
+    let cipher = Aes256CbcDec::new(&encryption_key.secret_bytes().into(), iv.into());
+    let decrypted: Vec<u8> = cipher.decrypt_padded_vec_mut::<Pkcs7>(bytes).unwrap();
+
+    Ok(decrypted)
+}
+
```
TonyGiorgio, line 124: could write a test for these

mutiny-core/src/vss.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub value: String,
+    pub version: u32,
+}
+
+impl EncryptedVssKeyValueItem {
+    pub(crate) fn decrypt(self, encryption_key: &SecretKey) -> VssKeyValueItem {
+        let bytes = base64::decode(&self.value).unwrap();
+        let decrypted = decrypt_with_key(encryption_key, bytes).unwrap();
+        let decrypted_value = String::from_utf8(decrypted).unwrap();
+        let value = serde_json::from_str(&decrypted_value).unwrap();
```
TonyGiorgio, line 62: These should probably not `unwrap` and instead return a result.

mutiny-core/src/vss.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub key: String,
+    pub value: String,
+    pub version: u32,
+}
+
+impl EncryptedVssKeyValueItem {
+    pub(crate) fn decrypt(self, encryption_key: &SecretKey) -> VssKeyValueItem {
+        let bytes = base64::decode(&self.value).unwrap();
+        let decrypted = decrypt_with_key(encryption_key, bytes).unwrap();
+        let decrypted_value = String::from_utf8(decrypted).unwrap();
```
TonyGiorgio, line 61: if you're calling this: 
```
        let value = base64::encode(encrypt_with_key(encryption_key, &bytes));
```

shouldn't you be doing `base64::decode` ?

TonyGiorgio (changes requested):

TonyGiorgio (approved):

PR #646, opened 2023-07-06
Upgrade to ldk 116 alpha

TonyGiorgio:
Need to test more and manage deps better.

### DIFF
--- a/mutiny-core/src/keymanager.rs
+++ b/mutiny-core/src/keymanager.rs
@@ -12,14 +12,14 @@ use bitcoin::secp256k1::ecdsa::Signature;
 use bitcoin::secp256k1::{PublicKey, Scalar, Secp256k1, Signing};
 use bitcoin::util::bip32::{ChildNumber, DerivationPath, ExtendedPrivKey};
 use bitcoin::{Script, Transaction, TxOut};
-use lightning::chain::keysinterface::{
+use lightning::ln::msgs::{DecodeError, UnsignedGossipMessage};
+use lightning::ln::script::ShutdownScript;
+use lightning::log_warn;
+use lightning::sign::{
     EntropySource, InMemorySigner, KeyMaterial, NodeSigner,
     PhantomKeysManager as LdkPhantomKeysManager, Recipient, SignerProvider,
     SpendableOutputDescriptor,
 };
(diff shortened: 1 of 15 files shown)


mutiny-core/src/keymanager.rs
```diff
@@ -156,21 +157,21 @@ impl<S: MutinyStorage> SignerProvider for PhantomKeysManager<S> {
...
+        Ok(wallet
             .get_address(AddressIndex::New)
             .address
-            .script_pubkey()
+            .script_pubkey())
     }
 
-    fn get_shutdown_scriptpubkey(&self) -> ShutdownScript {
+    fn get_shutdown_scriptpubkey(&self) -> Result<ShutdownScript, ()> {
         let mut wallet = self.wallet.wallet.try_write().unwrap();
```
benthecarman, line 169: can map this to an err too

mutiny-core/src/keymanager.rs
```diff
@@ -156,21 +157,21 @@ impl<S: MutinyStorage> SignerProvider for PhantomKeysManager<S> {
         self.inner.read_chan_signer(reader)
     }
 
-    fn get_destination_script(&self) -> Script {
+    fn get_destination_script(&self) -> Result<Script, ()> {
         let mut wallet = self.wallet.wallet.try_write().unwrap();
```
benthecarman, line 161: should map this as an err

TonyGiorgio: This is a trait we are inheriting and the signature they're specifying.

benthecarman:
you should be able to change this to use update_channel_from_announcement_no_lookup and delete the dummy utxo lookup impl i made

[master/mutiny-core/src/peermanager.rs#L227](https://github.com/MutinyWallet/mutiny-node/blob/master/mutiny-core/src/peermanager.rs?rgh-link-date=2023-07-06T21%3A30%3A38Z#L227)

benthecarman:
can also remove this whole match statement and replace it with `bal.claimable_amount_satoshis`

https://github.com/MutinyWallet/mutiny-node/blob/f1c7ea0400c2a10d3a8a4fc96743a1d1536a498b/mutiny-core/src/nodemanager.rs#L1268

TonyGiorgio:
Added to changes.

benthecarman (approved):

Issue #194, opened 2022-12-11
Payjoin Support


benthecarman:
https://github.com/bitcoin/bips/blob/master/bip-0078.mediawiki

TonyGiorgio:
Need rust-bitcoin to support it or to use some of the rust payjoin libraries still being developed on.

DanGould:
Adding the ability to send payjoin from bdk using [rust-payjoin](https://crates.io/keywords/payjoin) was [~a 200 line change including test for BitMask](https://github.com/diba-io/bitmask-core/pull/121). Receiving payjoin can be done securely using a relay.

A second reference BDK x Payjoin implementation [will be completed for BDK-CLI](https://github.com/bitcoindevkit/bdk-cli/issues/149) as part of the summer of bitcoin.

Where does this fall on the roadmap? It does not seem blocked by library support as implied by the tag. I'm happy to elaborate as appropriate.

TonyGiorgio:
> Adding the ability to send payjoin from bdk using [rust-payjoin](https://crates.io/keywords/payjoin) was [~a 200 line change including test for BitMask](https://github.com/diba-io/bitmask-core/pull/121). Receiving payjoin can be done securely using a relay.
> 
> A second reference BDK x Payjoin implementation [will be completed for BDK-CLI](https://github.com/bitcoindevkit/bdk-cli/issues/149) as part of the summer of bitcoin.
> 
> Where does this fall on the roadmap? It does not seem blocked by library support as implied by the tag. I'm happy to elaborate as appropriate.

Very cool. When I was looking through issues on rust-bitcoin/bdk I was only seeing old things so I didn't see any support. I also wasn't sure how it was coming along with the payjoin libraries. I had only seen this one: https://github.com/Kixunil/payjoin which looked pretty out of date. 

Glad to have a pointer to yours, I think that would be really great to add. I think the only thing I see as a problem is that we use `no-std` `rust-bitcoin` and need `default-features = false` in any bitcoin library that we import due to wasm restrictions. If that's not a problem to add then I think it would be great to put into Mutiny.

DanGould:
> I think that would be really great to add. I think the only thing I see as a problem is that we use no-std rust-bitcoin and need default-features = false in any bitcoin library that we import due to wasm restrictions.

[...]

PR #647, opened 2023-07-07
Send payjoin

DanGould:
Here's an outline of what a payjoin implementation might look like. There are a few dangling pieces but the bulk of the happy path is here.

- [-] ~display [BIP 78 well-known errors](https://github.com/bitcoin/bips/blob/master/bip-0078.mediawiki#user-content-Receivers_well_known_errors) in UI and log the rest~ Errors are only logged but no json error from the receiver is parsed into an error to Display yet. Wait for library to update https://github.com/payjoin/rust-payjoin/pull/110 as in this design users will not be phished by an undisplayed message.
- [x] ensure BDK will sign & finalize an actual payjoin response
- [x] Test success case with[ mutiny-web UI draft](https://github.com/MutinyWallet/mutiny-web/pull/572)

A mutinynet payjoin receiver is live and funded when you're ready to test this out by pulling a new payjoin bip21 from https://mutiny.payjoin.org:4433/bip21

Note, this version of payjoin depends on rust-bitcoin v0.30.0 to use the new version of bip21 that disables default features for mutiny. So it duplicates that dependency until Mutiny upgrades.

tag #194, but I don't think it closes it. Send is 1/2 of the equation.

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -27,10 +27,11 @@ use crate::{gossip::*, scorer::HubPreferentialScorer};
 use crate::{labels::LabelStorage, subscription::MutinySubscriptionClient};
 use anyhow::anyhow;
 use bdk::chain::{BlockId, ConfirmationTime};
-use bdk::{wallet::AddressIndex, LocalUtxo};
+use bdk::{wallet::AddressIndex, FeeRate, LocalUtxo};
 use bitcoin::blockdata::script;
 use bitcoin::hashes::hex::ToHex;
 use bitcoin::hashes::{sha256, Hash};
+use bitcoin::psbt::PartiallySignedTransaction;
 use bitcoin::secp256k1::{rand, PublicKey};
 use bitcoin::util::bip32::ExtendedPrivKey;
 use bitcoin::{Address, Network, OutPoint, Transaction, Txid};
@@ -51,9 +52,11 @@ use lnurl::lnurl::LnUrl;
 use lnurl::{AsyncClient as LnUrlClient, LnUrlResponse, Response};
 use nostr::key::XOnlyPublicKey;
 use nostr::{EventBuilder, Keys, Kind, Tag, TagKind};
+use payjoin::{PjUri, PjUriExt};
 use reqwest::Client;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use std::io::Cursor;
 use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::{collections::HashMap, ops::Deref, sync::Arc};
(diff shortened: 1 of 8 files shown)


TonyGiorgio:
Awesome! Really appreciate that. 

I think from the frontend point of view, it just uses https://github.com/MutinyWallet/bitcoin-waila for parsing the bitcoin URI and giving the result of that `mutiny-node`. 

So I think just exposing this like it is here https://github.com/MutinyWallet/mutiny-node/blob/bf98cfc32c62122130ba582168eeca98e924a75d/mutiny-core/src/nodemanager.rs#L984-L996 & https://github.com/MutinyWallet/mutiny-node/blob/bf98cfc32c62122130ba582168eeca98e924a75d/mutiny-wasm/src/lib.rs#L216-L233

DanGould:
Would the payjoin Uri check be a better fit in mutiny-wasm, then?

```rs
        let uri = payjoin::Uri::try_from(uri)
            .expect("Invalid payjoin URI")
            .require_network(self.network)
            .expect("Payjoin network mismatch")
            .check_pj_supported()
            .expect("Payjoin not supported");
```

TonyGiorgio:
> Would the payjoin Uri check be a better fit in mutiny-wasm, then?
> 
> ```rust
>         let uri = payjoin::Uri::try_from(uri)
>             .expect("Invalid payjoin URI")
>             .require_network(self.network)
>             .expect("Payjoin network mismatch")
>             .check_pj_supported()
>             .expect("Payjoin not supported");
> ```

Oh yeah good point, I think that would be good. That way it can just be passed over to the `mutiny-core` 's `pay_payjoin` method.

benthecarman:
> Would the payjoin Uri check be a better fit in mutiny-wasm, then?
> 
> ```rust
>         let uri = payjoin::Uri::try_from(uri)
>             .expect("Invalid payjoin URI")
>             .require_network(self.network)
>             .expect("Payjoin network mismatch")
>             .check_pj_supported()
>             .expect("Payjoin not supported");
> ```

maybe this should be added to waila

DanGould:
it would not be too hard to export the payjoin uri bits as a standalone feature if you wanted to depend on it in waila instead of duplicating

DanGould:
This has been refactored to handle errors, sign in a BDK environment, produce spec-recommended fee contribution, and separate `onchain` behavior from `nodemanager` behavior. I'll have to figure out how to hook it up to mutiny-web to proceed with testing.

mutiny-wasm/src/lib.rs
```diff
@@ -346,6 +346,27 @@ impl MutinyWallet {
...
 
+    #[wasm_bindgen]
+    pub async fn send_payjoin(
+        &self,
+        payjoin_uri: String,
+        amount: u64,      /* override the uri amount if desired */
+        labels: &JsValue, /* Vec<String> */
+        fee_rate: Option<f32>,
+    ) -> Result<String, MutinyJsError> {
+        // I know walia parses `pj=` and `pjos=` but payjoin::Uri parses the whole bip21 uri
```
benthecarman, line 357: I think this is fine, for mutiny-web we can just check if the pj uri is defined, and if it is then we use `send_payjoin` instead of `send_to_address`

DanGould:
I did have to alter payjoin-cli to filter mixed input contribution with the following

```rs
let candidate_inputs: HashMap<Amount, OutPoint> = available_inputs
    .iter()
    .filter(|i| i.address.clone().is_some_and(|a| a.assume_checked().address_type() == Some(bitcoin::AddressType::P2tr)))
    .map(|i| (i.amount, OutPoint { txid: i.txid, vout: i.vout }))
    .collect();
```
But this was an issue with payjoin-cli (https://github.com/payjoin/rust-payjoin/issues/70), not this draft.

mutiny-core/src/nodemanager.rs
```diff
@@ -1060,6 +1063,79 @@ impl<S: MutinyStorage> NodeManager<S> {
...
+            .map_err(|_| MutinyError::PayjoinConfigError)?;
+        let original_psbt = self.wallet.create_signed_psbt(address, amount, fee_rate)?;
+
+        let payout_scripts = std::iter::once(uri.address.script_pubkey());
+        let fee_rate = if let Some(rate) = fee_rate {
+            FeeRate::from_sat_per_vb(rate)
+        } else {
+            let sat_per_kwu = self
+                .fee_estimator
+                .get_est_sat_per_1000_weight(ConfirmationTarget::Normal);
```
benalleng, line 1083: I ran into some type errors when trying to link for tests, it seems to think there is no such thing as `ConfirmationTarget::Normal`

benalleng: might have broke after the upgrade to LDK 0.0.118

benthecarman: there's now `fee_estimator.get_normal_fee_rate()`

benalleng: tACK `.get_normal_fee_rate()` works

mutiny-core/src/nodemanager.rs
```diff
@@ -1016,6 +1018,79 @@ impl<S: MutinyStorage> NodeManager<S> {
...
```
benalleng, line 1038: so this should be
```suggestion
                .fee_estimator.get_normal_fee_rate()
```

DanGould:
@benalleng and I fixed the Failed to send issue. We're having expected solid reliability now 😎 This should be ready for a review.

mutiny-wasm/src/lib.rs
```diff
@@ -451,6 +452,31 @@ impl MutinyWallet {
             .to_string())
     }
 
+    #[wasm_bindgen]
+    pub async fn send_payjoin(
+        &self,
+        payjoin_uri: String,
+        amount: u64,      /* override the uri amount if desired */
+        labels: &JsValue, /* Vec<String> */
```
benthecarman, line 460: this can be `labels: Vec<String>` now

benthecarman (commented):
Overall this looks good to me, might want to hold off on merging for now. @TonyGiorgio is doing a pretty large refactor in #867 that will effect this.

TonyGiorgio:
> Overall this looks good to me, might want to hold off on merging for now. @TonyGiorgio is doing a pretty large refactor in #867 that will effect this.

I can work around it, there's not much I think will conflict.

TonyGiorgio (approved):
Code looks solid! Thank you both for testing this thoroughly and getting it in here. Exciting!

PR #651, opened 2023-07-09
Fix empty password change

TonyGiorgio:


mutiny-wasm/src/lib.rs
```diff
@@ -1170,9 +1170,9 @@ impl MutinyWallet {
         old_password: Option<String>,
         new_password: Option<String>,
     ) -> Result<(), MutinyJsError> {
-        self.inner
-            .change_password(old_password, new_password)
-            .await?;
+        let old_p = old_password.filter(|p| !p.is_empty());
+        let new_p = new_password.filter(|p| !p.is_empty());
+        self.inner.change_password(old_p, new_p).await?;
```
benthecarman, line 1175: we should probably be doing this deeper too

TonyGiorgio: We should really just remove `Option<String>` to be honest. We do these checks in so many places....

TonyGiorgio: But no, I would rather all the incorrect string parsing logic only happen in the wasm layer.

benthecarman: idk it is a better abstraction and works fine, it is just weird for js

benthecarman: looked through it, seems like this is the best spot actually

benthecarman (approved):

PR #653, opened 2023-07-11
Restore lightning state from VSS

benthecarman:
This ended up being pretty simple, we just need to save our node storage so we can map all of our lightning state to the correct keys.


TonyGiorgio (approved):
LFG

PR #655, opened 2023-07-12
Handle error when LSP says amount too high

TonyGiorgio:


mutiny-core/src/lspclient.rs
```diff
@@ -156,6 +156,8 @@ impl LspClient {
                         return Err(MutinyError::LspFundingError);
                     } else if error_body.message.starts_with("Failed to connect to peer") {
                         return Err(MutinyError::LspConnectionError);
+                    } else if error_body.message == "Invoice amount is too high" {
```
benthecarman, line 159: would be slightly more efficient to have this as a match but no big deal

benthecarman (approved):

PR #657, opened 2023-07-12
Optimize reading from VSS

benthecarman:
Some optimizations:

- requests objects in parallel
- only requests objects we can parse
- uses a match to evaluate keys


TonyGiorgio (approved):
This diff is way more problematic to read than it should be, but it seems right.

benthecarman:
restoring state LN failed for me with, don't merge yet

TonyGiorgio (approved):

PR #658, opened 2023-07-12
Return bitcoin address as bip21 string

TonyGiorgio:


mutiny-core/src/nodemanager.rs
```diff
@@ -940,8 +948,8 @@ impl<S: MutinyStorage> NodeManager<S> {
         };
 
         Ok(MutinyBip21RawMaterials {
-            address,
-            invoice: bolt11,
+            address: address.address,
+            invoice: Some(bolt11),
```
benthecarman, line 952: should we just make this None if creating the invoice fails?

benthecarman:
Might make sense to just do this at the mutiny-wasm level instead

mutiny-core/src/nodemanager.rs
TonyGiorgio: No, there's specific errors that need to be returned so that paul can do the right thing or make the right warning.

TonyGiorgio:
> Might make sense to just do this at the mutiny-wasm level instead

done

benthecarman (approved):

PR #666, opened 2023-07-13
Wait for usable channel before paying invoice

benthecarman:
Before we were just waiting for a connection, we really need to wait for a usable channel

### DIFF
--- a/mutiny-core/src/node.rs
+++ b/mutiny-core/src/node.rs
@@ -930,6 +930,11 @@ impl<S: MutinyStorage> Node<S> {
             return Err(MutinyError::NonUniquePaymentHash);
         }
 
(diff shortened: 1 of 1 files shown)


mutiny-core/src/node.rs
```diff
@@ -937,7 +937,7 @@ impl<S: MutinyStorage> Node<S> {
             if self.stop.load(Ordering::Relaxed) {
                 return Err(MutinyError::NotRunning);
             }
-            if !self.peer_manager.get_peer_node_ids().is_empty() {
+            if !self.channel_manager.list_usable_channels().is_empty() {
```
benthecarman, line 940: Probably wanna do a check first that we have any channels at all.

PR #667, opened 2023-07-14
Retry spendable outputs individually

benthecarman:
This might help with missed swept force closes

### DIFF
--- a/mutiny-core/src/node.rs
+++ b/mutiny-core/src/node.rs
@@ -409,7 +409,23 @@ impl<S: MutinyStorage> Node<S> {
                         log_info!(logger, "Successfully retried spendable outputs");
                         persister.clear_failed_spendable_outputs()?;
                     }
-                    Err(e) => log_warn!(logger, "Failed to retry spendable outputs {e}"),
+                    Err(_) => {
+                        // retry them individually then only save failed ones
+                        // if there was only one we don't need to retry
+                        if retry_spendable_outputs.len() > 1 {
+                            let mut failed = vec![];
+                            for o in retry_spendable_outputs {
+                                if event_handler
+                                    .handle_spendable_outputs(&[o.clone()])
+                                    .await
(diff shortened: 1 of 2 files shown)


mutiny-core/src/node.rs
```diff
@@ -409,7 +409,16 @@ impl<S: MutinyStorage> Node<S> {
...
-                    Err(e) => log_warn!(logger, "Failed to retry spendable outputs {e}"),
+                    Err(_) => {
+                        // retry them individually then delete
+                        // if there was only one we don't need to retry
+                        if retry_spendable_outputs.len() > 1 {
+                            for o in retry_spendable_outputs {
+                                let _ = event_handler.handle_spendable_outputs(&[o.clone()]).await;
+                            }
+                        };
+                        persister.clear_failed_spendable_outputs()?;
```
TonyGiorgio, line 420: Why clear them when they failed?

mutiny-core/src/node.rs
```diff
@@ -409,7 +409,16 @@ impl<S: MutinyStorage> Node<S> {
                         log_info!(logger, "Successfully retried spendable outputs");
                         persister.clear_failed_spendable_outputs()?;
                     }
-                    Err(e) => log_warn!(logger, "Failed to retry spendable outputs {e}"),
+                    Err(_) => {
+                        // retry them individually then delete
+                        // if there was only one we don't need to retry
```
TonyGiorgio, line 414: why?

mutiny-core/src/node.rs
benthecarman: fixed

mutiny-core/src/node.rs
benthecarman: We were batching them all together, if one has already been swept but one not, then the whole thing would fail. This should help prevent that.

TonyGiorgio (approved):

PR #668, opened 2023-07-16
properly disconnect from relays

benthecarman:
eventually this would cause resource exhaustion

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -212,7 +212,8 @@ impl<S: MutinyStorage> MutinyWallet<S> {
                     pin_mut!(delay_fut);
                     select! {
                         notification = read_fut => {
-                                if let Ok(RelayPoolNotification::Event(_url, event)) = notification {
+                            match notification {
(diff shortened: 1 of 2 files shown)


mutiny-core/src/lib.rs
```diff
@@ -235,6 +235,10 @@ impl<S: MutinyStorage> MutinyWallet<S> {
                         }
                     }
                 }
+
```
TonyGiorgio, line 238: shouldn't the `read_fut` above handle `Err` so it breaks out of the loop and disconnects?

benthecarman: Don't see what you're talking about, looks like every error is handled, the disconnect is outside the inner `loop`

TonyGiorgio: when read errors happen it just keeps going inside the inner loop.

benthecarman: Ah I see, good call out

TonyGiorgio (approved):

PR #671, opened 2023-07-18
Prevent multiple devices with lock

benthecarman:
This creates a device specific lock that just uses VSS. Each device generates a uuid to identify itself and then when running every minute will send the current time to claim the lock. On boot it will check that this hasn't happened for 2 minutes so be sure it is safe to startup as no other device is running.

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -11,7 +11,7 @@ use crate::scb::{
     EncryptedSCB, StaticChannelBackup, StaticChannelBackupStorage,
     SCB_ENCRYPTION_KEY_DERIVATION_PATH,
 };
-use crate::storage::{MutinyStorage, KEYCHAIN_STORE_KEY};
+use crate::storage::{MutinyStorage, DEVICE_ID_KEY, KEYCHAIN_STORE_KEY};
 use crate::utils::sleep;
 use crate::MutinyWalletConfig;
 use crate::{
@@ -64,6 +64,7 @@ use std::str::FromStr;
 use uuid::Uuid;
 
 const BITCOIN_PRICE_CACHE_SEC: u64 = 300;
+pub const DEVICE_LOCK_INTERVAL_SECS: u64 = 60;
 
 // This is the NodeStorage object saved to the DB
 #[derive(Debug, Serialize, Deserialize, Clone, Default)]
@@ -555,6 +556,26 @@ impl<S: MutinyStorage> NodeManager<S> {
             .unwrap_or_else(|| String::from("wss://p.mutinywallet.com"));
 
         let logger = Arc::new(MutinyLogger::with_writer(stop.clone(), storage.clone()));
+
+        // Need to prevent other devices from running at the same time
+        if !c.skip_device_lock {
+            storage.set_device_lock()?;
+        }
+
+        let storage_clone = storage.clone();
+        let logger_clone = logger.clone();
+        let stop_clone = stop.clone();
+        utils::spawn(async move {
+            loop {
+                if stop_clone.load(Ordering::Relaxed) {
+                    break;
+                }
+                sleep((DEVICE_LOCK_INTERVAL_SECS * 1_000) as i32).await;
+                if let Err(e) = storage_clone.set_device_lock() {
+                    log_error!(logger_clone, "Error setting device lock: {e}");
+                }
+            }
+        });
 
         let esplora_server_url = get_esplora_url(c.network, c.user_esplora_url);
         let tx_sync = Arc::new(EsploraSyncClient::new(esplora_server_url, logger.clone()));
@@ -2257,7 +2278,8 @@ impl<S: MutinyStorage> NodeManager<S> {
         let serde_map = serde_json::map::Map::from_iter(map.into_iter().filter(|(k, _)| {
             // filter out logs and network graph
             // these are really big and not needed for export
-            !matches!(k.as_str(), LOGGING_KEY | NETWORK_GRAPH_KEY)
+            // filter out device id so a new one is generated
+            !matches!(k.as_str(), LOGGING_KEY | NETWORK_GRAPH_KEY | DEVICE_ID_KEY)
         }));
 
         // shut back down after reading if it was already closed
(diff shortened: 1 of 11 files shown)


mutiny-wasm/src/indexed_db.rs
```diff
@@ -270,6 +268,20 @@ impl IndexedDbStorage {
                     None => return Ok(Some((kv.key, obj.value))),
                 }
             }
+            DEVICE_LOCK_KEY => {
+                let obj = vss.get_object(&kv.key).await?;
+                // we can get version from device lock, so we should compare
+                match current.get_data::<DeviceLock>(&kv.key)? {
+                    Some(versioned) => {
+                        if versioned.time <= kv.version
```
TonyGiorgio, line 276: Comparing time with version?

mutiny-wasm/src/indexed_db.rs
```diff
@@ -270,6 +268,20 @@ impl IndexedDbStorage {
                     None => return Ok(Some((kv.key, obj.value))),
                 }
             }
+            DEVICE_LOCK_KEY => {
```
TonyGiorgio, line 271: Why get the device key from the remote server? We shouldn't be replacing our remote key, right? We just want to compare our device id from the device id from remote?

mutiny-core/src/nodemanager.rs
```diff
@@ -556,6 +557,26 @@ impl<S: MutinyStorage> NodeManager<S> {
...
+
+        let storage_clone = storage.clone();
+        let logger_clone = logger.clone();
+        let stop_clone = stop.clone();
+        utils::spawn(async move {
+            loop {
+                if stop_clone.load(Ordering::Relaxed) {
+                    break;
+                }
+                sleep((DEVICE_LOCK_INTERVAL_SECS * 1_000) as i32).await;
```
TonyGiorgio, line 573: Each second seems like too much, if we're doing 2m limits then we can probably just do 30s here.

TonyGiorgio (changes requested):

TonyGiorgio:
I think there's still an open question on how this is going to work with the iOS PWA flow that @futurepaul is wanting. I'm still pretty unsure about the friction proposed for it.

mutiny-core/src/nodemanager.rs
benthecarman: this is every minute, `DEVICE_LOCK_INTERVAL_SECS` is 60 and time 1k to go from seconds to milliseconds

mutiny-wasm/src/indexed_db.rs
benthecarman: we use the time as the vss version number so it is always increasing

mutiny-wasm/src/indexed_db.rs
benthecarman: This isn't the device id but the struct containing the currently locked device + time it was locked.

mutiny-wasm/src/indexed_db.rs
TonyGiorgio: Ah interesting. There's no number conversion issues then right? I remember we had to do some number conversions before.

benthecarman: yeah both u32

benthecarman:
> I think there's still an open question on how this is going to work with the iOS PWA flow that @futurepaul is wanting

Yeah if they immediately restore from seed from the browser to PWA they'd be locked out for 2 minutes. Maybe we can do something like "transfer" and in there we send to VSS that the device has stopped running so it doesn't initially lock them out

TonyGiorgio:
> > I think there's still an open question on how this is going to work with the iOS PWA flow that @futurepaul is wanting
> 
> Yeah if they immediately restore from seed from the browser to PWA they'd be locked out for 2 minutes. Maybe we can do something like "transfer" and in there we send to VSS that the device has stopped running so it doesn't initially lock them out

Oh yeah that's a good one. I like that. Maybe if delete is called, it will delete the vss entry? I think that's pretty clean if you wanted to apply that here.

benthecarman:
> Oh yeah that's a good one. I like that. Maybe if delete is called, it will delete the vss entry? I think that's pretty clean if you wanted to apply that here.

added

TonyGiorgio:
Awesome. I'll test this tonight, seems like it'll work pretty cleanly. Excited about it.

TonyGiorgio (changes requested):
The delete functionality is either not working here or the frontend is not allowing enough time for that to apply. Everything else about this works pretty well.

mutiny-core/src/storage.rs
```diff
@@ -234,6 +253,22 @@ pub trait MutinyStorage: Clone + Sized + 'static {
...
+    async fn delete_all(&self) -> Result<(), MutinyError> {
+        Self::clear().await?;
+        // remove lock from VSS if is is enabled
+        if self.vss_client().is_some() {
+            let device = self.get_device_id()?;
+            // set time to 0 to unlock
+            let lock = DeviceLock { time: 0, device };
+            // still update the version so it is written to VSS
+            let time = now().as_secs() as u32;
+            self.set_data(DEVICE_LOCK_KEY, lock, Some(time))?;
```
TonyGiorgio, line 266: why does `set_data` want the current time?

benthecarman: we use the time as the version

benthecarman:
> The delete functionality is either not working here or the frontend is not allowing enough time for that to apply. Everything else about this works pretty well.

The front end uses importing an empty json object for delete, will need to change it to use this new delete_all function

mutiny-core/src/storage.rs
TonyGiorgio: Isn't that the problem then? 

> if versioned.time <= kv.version

benthecarman: here devicelock.time version is 0, but we are still setting the vss version number to 0

TonyGiorgio: this is all incredibly confusing. version is time and time is version and one is set to 0 but not the other.

TonyGiorgio:
Okay, seems to work pretty well with some changes to the frontend. Nice work!

TonyGiorgio (approved):

PR #674, opened 2023-07-22
Update bdk + ldk to latest releases

benthecarman:
We are now fork free!


mutiny-core/src/fees.rs
```diff
@@ -176,6 +176,7 @@ fn num_blocks_from_conf_target(confirmation_target: ConfirmationTarget) -> usize
     match confirmation_target {
         // Background is VERY lax and may never confirm if used directly
         // it is only meant for lower ranges of transaction to enter mempool
+        ConfirmationTarget::MempoolMinimum => 1008,
```
TonyGiorgio, line 179: nit: the comment above is meant for the `Background`, though I am now curious what this new one is supposed to be used for.

benthecarman: Moved the comment and added one for `MempoolMinimum`

TonyGiorgio (approved):

PR #669, opened 2023-07-16
Move VSS into mutiny-core

benthecarman:
Closes #652

To fix the backwards compatibility I had to change the channel monitor encoding (luckily we wanted to get this to hex). This _should_ make it so worst case that if a user with a password never opens to get the the update, they will get a restore with a force close.

We should heavily test this

PR #675, opened 2023-07-24
Remove unused encoding for channel monitor

benthecarman:
In #669 I played around with trying a different encoding for the channel monitors but decided against it since we don't actually locally encrypt those so they weren't effected by the bug, looks like I accidentally left the handling of both encodings for when decoding so we were getting confusing logs like `Failed to read channel monitors, trying old encoding`.

This reverts the ldkstorage file to what is was before that PR minus a log I added for when we read channel managers.

How I tested to make sure there is no break:

 - create wallet on master
 - make payments
 - upgrade to this PR, make sure can still make payments

also did the same with `860907c` (commit before #669)

### DIFF
--- a/mutiny-core/src/ldkstorage.rs
+++ b/mutiny-core/src/ldkstorage.rs
@@ -123,56 +123,20 @@ impl<S: MutinyStorage> MutinyNodePersister<S> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


TonyGiorgio (approved):
Feel free to ignore the mac failed build for now.

PR #678, opened 2023-07-26
Log LSP errors

### LINKED ISSUE #677: Log LSP errors when creating invoice
Doesn't look like they are getting logged when an `invalid amount` error is being returned to user for whatever reason. https://github.com/MutinyWallet/mutiny-web/issues/395

benthecarman:
Closes #677

### DIFF
--- a/mutiny-core/src/node.rs
+++ b/mutiny-core/src/node.rs
@@ -708,8 +708,13 @@ impl<S: MutinyStorage> Node<S> {
         if let Some(lsp) = self.lsp_client.clone() {
             self.connect_peer(PubkeyConnectionInfo::new(&lsp.connection_string)?, None)
                 .await?;
-            let lsp_invoice_str = lsp.get_lsp_invoice(invoice.to_string()).await?;
-            let lsp_invoice = Bolt11Invoice::from_str(&lsp_invoice_str)?;
+            let lsp_invoice = match lsp.get_lsp_invoice(invoice.to_string()).await {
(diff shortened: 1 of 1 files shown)


mutiny-core/src/node.rs
```diff
@@ -708,8 +708,13 @@ impl<S: MutinyStorage> Node<S> {
         if let Some(lsp) = self.lsp_client.clone() {
             self.connect_peer(PubkeyConnectionInfo::new(&lsp.connection_string)?, None)
                 .await?;
-            let lsp_invoice_str = lsp.get_lsp_invoice(invoice.to_string()).await?;
-            let lsp_invoice = Bolt11Invoice::from_str(&lsp_invoice_str)?;
+            let lsp_invoice = match lsp.get_lsp_invoice(invoice.to_string()).await {
+                Ok(lsp_invoice_str) => Bolt11Invoice::from_str(&lsp_invoice_str)?,
+                Err(e) => {
+                    log_error!(self.logger, "Failed to get invoice from LSP: {e}");
+                    return Err(MutinyError::InvoiceCreationFailed);
```
TonyGiorgio, line 715: actually can you just return the same error? I believe it does handle some of the errors inside of the LSP and we need to return those properly.

TonyGiorgio (changes requested):

mutiny-core/src/node.rs
benthecarman: oh fixed, thought it was a non-mutiny error

TonyGiorgio (approved):

PR #679, opened 2023-07-27
Pull in remote scorer

TonyGiorgio:
One of these days we will do a proper builder, today is not one of those days.

### DIFF
--- a/mutiny-core/src/gossip.rs
+++ b/mutiny-core/src/gossip.rs
@@ -12,13 +12,14 @@ use lightning::{
 };
 use lightning::{log_debug, log_error, log_info, log_warn};
 use reqwest::Client;
+use reqwest::{Method, Url};
 use serde::{Deserialize, Serialize};
 
-use crate::error::MutinyError;
 use crate::logging::MutinyLogger;
 use crate::node::{NetworkGraph, ProbScorer, RapidGossipSync};
 use crate::storage::MutinyStorage;
 use crate::utils;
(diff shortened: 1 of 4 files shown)


mutiny-core/src/gossip.rs
```diff
@@ -64,40 +83,82 @@ async fn get_gossip_data(
...
+        scorer,
     };
 
     Ok(Some(gossip))
 }
 
+/// Scorer is the scorer that gets pulled remotely
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Scorer {
+    pub value: Vec<u8>,
```
benthecarman, line 139: Would be better if we encode as hex or even better base64 to reduce bandwidth.

TonyGiorgio: Encoding will increase bandwidth

benthecarman: It's written as [1, 2, 3,..] in the json so hex is more effienct

TonyGiorgio: Damn you're right actually. I guess I can get rid of all the serialization completely then and just deal with raw bytes, no object?

benthecarman: Yeah that'd be best I guess

TonyGiorgio: I spent far too long messing with stupid byte responses that wouldn't properly decode, kept the json but swapped for base64.

benthecarman (approved):

PR #681, opened 2023-07-28
Sync contacts from nostr

benthecarman:
Need to make an easy way to test but I think this will work


mutiny-core/src/labels.rs
```diff
@@ -34,10 +36,42 @@ pub struct Contact {
...
     #[serde(skip_serializing_if = "Option::is_none")]
+    pub image_url: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub archived: Option<bool>,
     pub last_used: u64,
 }
 
+impl Contact {
+    /// Update the contact with metadata from their Nostr profile
+    /// Todo can add more data like profile picture
```
benthecarman, line 47: Remove this todo

mutiny-wasm/src/lib.rs
```diff
@@ -1170,6 +1172,13 @@ impl MutinyWallet {
         Ok(())
     }
 
+    /// Pay the subscription invoice. This will post a NWC automatically afterwards.
```
TonyGiorgio, line 1175: nit: wrong comment

TonyGiorgio (changes requested):

mutiny-core/src/lib.rs
```diff
@@ -308,6 +311,58 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+        let client = Client::new(&keys);
+
+        #[cfg(target_arch = "wasm32")]
+        client.add_relay("wss://relay.damus.io").await.unwrap();
+
+        #[cfg(not(target_arch = "wasm32"))]
+        client
+            .add_relay("wss://relay.damus.io", None)
+            .await
+            .unwrap();
```
TonyGiorgio, line 330: If we're doing more nostr things, we should probably make relay(s) configurable. Can be done in a follow up later this month if needed.

TonyGiorgio (approved):

PR #684, opened 2023-07-31
Change invoice expiry to 1 hour

benthecarman:
IIRC the Voltage LSP has limits around the invoice expiry, we may need to handle that here.

For https://github.com/MutinyWallet/mutiny-web/issues/416

### DIFF
--- a/mutiny-core/src/node.rs
+++ b/mutiny-core/src/node.rs
@@ -768,7 +768,7 @@ impl<S: MutinyStorage> Node<S> {
                     amount_msat,
                     description,
                     now,
-                    1500,
+                    3600,
                     Some(40),
                 )
             }
@@ -776,7 +776,7 @@ impl<S: MutinyStorage> Node<S> {
                 amount_msat,
                 None,
                 description,
-                1500,
+                3600,
                 r,
                 self.keys_manager.clone(),
                 self.keys_manager.clone(),


benthecarman:
Tested and on signet was able to receive successfully. One that opened a channel and one that worked with the channel

TonyGiorgio:
I'll double check the conditions. It definitely will reject invoices with expiry too high.

TonyGiorgio:
3600 is the max right now in the LSP. Do we want to just make that the hardcoded value instead of making it configurable?

TonyGiorgio:
wow, i'm so sorry. I accidentally clicked a button trying to move up my screen

benthecarman:
> 3600 is the max right now in the LSP. Do we want to just make that the hardcoded value instead of making it configurable? 

Maybe allow configurable but throw an error if over the limit

TonyGiorgio:
> > 3600 is the max right now in the LSP. Do we want to just make that the hardcoded value instead of making it configurable?
> 
> Maybe allow configurable but throw an error if over the limit

Is there any reason why it should be any lower than an hour? I think the max can just be the default. It would still solve the problem in the issue, I think they were wanting it to be an hour too.

benthecarman:
Okay just changed to 1 hour time

TonyGiorgio (approved):

PR #685, opened 2023-07-31
Lower gap limit of on-chain wallet

benthecarman:
We are sending a lot of requests when syncing, this should speed up syncs and reduce some of the overhead


mutiny-core/src/onchain.rs
```diff
@@ -115,7 +115,7 @@ impl<S: MutinyStorage> OnChainWallet<S> {
                 spks,
                 core::iter::empty(),
                 core::iter::empty(),
-                50,
+                25,
```
TonyGiorgio, line 118: nit: how about 20? I think that's been the norm in general.

TonyGiorgio (approved):

PR #686, opened 2023-07-31
Skip invalid invoices for nwc

benthecarman:
skips invoices with no amount and invoices we've already paid

### DIFF
--- a/mutiny-core/src/nostr/nwc.rs
+++ b/mutiny-core/src/nostr/nwc.rs
@@ -160,6 +160,22 @@ impl NostrWalletConnect {
                 return Ok(None);
             }
 
+            // if the invoice has no amount, we cannot pay it
+            if invoice.amount_milli_satoshis().is_none() {
+                log_warn!(
+                    node_manager.logger,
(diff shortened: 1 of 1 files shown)


mutiny-core/src/nostr/nwc.rs
```diff
@@ -162,6 +162,22 @@ impl NostrWalletConnect {
...
+                // if the invoice has no amount, we cannot pay it
+                if invoice.amount_milli_satoshis().is_none() {
+                    log_warn!(
+                        node_manager.logger,
+                        "NWC Invoice amount not set, cannot pay: {invoice}"
+                    );
+                    return Ok(None);
+                }
+
+                // if we have already paid this invoice, skip it
```
TonyGiorgio, line 174: This should probably apply to both auto or manual nwc requests

TonyGiorgio (changes requested):

mutiny-core/src/nostr/nwc.rs
benthecarman: good call

TonyGiorgio (approved):

PR #688, opened 2023-08-02
Multi esplora client

benthecarman:
This makes it so we can use multiple esplora clients instead of relying on a single one. This randomly selects between a set of them when making a request.

This essentially is just a wrapper for the esplora client. I did need to copy paste the bdk syncing code since it was was not implemented to handle this new type.

Tested by setting `VITE_ESPLORA="https://mempool.space/api https://blockstream.info/api"` for mutiny-web and had network requests going to both (but no duplicates)


TonyGiorgio:
code looks good, though I'm unsure if we still want this or not.

TonyGiorgio (approved):

PR #690, opened 2023-08-04
More effienct esplora syncing

benthecarman:
Replaces the current syncing with a more efficient version copied from https://github.com/bitcoindevkit/bdk/pull/1040

Keeps the old syncing mechanism that is used for the reset keychain tracker so we can get txs if they are missed somehow.

Needed to add a flag to do a full sync after a restore so we can find our on-chain balances


benthecarman:
need to fix for restores

TonyGiorgio:
if the "reset chain state" button is clicked, does it run through full sync again? should it delete `NEED_FULL_SYNC_KEY` or set to false?

benthecarman:
the reset keychain tracker just deleted the bdk history so added it so it sets need full sync

TonyGiorgio (approved):

PR #691, opened 2023-08-05
more ben: Add nwc profile to vss

bennyhodl:
Saves created NWC profiles to vss.

Version is updated on creating new NWC profiles in `indexed_db`. Could use some direction on checking the vss storage.

### DIFF
--- a/mutiny-core/src/nostr/mod.rs
+++ b/mutiny-core/src/nostr/mod.rs
@@ -21,7 +21,7 @@ use nostr::{Event, EventBuilder, EventId, Filter, Keys, Kind, Tag};
 use nostr_sdk::{Client, RelayPoolNotification};
 use std::ops::Deref;
 use std::str::FromStr;
-use std::sync::atomic::Ordering;
+use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


TonyGiorgio:
Nice. We should eventually have more things in VSS but kept it to critical data. However I can see NWC being annoying to keep adding if someone reset their wallet. We want to change to a more deterministic key generation approach for NWC based on the profile name, but haven't gotten around to that yet. We may have to do that first before this, since doing so will break all NWCs.

Quick code base looks alright, I'll have to let @benthecarman comment on the specifics since he has more experience with these two things.

bennyhodl:
@benthecarman rebased and squashed to one commit

mutiny-core/src/nostr/mod.rs
```diff
@@ -60,6 +78,8 @@ pub struct NostrManager<S: MutinyStorage> {
     pub storage: S,
     /// Lock for pending nwc invoices
     pending_nwc_lock: Arc<Mutex<()>>,
+    // Versioned nwc profiles for VSS. Option for backwards compatability.
+    nwc_profile_version: Option<NwcProfileVersion>,
```
benthecarman, line 82: When this is `None` it is never set to be `Some` after it is updated. Probably what would be best is to not have this be optional and on init, if we don't have it, create a new one with version=0

mutiny-core/src/nostr/mod.rs
```diff
@@ -568,12 +608,15 @@ impl<S: MutinyStorage> NostrManager<S> {
             .map(|profile| NostrWalletConnect::new(&context, xprivkey, profile).unwrap())
             .collect();
 
+        let nwc_profile_version = NwcProfileVersion::new();
```
benthecarman, line 611: this needs to be read from storage, otherwise you'll always be starting with version 0

benthecarman (changes requested):
This still also doesn't handle backwards compat for reading the nwc configs from _before_ this PR. 

Also looks like there are some more places where we write the nwc storage without doing vss

PR #695, opened 2023-08-07
Don't persist empty keychain store

benthecarman:


TonyGiorgio (changes requested):
This didn't work. chain state wasn't fully empty but still wiped. We need to revert the commit that introduces full sync.

Screenshot from 2023-08-07 11-33-51

PR #697, opened 2023-08-07
Get contacts' zaps

benthecarman:
Not sure the best way to test, maybe having paul hookup a simple UI?

This should get the most recent 100 zaps received by your contacts, however, it'll exclude any that aren't sent by your own contact.


mutiny-wasm/src/lib.rs
```diff
@@ -1210,6 +1210,12 @@ impl MutinyWallet {
         Ok(())
     }
 
+    /// Gets all the Zaps that have been received by our contacts
+    pub async fn get_contact_zaps(&self) -> Result<(), MutinyJsError> {
+        self.inner.get_contact_zaps(None).await?;
+        Ok(())
+    }
```
benthecarman, line 1217: woops, need to actually return it

PR #698, opened 2023-08-08
Fix BDK needing resync

benthecarman:
Finally found the reason why we would always need to resync bdk. We were removing the bdk keychain from memory after reading so we didnt overload memory, however, we needed this because we read the current bdk keychain and merge with the update, if we remove it from our in memory cache, then we have nothing to merge with.

This should let us bring back #690


TonyGiorgio (approved):
good find, yeah that makes sense.

PR #699, opened 2023-08-08
support different arch in justfile

bennyhodl:
Linux arch is hard coded in the `justfile`. Adds an attribute for system arch for the correct target.


TonyGiorgio (approved):
Works for me, thanks!

PR #702, opened 2023-08-08
Ensure mutiny+ nwc is active for current subs

TonyGiorgio:


mutiny-core/src/lib.rs
```diff
@@ -280,39 +302,39 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+            // profile with the reserved index does not exist, create a new one
+            let profile = self
+                .nostr
+                .create_new_nwc_profile(
+                    ProfileType::Reserved(ReservedProfile::MutinySubscription),
+                    SpendingConditions::RequireApproval,
+                )
+                .await?;
+            // only should have to submit the NWC if never created locally before
+            subscription_client.submit_nwc(profile.nwc_uri).await?;
```
benthecarman, line 333: it is possible we've submitted but lost it locally, will this overwrite?

TonyGiorgio: it will overwrite, but it should be the same nwc string because it's kept to a specific index.

benthecarman (approved):

PR #704, opened 2023-08-08
WIP nostr filter debugging

TonyGiorgio:
This was all I ended up doing when I was investigating it before. 

The other issue is that something is causing the web sockets to open twice. I think it's a nostr sdk thing. Will keep progressing on this.


mutiny-core/src/lib.rs
```diff
@@ -220,16 +220,29 @@ impl<S: MutinyStorage> MutinyWallet<S> {
 
                 add_relay_res.expect("Failed to add relays");
                 client.connect().await;
-                client.subscribe(nostr.get_nwc_filters()).await;
+
+                let mut last_filters = nostr.get_nwc_filters();
```
benthecarman, line 224: Just having this as mut and changing in place won't cause the filters to change for the subscription I don't think. Could just have it break and then it'll reconnect.

benthecarman (changes requested):

mutiny-core/src/lib.rs
TonyGiorgio: What do you mean? The new filters are sent across. I'm working on the blastr side to make sure filters are handle more efficiently.

benthecarman: Oh I missed that you were calling subscribe again. I guess that works. Maybe we want to cancel the previous subscription, guess it's not a big deal if we don't.

benthecarman (approved):

PR #705, opened 2023-08-09
Handle bad metadata when syncing nostr contacts

benthecarman:
There is a bug in the nostr-sdk with parsing some user's metadata. This gets around it by skipping those profiles instead of erroring out on the entire sync. To do this I needed to copy some code from the nostr-sdk. Also better handled the errors instead of unwrapping.


TonyGiorgio (approved):
Looks good though I have to ask, should we just publish our own version of nostr-sdk since it'll be weeks till they release another? I'm sure there's a lot of improvements and there's bugs I'm seeing that might be fixed too.

benthecarman:
I think it's fine for now. If we have to work around more bugs then maybe.

PR #708, opened 2023-08-09
Allow anon zaps to lnurls

benthecarman:
While we still haven't figured out how to best do the keys for sending zaps (since we don't have the user's nsec), this will send an anonymous zap to the user

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -59,6 +59,8 @@ use lightning::{log_debug, log_error, log_info, log_warn};
 use lightning_invoice::{Bolt11Invoice, Bolt11InvoiceDescription};
 use lnurl::lnurl::LnUrl;
 use lnurl::{AsyncClient as LnUrlClient, LnUrlResponse, Response};
+use nostr::key::XOnlyPublicKey;
+use nostr::{EventBuilder, Keys, Kind, Tag, TagKind};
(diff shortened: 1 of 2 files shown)


mutiny-core/src/nodemanager.rs
```diff
@@ -1638,14 +1640,37 @@ impl<S: MutinyStorage> NodeManager<S> {
         from_node: &PublicKey,
         lnurl: &LnUrl,
         amount_sats: u64,
+        zap_npub: Option<XOnlyPublicKey>,
```
TonyGiorgio, line 1643: which npub is this? the wallet owner or the reciever? And how should it be retrieved?

benthecarman: the receiver, I figured we'd just be pulling it from the contacts list

TonyGiorgio: is this somehow not combined with lnurl in a way? I thought zaps were just an lnurl extension and the receiver pubkey was part of that somehow.

benthecarman: No, in the lnurl response it has the pubkey of the key that will sign the zap receipt, but that is just so you can verify that it is not a fake zap

TonyGiorgio (approved):
Need to make a draft PR afterwards to `mutiny-web` before we forget

PR #711, opened 2023-08-14
Handle gifting errors

benthecarman:
Makes it so `claim_single_use_nwc` will wait until the invoice is paid, given a timeout. Will also return an error string if one occurs

### DIFF
--- a/mutiny-core/src/nostr/mod.rs
+++ b/mutiny-core/src/nostr/mod.rs
@@ -1,22 +1,26 @@
 use crate::nodemanager::NodeManager;
 use crate::nostr::nwc::{
-    NostrWalletConnect, NwcProfile, PendingNwcInvoice, Profile, SingleUseSpendingConditions,
-    SpendingConditions, PENDING_NWC_EVENTS_KEY,
+    NostrWalletConnect, NwcProfile, NwcProfileTag, PendingNwcInvoice, Profile,
+    SingleUseSpendingConditions, SpendingConditions, PENDING_NWC_EVENTS_KEY,
 };
 use crate::storage::MutinyStorage;
+use crate::utils;
 use crate::{error::MutinyError, utils::get_random_bip32_child_index};
-use bitcoin::hashes::sha256;
+use bitcoin::hashes::hex::{FromHex, ToHex};
+use bitcoin::hashes::{sha256, Hash};
 use bitcoin::secp256k1::{PublicKey, Secp256k1, Signing};
 use bitcoin::util::bip32::{ChildNumber, DerivationPath, ExtendedPrivKey};
+use futures::{pin_mut, select, FutureExt};
 use futures_util::lock::Mutex;
 use nostr::key::SecretKey;
 use nostr::nips::nip47::*;
-use nostr::prelude::{encrypt, XOnlyPublicKey};
+use nostr::prelude::{decrypt, encrypt, XOnlyPublicKey};
 use nostr::{Event, EventBuilder, EventId, Filter, Keys, Kind, Metadata, Tag};
-use nostr_sdk::Client;
+use nostr_sdk::{Client, RelayPoolNotification};
 use std::collections::HashMap;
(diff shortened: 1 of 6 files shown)


TonyGiorgio (approved):

benthecarman:
Need to add url suffix and gift amt to the json

futurepaul (approved):
haven't read the code but it works on my machine (https://github.com/MutinyWallet/mutiny-web/pull/421)

mutiny-core/src/nostr/nwc.rs
```diff
@@ -210,15 +237,19 @@ impl NostrWalletConnect {
...
+                    let content = if single_use.spent {
+                        Response {
+                            result_type: Method::PayInvoice,
+                            error: Some(NIP47Error {
+                                code: ErrorCode::QuotaExceeded,
+                                message: "Already Claimed".to_string(),
+                            }),
+                            result: None,
+                        }
+                    } else if msats <= single_use.amount_sats * 1_000 {
```
TonyGiorgio, line 252: I think `todo handle timeout errors` need to be handled. I think there's restart risk with in flight payments being redeemed and the NWC profile isn't saved because that happens only here.

Like maybe check inflight payments somehow and see if there's any associated with this specific NWC profile?

TonyGiorgio (changes requested):

mutiny-core/src/nostr/mod.rs
```diff
@@ -247,7 +265,7 @@ impl<S: MutinyStorage> NostrManager<S> {
             amount_sats,
             spent: false,
         });
-        self.create_new_nwc_profile(profile, spending_conditions)
+        self.create_new_nwc_profile(profile, spending_conditions, NwcProfileTag::Gift)
```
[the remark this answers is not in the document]
benthecarman: Yeah that was my thinking

mutiny-core/src/nostr/nwc.rs
benthecarman: We currently check if we've paying an invoice already. We only allow retrying if the payment status is failed

https://github.com/MutinyWallet/mutiny-node/blob/5c8288695d520b4e8252c86e2580cba4eb1aea2d/mutiny-core/src/node.rs#L945

mutiny-core/src/nostr/mod.rs
```diff
@@ -485,28 +504,112 @@ impl<S: MutinyStorage> NostrManager<S> {
...
+                                            let preimage: Vec<u8> = FromHex::from_hex(&result.preimage)?;
+                                            if sha256::Hash::hash(&preimage) == invoice.payment_hash {
+                                                return Ok(None);
+                                            } else {
+                                                return Ok(Some(NIP47Error {
+                                                    code: ErrorCode::QuotaExceeded,
+                                                    message: "Already Claimed".to_string(),
+                                                }));
+                                            }
+                                        },
```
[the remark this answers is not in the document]
benthecarman: Changed to a log

mutiny-core/src/nostr/nwc.rs
TonyGiorgio: It's not about paying the same invoice, it's about another NWC message coming in with a new invoice. If the payment stalls out, or it completes outside of this process running, then the NWC profile is never marked as paid. Another NWC message can come in with a new invoice and repeat the process.

benthecarman: Yeah this is defnitely an issue but not sure the best way to go about it without having to tie this into the node's event manager.

For now just having it log and disable the nwc profile on payment timeout to prevent double spends.

TonyGiorgio: Okay, that's a good approach for now. I think long term we can save pending NWC invoices and have those get checked before attempting to pay another for the same NWC. We will need that for auto approval w/ budget too.

TonyGiorgio (approved):

PR #713, opened 2023-08-16
Sync RGS in background

benthecarman:
Another attempt to speed up startup times. Moves the syncing of RGS to after startup in our background sync thread. I think this will work this time since we've fixed the on-chain sync being so spammy.

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -21,6 +21,7 @@ use crate::{
     esplora::EsploraSyncClient,
     fees::MutinyFeeEstimator,
     gossip,
+    gossip::{fetch_updated_gossip, get_rgs_url},
     logging::MutinyLogger,
     lspclient::LspClient,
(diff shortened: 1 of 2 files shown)


mutiny-core/src/nodemanager.rs
```diff
@@ -1356,6 +1358,37 @@ impl<S: MutinyStorage> NodeManager<S> {
...
+                    &self.gossip_sync,
+                    &self.storage,
+                    &self.logger,
+                )
+                .await;
+
+                if fetch_result.is_err() {
+                    log_warn!(self.logger, "Failed to fetch updated gossip data");
+                }
+            }
```
TonyGiorgio, line 1389: I don't think we should be fetching gossip every minute. This works well in the background but it should just run once.

TonyGiorgio (changes requested):

TonyGiorgio (approved):

PR #714, opened 2023-08-16
Only read from vss keys with better version numbers

benthecarman:
This should reduce the number of reads we do on startup for VSS. This will compare the version numbers before we request the data from the server instead of after. This way we only download the data from VSS if it has a higher version number

### DIFF
--- a/mutiny-wasm/src/indexed_db.rs
+++ b/mutiny-wasm/src/indexed_db.rs
@@ -255,36 +255,42 @@ impl IndexedDbStorage {
 
         match kv.key.as_str() {
             NODES_KEY => {
-                let obj = vss.get_object(&kv.key).await?;
                 // we can get version from node storage, so we should compare
                 match current.get_data::<NodeStorage>(&kv.key)? {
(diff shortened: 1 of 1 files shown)


mutiny-wasm/src/indexed_db.rs
```diff
@@ -255,36 +255,41 @@ impl IndexedDbStorage {
...
             NODES_KEY => {
-                let obj = vss.get_object(&kv.key).await?;
                 // we can get version from node storage, so we should compare
                 match current.get_data::<NodeStorage>(&kv.key)? {
                     Some(versioned) => {
-                        if versioned.version <= kv.version
-                            && serde_json::from_value::<NodeStorage>(obj.value.clone()).is_ok()
-                        {
-                            return Ok(Some((kv.key, obj.value)));
+                        if versioned.version < kv.version {
```
TonyGiorgio, line 261: can `versioned.version` be renamed? it's very confusing. it sounds like it should be called `local.version`.

TonyGiorgio (approved):

PR #718, opened 2023-08-16
Adds support for LSPs using the new jit-channel specification

johncantrell97:
Allows for both the current voltage-style and the new LSPS-style LSPs.  For now just added two new env vars that let you set lsp_connection_string (comma separated) and lsp_token (comma separated) similar to how lsp_url currently works.

Had to create a composed custom message handler that supports both the LSPS messages and your existing SCB messages.

Introduced an LSP trait that abstracts the fee and invoice negotiations so the node doesn't have to care what style LSP it is using.

Introduced a LspConfig enum to support both style LSPs.  Voltage style only needs an url to reach them at but LSPS supports node connection string and an optional token.

Instead of storing the lsp_url in the NodeIndex I am now storing the entire LspConfig as we'd need more information to rebuild the lsp client.  This is still a bit of a todo because it will break existing users.  Either need to move the config to a separate field and handle both versions in the code for now or handle an automatic migration where we try to read the url first and store it as a LspConfig::Voltage(lsp_url).

Still some cleanup required but it's working.


mutiny-core/src/nodemanager.rs
```diff
@@ -82,7 +82,7 @@ pub struct NodeStorage {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
 pub struct NodeIndex {
     pub child_index: u32,
-    pub lsp: Option<String>,
+    pub lsp: Option<LspConfig>,
```
benthecarman, line 85: This won't be backwards compatible, could just rename to `lsp_config` or write a custom parser. Should be fine to change to `lsp_config` as there hasn't been any reason to change your LSP as of yet (voltage is only option)

mutiny-core/src/peermanager.rs
```diff
@@ -323,7 +323,7 @@ pub(crate) async fn connect_peer_if_necessary<S: MutinyStorage>(
         // first check to see if the fee rate is mostly up to date
         // if not, we need to have updated fees or force closures
         // could occur due to UpdateFee message conflicts.
-        fee_estimator.update_fee_estimates_if_necessary().await?;
+        // fee_estimator.update_fee_estimates_if_necessary().await?;
```
benthecarman, line 326: I'm guessing this is just from testing, should be able to bring back. Maybe just put this behind `if self.network != Network::Regtest` this could be a problem for other people too.

mutiny-core/src/nodemanager.rs
```diff
@@ -631,30 +631,41 @@ impl<S: MutinyStorage> NodeManager<S> {
...
         };
 
+        if let Some(lsp_connection_string) = c.lsp_connection_string.clone() {
+            if !lsp_connection_string.is_empty() {
+                let lsp_connection_strings: Vec<&str> = lsp_connection_string.split(',').collect();
+                let lsp_token = c.lsp_token.unwrap_or_default();
+                let lsp_tokens: Vec<&str> = lsp_token.split(',').collect();
+
+                for (index, connection_string) in lsp_connection_strings.into_iter().enumerate() {
+                    lsp_configs.push(LspConfig::LSPS(LspsConfig { connection_string: connection_string.to_string(), token: lsp_tokens.get(index).map(|t| t.to_string()) }));
```
benthecarman, line 664: Don't think will work properly if you have say 3 lsps and you don't need a token for the 2nd. Not sure the best way to handle this.

mutiny-core/src/nodemanager.rs
```diff
@@ -631,30 +631,41 @@ impl<S: MutinyStorage> NodeManager<S> {
...
+                // results
+                //     .into_iter()
+                //     .flat_map(|res| match res {
+                //         Ok(client) => Some(AnyLsp::Voltage(client)),
+                //         Err(e) => {
+                //             log_warn!(logger, "Error starting up lsp client: {e}");
+                //             None
+                //         }
+                //     })
+                //     .collect()
```
benthecarman, line 652: you handled this in `node.rs` so should be safe to delete

mutiny-core/src/lsp/lsps.rs
```diff
@@ -0,0 +1,228 @@
...
+                            let lsp_route_hint = RouteHint(vec![RouteHintHop {
+                                src_node_id: counterparty_node_id,
+                                short_channel_id: scid,
+                                fees: RoutingFees { base_msat: 0, proportional_millionths: 0 },
+                                cltv_expiry_delta: cltv_expiry_delta.try_into().unwrap(),
+                                htlc_minimum_msat: None,
+                                htlc_maximum_msat: None,
+                            }]);
+
+                            let mut invoice = InvoiceBuilder::new(self.network.into())
```
benthecarman, line 116: would it not be better to use `create_invoice_from_channelmanager_and_duration_since_epoch` here?

mutiny-core/src/lsp/lsps.rs
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+
+    pub(crate) fn compute_opening_fee(&self, payment_size_msat: u64, opening_fee_min_fee_msat: u64, opening_fee_proportional: u64) -> Option<u64> {
+        payment_size_msat
+            .checked_mul(opening_fee_proportional)
+            .and_then(|f| f.checked_add(999999))
+            .and_then(|f| f.checked_div(1000000))
+            .map(|f| std::cmp::max(f, opening_fee_min_fee_msat))
+    }
```
benthecarman, line 156: this function confuses me, adding docs to how it works would be nice

benthecarman (commented):
Really awesome! Left some review on some things I noticed. Overall really great through, wasn't too hard to review

mutiny-core/src/lsp/lsps.rs
johncantrell97: can't because we need to add the lsp route hint

mutiny-core/src/lsp/lsps.rs
johncantrell97: yeah, added docs and opened a PR to lightning-liquidity to expose it -- it's part of the spec

PR #719, opened 2023-08-17
Make nwc keys random

TonyGiorgio:
Thought about it and there's really no reason to have determinism with these nwc keys. As long as we save the random child key index, then it should be fine. When these are stored remotely then it should be fine as well.

This allows for backwards compat with existing keys that are based on `index` as before. So no need to recreate NWC profiles.


benthecarman:
needs formatting

benthecarman (approved):

PR #722, opened 2023-08-21
Log node id and tx ids

benthecarman:
After people send us their logs a lot of the time we have to ask for node ids and txids, we should just log them so we have them right away

### DIFF
--- a/mutiny-core/src/node.rs
+++ b/mutiny-core/src/node.rs
@@ -41,7 +41,7 @@ use lightning::{
 use crate::multiesplora::MultiEsploraClient;
 use bitcoin::util::bip32::ExtendedPrivKey;
 use lightning::ln::PaymentSecret;
-use lightning::sign::{EntropySource, InMemorySigner};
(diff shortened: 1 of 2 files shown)


mutiny-core/src/chain.rs
```diff
@@ -52,6 +52,7 @@ impl<S: MutinyStorage> BroadcasterInterface for MutinyChain<S> {
         let logger = self.logger.clone();
         utils::spawn(async move {
             for tx in txs_clone {
+                log_info!(logger, "Broadcasting transaction: {}", tx.txid());
```
TonyGiorgio, line 55: actually tx hex would be useful too, to see what it is trying to spent in case it has inputs that can't be spent yet, or at all.

TonyGiorgio (changes requested):

mutiny-core/src/chain.rs
benthecarman: added

TonyGiorgio (approved):

PR #723, opened 2023-08-21
Add channel manager and monitors to used once

benthecarman:
Should give us slightly better guarantees around getting the best version of the manager and monitors.

Also added them to `used_once` so they won't be in memory twice, should help with the out of memory issues


TonyGiorgio (approved):
it worked for ben

PR #734, opened 2023-08-29
Auto approve NWC with budget

### LINKED ISSUE #724: Track pending NWC invoices and lookup before spending more
Need a way to track which NWC's have had payment attempts before so we can check the status of those payments. This is needed to safely handle one-time NWC's that may have a long timeout, and then also auto approval with overall budget.

Follow up for: https://github.com/MutinyWallet/mutiny-node/pull/711

Currently we just disable the one time use profile if it has a payment timeout. For the use case of mutiny-to-mutiny in person gifts, this should be fine for now.

benthecarman:
Closes #724

First commit fixes some types up to make it easier to lookup invoices without having to convert between `PaymentHash` and `sha256::Hash` which were used interchangeably 

2nd commit is the meat. This adds the auto approval with a budget. The idea is we save payments that we've made (or attempted) over the given time period. The most complicated part is handling payment timeouts. To handle that we save the payment hash of invoices and when we are checking our budget, we only filter out ones that are explicitly failed, making sure inflight payments are counted to budget.

test with https://github.com/MutinyWallet/mutiny-web/pull/501

### DIFF
--- a/mutiny-core/src/nostr/mod.rs
+++ b/mutiny-core/src/nostr/mod.rs
@@ -1,7 +1,8 @@
 use crate::nodemanager::NodeManager;
 use crate::nostr::nwc::{
-    NostrWalletConnect, NwcProfile, NwcProfileTag, PendingNwcInvoice, Profile,
-    SingleUseSpendingConditions, SpendingConditions, PENDING_NWC_EVENTS_KEY,
+    BudgetPeriod, BudgetedSpendingConditions, NostrWalletConnect, NwcProfile, NwcProfileTag,
+    PendingNwcInvoice, Profile, SingleUseSpendingConditions, SpendingConditions,
+    PENDING_NWC_EVENTS_KEY,
 };
 use crate::storage::MutinyStorage;
 use crate::utils;
@@ -19,7 +20,6 @@ use nostr::nips::nip47::*;
 use nostr::prelude::{decrypt, encrypt};
 use nostr::{Event, EventBuilder, EventId, Filter, Keys, Kind, Tag};
 use nostr_sdk::{Client, RelayPoolNotification};
-use std::ops::Deref;
 use std::str::FromStr;
 use std::sync::atomic::Ordering;
 use std::sync::{Arc, RwLock};
@@ -133,6 +133,41 @@ impl<S: MutinyStorage> NostrManager<S> {
             .ok_or(MutinyError::NotFound)?;
 
         nwc.profile = profile.profile();
+
+        let nwc_profile = nwc.nwc_profile();
+
+        // save to storage
+        {
+            let profiles = profiles
+                .iter()
+                .map(|x| x.profile.clone())
+                .collect::<Vec<_>>();
+            self.storage.set_data(NWC_STORAGE_KEY, profiles, None)?;
+        }
+
+        Ok(nwc_profile)
+    }
+
+    pub fn set_nwc_profile_budget(
+        &self,
+        profile_index: u32,
+        budget_sats: u64,
+        budget_period: BudgetPeriod,
+        single_max_sats: Option<u64>,
+    ) -> Result<NwcProfile, MutinyError> {
+        let mut profiles = self.nwc.write().unwrap();
+
+        let nwc = profiles
+            .iter_mut()
+            .find(|nwc| nwc.profile.index == profile_index)
+            .ok_or(MutinyError::NotFound)?;
+
+        nwc.profile.spending_conditions = SpendingConditions::Budget(BudgetedSpendingConditions {
+            budget: budget_sats,
+            single_max: single_max_sats,
+            payments: vec![],
+            period: budget_period,
+        });
 
         let nwc_profile = nwc.nwc_profile();
 
@@ -449,37 +484,32 @@ impl<S: MutinyStorage> NostrManager<S> {
         };
 
         if let Some(mut nwc) = nwc {
-            let (event, needs_save) = nwc
-                .handle_nwc_request(
-                    event,
-                    node_manager,
-                    from_node,
-                    self.pending_nwc_lock.deref(),
-                )
+            let event = nwc
+                .handle_nwc_request(event, node_manager, from_node, self)
                 .await?;
-
-            // update the profile if needed
-            if needs_save {
-                let mut vec = self.nwc.write().unwrap();
-
-                // update the profile
-                for item in vec.iter_mut() {
-                    if item.profile.index == nwc.profile.index {
-                        item.profile = nwc.profile;
-                        break;
-                    }
-                }
-
-                let profiles = vec.iter().map(|x| x.profile.clone()).collect::<Vec<_>>();
-                drop(vec); // drop the lock, no longer needed
-
-                self.storage.set_data(NWC_STORAGE_KEY, profiles, None)?;
-            }
-
             Ok(event)
         } else {
             Ok(None)
         }
+    }
+
+    pub(crate) fn save_nwc_profile(&self, nwc: NostrWalletConnect) -> Result<(), MutinyError> {
+        let mut vec = self.nwc.write().unwrap();
+
+        // update the profile
+        for item in vec.iter_mut() {
+            if item.profile.index == nwc.profile.index {
+                item.profile = nwc.profile;
+                break;
+            }
+        }
+
+        let profiles = vec.iter().map(|x| x.profile.clone()).collect::<Vec<_>>();
+        drop(vec); // drop the lock, no longer needed
+
+        self.storage.set_data(NWC_STORAGE_KEY, profiles, None)?;
+
+        Ok(())
     }
 
     pub async fn claim_single_use_nwc(
--- a/mutiny-core/src/nostr/mod.rs
+++ b/mutiny-core/src/nostr/mod.rs
@@ -266,16 +266,16 @@ impl<S: MutinyStorage> NostrManager<S> {
             }
             // Ensure normal profiles start from 1000
             ProfileType::Normal { name } => {
-                let normal_profiles_count = profiles
+                let next_index = profiles
                     .iter()
                     .filter(|&nwc| nwc.profile.index >= USER_NWC_PROFILE_START_INDEX)
-                    .count() as u32;
+                    .max_by(|a, b| a.profile.index.cmp(&b.profile.index))
+                    .map(|nwc| nwc.profile.index + 1)
+                    .unwrap_or(USER_NWC_PROFILE_START_INDEX);
 
-                (
-                    name,
-                    normal_profiles_count + USER_NWC_PROFILE_START_INDEX,
-                    Some(get_random_bip32_child_index()),
-                )
+                debug_assert!(next_index >= USER_NWC_PROFILE_START_INDEX);
+
+                (name, next_index, Some(get_random_bip32_child_index()))
             }
         };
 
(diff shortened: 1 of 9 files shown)


mutiny-core/src/nostr/nwc.rs
```diff
@@ -333,6 +381,101 @@ impl NostrWalletConnect {
...
+                                }
+                                Err(e) => {
+                                    // if the payment timed out, we should add it to the budget
+                                    // in case it is inflight and succeeds later
+                                    if let MutinyError::PaymentTimeout = e {
+                                        budget.add_payment(&invoice);
+                                        self.profile.spending_conditions =
+                                            SpendingConditions::Budget(budget);
+                                        needs_save = true;
+                                    }
```
TonyGiorgio, line 454: I think saving needs to happen first before the payment is attempted. I'm sure that may screw up how persistance of NWC profiles happens today but someone that opens the app, pulls down a pending nwc, then leaves before the 30s timeout will not have this counted towards their budget.

mutiny-core/src/nostr/nwc.rs
```diff
@@ -27,11 +29,57 @@ pub struct SingleUseSpendingConditions {
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct BudgetedSpendingConditions {
+    /// Amount in sats for the allotted budget period
+    pub budget: u64,
+    /// Max amount in sats for a single payment
+    pub single_max: Option<u64>,
+    /// Payment history
+    pub payments: Vec<TrackedPayment>,
```
TonyGiorgio, line 49: We need some sort of lock around tracked payments and adding / removing them. I don't know if it's just a bug in our pending approval list or not but those are more miss than hit at the moment, when there's multiple involved.

mutiny-core/src/nostr/nwc.rs
```diff
@@ -27,11 +29,57 @@ pub struct SingleUseSpendingConditions {
...
+    /// Time period in seconds
+    pub period: u64,
```
TonyGiorgio, line 51: Is the idea that it is over the last X amount of seconds? So instead of a "monthly" budget, it's more of a over the last 30 days kind of budget?

mutiny-wasm/src/lib.rs
```diff
@@ -1126,6 +1126,31 @@ impl MutinyWallet {
             .into())
     }
 
+    /// Create a budgeted nostr wallet connect profile
+    #[wasm_bindgen]
+    pub async fn create_budget_nwc_profile(
```
TonyGiorgio, line 1131: Is there already a method for editing these? changing the budget / max use for a given profile?

TonyGiorgio (changes requested):
Mostly concerns around persisting lists, persisting before payment is attempted, and I think once we get something final we should also apply it to the one time use nwc as well, but that can always happen in a follow up.

TonyGiorgio:
Have you thought about how the UX for this will look like at all? Maybe we should talk to @futurepaul on it as well. I think we should start pulling in pending items (at least outgoing payments) into the activity list, otherwise there will be no indication payments are getting attempted behind the scenes. Otherwise, not sure if it makes sense to also show these automated payments in the pending approvals location (with a slight name to maybe pending payments or something).

mutiny-core/src/nostr/nwc.rs
benthecarman: It is currently safe because we don't process NWCs concurrently but yeah, probably want that just to be safe

mutiny-core/src/nostr/nwc.rs
benthecarman: Yeah exactly, for testing I was using a 60 second budget but I am expecting for us to expose like daily options

mutiny-wasm/src/lib.rs
benthecarman: Yeah, might be a bit difficult for the front end to use, maybe can add a set budget option

mutiny-core/src/nostr/nwc.rs
benthecarman: Good point, added

benthecarman:
> Have you thought about how the UX for this will look like at all? Maybe we should talk to @futurepaul on it as well. I think we should start pulling in pending items (at least outgoing payments) into the activity list, otherwise there will be no indication payments are getting attempted behind the scenes. Otherwise, not sure if it makes sense to also show these automated payments in the pending approvals location (with a slight rename to maybe pending payments or something).

Yeah I like this idea, maybe we should just add all in-flight payments to the activity. Could do that in a follow up

sudocarlos:
maybe just have a single activity stream and display a pill next to each transactions with a one-word description like "pending", "sent", "received". and tapping any of the pills filters the activity stream to just those types of transactions. add a little filter cone at the top right and call it a day.

TonyGiorgio:
Needs rebase

mutiny-core/src/nostr/nwc.rs
```diff
@@ -200,8 +252,8 @@ impl NostrWalletConnect {
         event: Event,
         node_manager: &NodeManager<S>,
         from_node: &PublicKey,
-        pending_nwc_lock: &Mutex<()>,
-    ) -> anyhow::Result<(Option<Event>, bool)> {
+        nostr_manager: &NostrManager<S>,
+    ) -> anyhow::Result<Option<Event>> {
```
TonyGiorgio, line 256: Nit: remove the comment above saying "if the profile needs to be saved to disk"

mutiny-core/src/nostr/nwc.rs
TonyGiorgio: This can be done in a follow up for later, but we'd also need to ensure some of the same guarantees for gifting too.

mutiny-wasm/src/lib.rs
TonyGiorgio: Still think we need this.

TonyGiorgio (changes requested):
Code looks good, I double checked the upgrade path on a few NWC profiles and those seem to be working still. Not sure how I'm supposed to test this yet and I don't see any tests here. I think we still need some NWC profile update functionality, but I wasn't sure what I needed to stick into my storage to test the budget capabilities.

TonyGiorgio:
Trying to test this with the frontend PR but it never triggers. I think there needs to also be a way to see what the current spending has been too. Perhaps this is as far as the PR can go without @futurepaul trying to implement it and seeing how it goes.

mutiny-core/src/nostr/mod.rs
```diff
@@ -148,6 +147,41 @@ impl<S: MutinyStorage> NostrManager<S> {
         Ok(nwc_profile)
     }
 
+    pub fn set_nwc_profile_budget(
+        &self,
+        profile_index: u32,
+        budget_sats: u64,
+        budget_period_secs: u64,
+        single_max_sats: Option<u64>,
+    ) -> Result<NwcProfile, MutinyError> {
```
TonyGiorgio, line 156: What if we want to remove it from budget to not budget? Can the update method be more generic? 

It would be great if people could set whether or not mutiny+ is auto approval as well and they could convert it to manual or disable at any time.

benthecarman: we have `edit_nwc_profile` where we could just overwrite the spending conditions. Not sure how to make it much easier without making a function for every spending condition

mutiny-core/src/nostr/nwc.rs
TonyGiorgio: I think it would be good to do something that is interval based instead of past X time. If it were specific then you could do accurate monthly invoicing on the server side. Past 30 days doesn't really map to any real world logic. We would have this problem with Mutiny+ since we do 3 days early. We should remove that and be more precise and match that logic here. Would allow Mutiny+ to be auto approval with budget if a user wanted. Or other services.

TonyGiorgio:
Made a comment last week, I think it would be best if we did like hourly, daily, weekly, monthly intervals instead of past X time: https://github.com/MutinyWallet/mutiny-node/pull/734#discussion_r1319397930

benthecarman:
> Made a comment last week, I think it would be best if we did like hourly, daily, weekly, monthly intervals instead of past X time: [#734 (comment)](https://github.com/MutinyWallet/mutiny-node/pull/734#discussion_r1319397930)

Done, harder to test without waiting a week though :joy:

TonyGiorgio:
> > Made a comment last week, I think it would be best if we did like hourly, daily, weekly, monthly intervals instead of past X time: [#734 (comment)](https://github.com/MutinyWallet/mutiny-node/pull/734#discussion_r1319397930)
> 
> Done, harder to test without waiting a week though 😂

Do you want to implement a mock test timer? :P

benthecarman:
Added tests :)

mutiny-core/src/nostr/nwc.rs
```diff
@@ -27,11 +29,100 @@ pub struct SingleUseSpendingConditions {
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum BudgetPeriod {
+    Day,
+    Week,
+    Month,
+    Year,
+    Seconds(u64),
+}
```
TonyGiorgio, line 49: Would be good to have some docs here about how it works. For instance, from reading the code, these reset at each interval right? 

- Budget resets at midnight
- Budget resets on sunday
- Budget resets on Sept 1
- Budget reset on Jan 1

Seconds it's a bit different right? It checks the last X seconds?

TonyGiorgio (commented):
Code looks good, just a doc nit. Still need to test it a bit. 

Do we need to update any of the other gifting stuff with this or save that for another PR to refactor how that works when we start getting close to releasing that again.

mutiny-core/src/nostr/mod.rs
```diff
@@ -183,6 +183,34 @@ impl<S: MutinyStorage> NostrManager<S> {
         Ok(nwc_profile)
     }
 
+    pub fn set_nwc_profile_require_approval(
```
TonyGiorgio, line 186: these are helpers? I believe there's a generic method that you can override any profile IIRC from a conversation with @benthecarman

TonyGiorgio:
If we move from a disabled auto approved to enabled auto approved nwc, it will pull in all of the NWC messages that may have been queued up during the disabled time and pays them automatically. Maybe we should have some sort of time period where we don't go back to that gets reset whenever disabled goes to enabled?

sudocarlos:
wouldn't want to pay queued zaps automatically. make those manual approve

TonyGiorgio:
> wouldn't want to pay queued zaps automatically. make those manual approve

yeah agreed they shouldn't be queued up automatically. It's probably too steep of a feature right now to make those turn into manual approval and might be riddled with confusion as to why they are there at all. Would rather nothing show up for the user on NWC's that were marked as disabled during the time they were disabled. It was disabled for a reason.

benthecarman:
Updates so now we no longer have archiving / disabling NWC profiles, now there is only delete.

Also now uses similar logic for budget NWC for gifts.

benthecarman:
Need to make it so we can't have overlapping profile indexes after a delete

benthecarman:
> Need to make it so we can't have overlapping profile indexes after a delete

Fixed

TonyGiorgio (approved):

sudocarlos:
👀 lfg

PR #736, opened 2023-08-31
Add support for all currencies

benalleng:
My first rust PR!

should add support for us to chose any possible currency.

As for the types, I wasn't sure if adding more restrictive types was the move here, and to be honest I'm not sure exactly how to limit it to only available currencies at coingecko.

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -2281,7 +2281,7 @@ impl<S: MutinyStorage> NodeManager<S> {
     }
 
     /// Gets the current bitcoin price in USD.
-    pub async fn get_bitcoin_price(&self) -> Result<f32, MutinyError> {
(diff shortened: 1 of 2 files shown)


mutiny-core/src/nodemanager.rs
```diff
@@ -2316,27 +2316,33 @@ impl<S: MutinyStorage> NodeManager<S> {
         Ok(price)
     }
 
-    async fn fetch_bitcoin_price(&self) -> Result<f32, MutinyError> {
-        log_debug!(self.logger, "fetching new bitcoin price");
+    async fn fetch_bitcoin_price(&self, fiat: String) -> Result<f32, MutinyError> {
```
benthecarman, line 2319: you can make this `fiat: &str` and then you change `fiat.clone()` on line 2298 to `&fiat`

benthecarman:
Looks like you also need to run `cargo fmt`

benthecarman (approved):

PR #745, opened 2023-09-03
Fix reserve amount error calculation

benthecarman:
`unspendable_punishment_reserve` is given in satoshis but everything else in the function we are comparing with `ln_balance` and `amt_msats`  are in millisats. This should fix so we get the reserve error when it is actually applicable


TonyGiorgio (approved):
Nice catch

Issue #710, opened 2023-08-10
Memory access out of bounds 


TonyGiorgio:
I think I'm seeing this more now with the in memory bdk stuff

```
caught (in promise) RuntimeError: memory access out of bounds
    at mutiny_core::multiesplora::MultiEsploraClient::get_block_hash::{{closure}}::he6608baf4ab56223 (mutiny_wasm_bg.wasm:0x1491fd7)
    at <mutiny_core::multiesplora::MultiEsploraClient as bdk_esplora::async_ext::EsploraAsyncExt>::scan::{{closure}}::hdc43222d4e2120a4 (mutiny_wasm_bg.wasm:0x2c8b8f)
    at <core::pin::Pin<P> as core::future::future::Future>::poll::he25aaf69944f8ea4 (mutiny_wasm_bg.wasm:0x21442a7)
    at bdk_esplora::async_ext::EsploraAsyncExt::scan_without_keychain::{{closure}}::hc27d5967bb951037 (mutiny_wasm_bg.wasm:0x1311589)
    at <core::pin::Pin<P> as core::future::future::Future>::poll::he25aaf69944f8ea4 (mutiny_wasm_bg.wasm:0x21442a7)
    at <mutiny_core::multiesplora::MultiEsploraClient as bdk_esplora::async_ext::EsploraAsyncExt>::scan::{{closure}}::hdc43222d4e2120a4 (mutiny_wasm_bg.wasm:0x2d2452)
    at <core::pin::Pin<P> as core::future::future::Future>::poll::he25aaf69944f8ea4 (mutiny_wasm_bg.wasm:0x21442a7)
    at bdk_esplora::async_ext::EsploraAsyncExt::scan_without_keychain::{{closure}}::hc27d5967bb951037 (mutiny_wasm_bg.wasm:0x1311589)
    at <core::pin::Pin<P> as core::future::future::Future>::poll::he25aaf69944f8ea4 (mutiny_wasm_bg.wasm:0x21442a7)
    at <mutiny_core::multiesplora::MultiEsploraClient as bdk_esplora::async_ext::EsploraAsyncExt>::scan::{{closure}}::hdc43222d4e2120a4 (mutiny_wasm_bg.wasm:0x2d2452)
$
```

benthecarman:
Probably related: https://github.com/rustwasm/wasm-bindgen/issues/3130

benthecarman:
[...]

PR #746, opened 2023-09-03
Use a reference to JsValue in mutiny-wasm

benthecarman:
Maybe helps with #710 

First commit changes to use `&JsValue` for all our function commands, this seems like it is the standard thing to do for our usecase.

https://rustwasm.github.io/docs/wasm-bindgen/reference/reference-types.html

Second commit enables weak refs for wasm-bindgen, this seems like it will really help. From my understanding, this will make it so after something is returned from rust to the browser, before it would wait for us to call `free()` on it to get rid of it (which we never do). With this it leaves it available to the js garbage collector.

https://rustwasm.github.io/docs/wasm-bindgen/reference/weak-references.html

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -69,7 +69,7 @@ jobs:
       - name: Build wasm
         env:
           RUSTUP_TOOLCHAIN: nightly
-        run: wasm-pack build ./mutiny-wasm --release --target web --scope mutinywallet
+        run: wasm-pack build ./mutiny-wasm --release --weak-refs --target web --scope mutinywallet
 
       - name: Publish wasm
         run: wasm-pack publish --access public -t web
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,7 +62,7 @@ jobs:
       - name: Build wasm package
         env:
           RUSTUP_TOOLCHAIN: nightly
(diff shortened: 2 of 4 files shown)


TonyGiorgio:
> Currently this feature is supported in Firefox 79+ and Chrome. Support in other browsers is likely coming soon!

What about safari? Maybe we need @futurepaul to test this.

benthecarman:
> > Currently this feature is supported in Firefox 79+ and Chrome. Support in other browsers is likely coming soon!
> 
> What about safari? Maybe we need @futurepaul to test this.

supported everywhere [developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef) now

TonyGiorgio:
Seems to work. though while profiling this, I noticed waves of gradual spikes every 15 seconds that span ~1MB in size. Seems like it could be related to the background process since that's the only thing that is running at the top of the spike, but I don't see anything that calls every 15s, so maybe it's just garbage collection?

Screenshot from 2023-09-04 12-39-48

Not sure, but not related to this PR.

TonyGiorgio (approved):

PR #754, opened 2023-09-07
Remove unused file

benthecarman:
I think this is left over from removing the front end


TonyGiorgio:
CI does not like you removing it lol

benthecarman:
Looks like rust update broke CI

benthecarman:
merging this because it has the fix for CI too

PR #757, opened 2023-09-08
Windowless sleep

futurepaul:
web-sys calls the "window" object for setTimeout, but that's not actually necessary (setTimeout is available in js environments without a window, like node). Also, it seems like the web-sys console_log is only used for tests, so with that we can move the web-sys dep entirely to dev-dependencies

this all seems to work fine in browser but I can't figure out how to run the tests on my mac

### DIFF
--- a/mutiny-wasm/src/utils.rs
+++ b/mutiny-wasm/src/utils.rs
@@ -27,16 +27,15 @@ pub fn now() -> Duration {
         .unwrap()
 }
 
-#[allow(dead_code)]
 pub async fn sleep(millis: i32) {
-    let mut cb = |resolve: js_sys::Function, _reject: js_sys::Function| {
-        web_sys::window()
-            .unwrap()
-            .set_timeout_with_callback_and_timeout_and_arguments_0(&resolve, millis)
-            .unwrap();
-    };
-    let p = js_sys::Promise::new(&mut cb);
-    wasm_bindgen_futures::JsFuture::from(p).await.unwrap();
+    #[cfg(target_arch = "wasm32")]
+    {
+        windowless_sleep::sleep(millis).await;
+    }
+    #[cfg(not(target_arch = "wasm32"))]
+    {
+        tokio::time::sleep(Duration::from_millis(millis.try_into().unwrap())).await;
+    }
 }
 
 #[cfg(test)]
--- a/mutiny-wasm/src/utils.rs
+++ b/mutiny-wasm/src/utils.rs
@@ -1,5 +1,3 @@
-use core::time::Duration;
-use instant::SystemTime;
 use log::{debug, Level};
 use wasm_bindgen::prelude::*;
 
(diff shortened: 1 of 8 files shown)


benthecarman:
Is there a way we can do this without having to add javascript, or at least a library that does this for us?

benthecarman:
This could be worth trying instead

https://docs.rs/async-std/latest/async_std/task/fn.sleep.html

TonyGiorgio:
Agreed, no js. Make it into a rust crate and we'll add it :P 

> web-sys calls the "window" object for setTimeout,

Why is this a problem though?

benthecarman:
For Javascript environments that aren't a browser

futurepaul:
> This could be worth trying instead
> 
> https://docs.rs/async-std/latest/async_std/task/fn.sleep.html

does this work with wasm? tried looking at the impl but I couldn't quite follow it

futurepaul:
> Agreed, no js. Make it into a rust crate and we'll add it :P
> 
> > web-sys calls the "window" object for setTimeout,
> 
> Why is this a problem though?

yeah this is helpful for WASI or node contexts, but also there's no "window" object in a web worker

benthecarman:
> does this work with wasm? tried looking at the impl but I couldn't quite follow it

idk, was hoping you'd test :sweat_smile:

futurepaul:
> idk, was hoping you'd test 😅

lol it's not if this tracking issue is real https://github.com/async-rs/async-std/issues/220

futurepaul:
windowless_sleep = "0.1.1"

TonyGiorgio:
Is this good to merge? It would be good to remove local storage.

benthecarman:
Should be okay

mutiny-wasm/src/utils.rs
```diff
@@ -27,16 +27,15 @@ pub fn now() -> Duration {
...
-    let p = js_sys::Promise::new(&mut cb);
-    wasm_bindgen_futures::JsFuture::from(p).await.unwrap();
+    #[cfg(target_arch = "wasm32")]
+    {
+        windowless_sleep::sleep(millis).await;
+    }
+    #[cfg(not(target_arch = "wasm32"))]
+    {
+        tokio::time::sleep(Duration::from_millis(millis.try_into().unwrap())).await;
+    }
```
TonyGiorgio, line 38: This should only be targetting wasm32, so I'm not sure why tokio stuff is in here.

TonyGiorgio (changes requested):

mutiny-wasm/src/utils.rs
benthecarman: we should be able to just get rid of this and re-export from mutiny core

TonyGiorgio (approved):

TonyGiorgio:
Needs rebase.

PR #758, opened 2023-09-08
Cache previously chosen fiat

benalleng:
This will allow currencies to be switched before cache timeout / expiry as it checks to see the value of the previous fiat as compared to the arg when get_bitcoin_price is run


benthecarman (changes requested):
I think a better way to do this is to make the cache a `HashMap<String, (f32, Duration)>`. That way we can cache for all the different currencies.

mutiny-core/src/nodemanager.rs
```diff
@@ -539,6 +539,8 @@ pub struct Plan {
     pub amount_sat: u64,
 }
 
+type CurrencyHashMap = Option<HashMap<String, (f32, Duration)>>;
```
benthecarman, line 542: You don't need to make a `type` definition here and it doesn't need to be an `Option`. `None` here is the same as an empty `HashMap` so may as well just use `HashMap`

benthecarman (approved):

PR #759, opened 2023-09-10
Use Ordering::SeqCst for manager version

benthecarman:
Not sure if this will help but it won't hurt. This _should_ give us better guarantees that the manager version will always be increased in the correct order the persists were called


benthecarman:
Only other spot we're using atomics is stop which isn't being called upon extremely often. Won't really matter

PR #761, opened 2023-09-16
Use saturating_sub in budget_remaining

benthecarman:
@benalleng found this, we can underflow after a user edits their profile budget to be less than the current budget remaining


benalleng (commented):
tACK from me, no longer able to cause an underflow

TonyGiorgio (approved):

PR #762, opened 2023-09-16
Anchors support

benthecarman:
haven't tested but this would at least let us play with anchors

we probably would want to make this optional


TonyGiorgio:
Should we change this value at all? 

https://github.com/MutinyWallet/mutiny-node/blob/d200111bca6001af291f171877ee5f8fb3b4a535/mutiny-core/src/fees.rs#L175

benthecarman:
Looks like for ldk-sample they are using 144 blocks for `Background`, we could change to that and keep `MempoolMinimum`. However it is probably better to be safe than sorry, we don't wanna cause force closures again.

PR #764, opened 2023-09-19
Better bitcoin price cache

benthecarman:
This will now save the cache to storage so we can use it on first load. Adds logic to handle if it is a first load price, to fetch it for next time.


benalleng (approved):
tACK, tested with https://github.com/MutinyWallet/mutiny-web/pull/539

TonyGiorgio (approved):

PR #765, opened 2023-09-20
Change to new storage url

benthecarman:
Change tests to run on new vss staging url


TonyGiorgio (approved):
There's a few places on the frontend you might want to change as well.

PR #767, opened 2023-09-21
Validate jwt expiry before request

benthecarman:
Needs https://github.com/MutinyWallet/auth-api/pull/9 to be deployed (is on staging)

This should help save on requests in a very slight edge case

### DIFF
--- a/mutiny-core/src/auth.rs
+++ b/mutiny-core/src/auth.rs
@@ -1,4 +1,5 @@
 #![allow(dead_code)]
+
 use crate::{
     error::MutinyError,
     lnurlauth::{make_lnurl_auth_connection, AuthManager},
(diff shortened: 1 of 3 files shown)


mutiny-core/src/auth.rs
```diff
@@ -164,8 +182,24 @@ impl MutinyAuthClient {
             }
         };
 
-        log_info!(self.logger, "Retrieved new JWT token");
-        *self.jwt.try_write()? = Some(jwt.clone());
+        let untrusted_token = UntrustedToken::new(&jwt).map_err(|_| MutinyError::LnUrlFailure)?;
```
TonyGiorgio, line 185: We already validate it once above

PR #769, opened 2023-09-21
VSS Updates

benthecarman:
Updates for having the `value` be bytes encoded instead of base64 string encoded

### DIFF
--- a/mutiny-core/src/vss.rs
+++ b/mutiny-core/src/vss.rs
@@ -37,7 +37,7 @@ impl VssKeyValueItem {
         // should we handle this unwrap better?
         let bytes = self.value.to_string().into_bytes();
 
-        let value = base64::encode(encrypt_with_key(encryption_key, &bytes));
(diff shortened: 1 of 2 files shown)


mutiny-core/src/vss.rs
```diff
@@ -109,7 +108,7 @@ impl MutinyVssClient {
     }
 
     pub async fn get_object(&self, key: &str) -> Result<VssKeyValueItem, MutinyError> {
-        let url = Url::parse(&format!("{}/getObject", self.url)).map_err(|e| {
+        let url = Url::parse(&format!("{}/v2/getObject", self.url)).map_err(|e| {
```
TonyGiorgio, line 111: I think `v2` should apply to everything now or we can code it in to the URL.

TonyGiorgio (changes requested):

TonyGiorgio (approved):

PR #771, opened 2023-09-24
Don't persist NetworkGraph

benthecarman:
We no longer read the graph from storage, we just get it from RGS on boot, so there's no reason to persist it.

We can also not save it in our memory storage because we don't need to read it.

This should help with the memory out of bounds issue


TonyGiorgio (approved):

TonyGiorgio:
Wait it's possible someone doesn't have it set to read from RGS tho right?

benthecarman:
> Wait it's possible someone doesn't have it set to read from RGS tho right?

Yeah I guess, we don't do anything for that right now

TonyGiorgio (approved):

PR #779, opened 2023-09-27
Hub Scorer

TonyGiorgio:


mutiny-core/src/scorer.rs
```diff
@@ -0,0 +1,374 @@
...
+        } else {
+            // Neither the source nor target is a preferred hub (staying "off the highway")
+            penalty = penalty
+                .checked_mul(15)
+                .unwrap_or(u64::MAX)
+                .checked_div(10)
+                .unwrap_or(u64::MAX); // 50% extra penalty
+            penalty = penalty
+                .checked_add(HUB_BASE_DISCOUNT_PENALTY_MSAT)
+                .unwrap_or(u64::MAX); // Base fee penalty
```
benthecarman, line 346: all of these `checked_X().unwrap_or(u64::MAX)` can be replaced with `saturating_X`

mutiny-core/src/lib.rs
```diff
@@ -35,6 +35,7 @@ mod onchain;
 mod peermanager;
 pub mod redshift;
 pub mod scb;
+mod scorer;
```
benthecarman, line 38: should make this `pub` might be useful to someone using as a library

benthecarman (changes requested):

benthecarman (approved):

PR #780, opened 2023-09-28
VSS migration tool

benthecarman:
Idea @futurepaul and I had, we could have a simple flow that migrates their cloud storage from one server to another.

This only would not work from transfering from a self hosted back to ours because we use an authenticated client. We could do some matching on the storage url for ours, but I think this is probably sufficient


TonyGiorgio:
That's interesting, but couldn't it just be easier to say to import their file into mutinywallet.com ?

benthecarman:
> That's interesting, but couldn't it just be easier to say to import their file into mutinywallet.com ?

this is for leaving from app.mutinywallet.com

the think is instead of just changing the string to a new url, we can have it be its own screen so the user can wait for everthing to safely transfer over

PR #784, opened 2023-10-02
Detect subscribed in backend

benthecarman:
Detects if subscribed and will save nwc profiles to vss if you have mutiny+

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -37,7 +37,7 @@ pub mod redshift;
 pub mod scb;
 pub mod scorer;
 pub mod storage;
-mod subscription;
+pub mod subscription;
 pub mod vss;
 
 #[cfg(any(test, feature = "test-utils"))]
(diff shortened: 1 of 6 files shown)


benalleng:
Will this remove the need for the frontend to store sub-status in localStorage?

benthecarman:
> Is NWC profiles being saved for everyone now or just those being subscribed?

just subscribed

TonyGiorgio:
What's going to happen to people that stopped subscribing? Is their data going to be deleted? How is it fetched when they stopped subscribing?

TonyGiorgio:
I just really don't want to see anything stored in VSS besides critical data until we have a better handle on storage stuff.

benthecarman:
> I just really don't want to see anything stored in VSS besides critical data until we have a better handle on storage stuff.

yeah i understand that, this should handle all the cases but we can leave it out for now

TonyGiorgio (approved):

PR #787, opened 2023-10-03
Improvements to startup speed

benthecarman:
With the service worker / push notification stuff I imagine we'll need to get our startup time faster. This is one step towards that.

This changes so we don't read all of storage an extra time just to get our seed for vss.

Found a couple clean ups to go along with this
- found somewhere we were making an extra cipher
- small error conversion cleanup
- fixed a potential issue where if you passed in a mnemonic while you already had a wallet created, it would just overwrite it, now we check to make sure there isn't already a wallet.

### DIFF
--- a/mutiny-wasm/src/indexed_db.rs
+++ b/mutiny-wasm/src/indexed_db.rs
@@ -1,10 +1,10 @@
 use anyhow::anyhow;
+use bip39::Mnemonic;
 use gloo_storage::{LocalStorage, Storage};
 use gloo_utils::format::JsValueSerdeExt;
 use lightning::util::logger::Logger;
 use lightning::{log_debug, log_error};
 use log::error;
-use mutiny_core::encrypt::encryption_key_from_pass;
 use mutiny_core::logging::MutinyLogger;
 use mutiny_core::nodemanager::NodeStorage;
 use mutiny_core::storage::*;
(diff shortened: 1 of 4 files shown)


mutiny-wasm/src/indexed_db.rs
```diff
@@ -61,6 +68,65 @@ impl IndexedDbStorage {
...
+        } else {
+            // if there is a mnemonic in indexed db, then decrypt it
+            let value = decrypt_value(MNEMONIC_KEY, read.into_serde()?, password)?;
+
+            let seed: Mnemonic = serde_json::from_value(value)?;
+
+            // if we hae an override mnemonic, then we need to check that it matches the one in indexed db
+            if override_mnemonic.is_some_and(|m| m != seed) {
+                panic!("Mnemonic already exists in indexed db, cannot override");
+            }
```
TonyGiorgio, line 118: Why override mnemonic?

TonyGiorgio (changes requested):
I don't see any tests here. Last time we changed encryption stuff it ended badly.

mutiny-wasm/src/indexed_db.rs
benthecarman: We currently support entering a mnemonic, we just didnt handle if a wallet was already created. This will handle it better

benthecarman:
Added a test for a different mnemonic, needed to remove the panic and just change to an error, otherwise couldnt test it properly

TonyGiorgio:
Okay that looks good. Last thought, I was revisiting some of this with this pr: https://github.com/MutinyWallet/mutiny-node/pull/752

Is that something we'd want to do? I dont know if we want to attempt to remove some of the cipher re-initializations, but this PR does fix some of the investigations I was looking through when I was attempting it again. Namely, the double storage init.

benthecarman:
Yeah I think the idea of that is good, probably good for a follow up. I think this handles it so we aren't doing anything inefficiently now but the API isn't as clean as it could be with that

TonyGiorgio (approved):
Nice work, sorry it took awhile. Needed to grok it again.

PR #793, opened 2023-10-09
surreal-db

benthecarman:
if we had to do mutiny over surreal db would be a cool storage option, tried playing around with it. doesn't fully work but in case anyone wanted to look at it


benthecarman:
was just for fun

PR #797, opened 2023-10-15
Use alternative retry method for async storage

benthecarman:
@TonyGiorgio and some other people have been reporting some issues with storage getting stuck. Not entirely sure what the issue is but this could be a fix. Changes the storage retries to be done by listing all the pending updates and persisting them in a loop rather than continually retrying on an individual update. This also lets us persist locally for retries since we are persisting the current monitor for retries, instead of the one that failed, so we don't have to worry about a race condition.


TonyGiorgio:
Cool, thanks. Will try to reproduce the issue locally and then see if this fixes it. It wasn't something I could reliably reproduce, even turning the browser network speed to 3g.

TonyGiorgio (approved):
Code looks good, logic makes sense, can't test this for awhile so if you've done so then lgtm

PR #802, opened 2023-10-22
Start on nix flakes

TonyGiorgio:


.gitignore
```diff
@@ -1,8 +1,6 @@
 .DS_Store
 debug/
 target/
-./*/Cargo.lock
-Cargo.lock
```
benthecarman, line 5: i think we still want these to be ignored, since this is a library

TonyGiorgio: It shouldn't hurt anything to have it and the rust lock file recommendations have [changed](https://www.infoworld.com/article/3705930/rust-changes-cargo-package-guidance.html#:~:text=Rust's%20Cargo%20team%20used%20to,is%20best%20for%20your%20project.&text=Rather%20than%20encouraging%20Rust%20developers%20to%20commit%20their%20Cargo.)

TonyGiorgio:
I'll rebase after the tests pass but essentially this is ready

benthecarman (approved):
for me to get it to work i have to change `aarch64-unknown-linux-gnu` to `x86_64-unknown-linux-gnu`, not sure if we can have it detect, but this is fine for now!

PR #807, opened 2023-10-25
Fix borrowed mut bug

benthecarman:
Fixes the already borrowed error that was happening on debug builds. My guess is that sometimes on startup we were grabbing the monitors at the same time as ldk so it would be borrowed twice once. The 3 second sleep at the beginning should prevent this.

Not sure why this never happened to me when testing #797, maybe something merged to master since then interfered?

image


benthecarman:
Merging this to fix bug, is very simple fix

PR #812, opened 2023-10-28
export npub to hexpub function

futurepaul:
this one function can get rid of our entire nostr dep on the frontend, which is a decent size saving: https://bundlephobia.com/package/@nostr-dev-kit/ndk@2.0.3


TonyGiorgio (approved):
LGTM. Eventually we may want a place to stick static helper functions but for now this is fine.

PR #814, opened 2023-10-30
Make singleton over MutinyWallet

benthecarman:
Should make the singleton more user friendly, instead of getting an error it'll just give you the currently instantiated one.

Problem with this is that if you tried to initiate with a different configuration you would get the same wallet back. Also have to impl send and sync which I am not totally sure is going to work.


TonyGiorgio:
Are we hitting this error? Because if we are, we should fix the root cause.

benthecarman:
> Are we hitting this error? Because if we are, we should fix the root cause.

I am not sure

PR #815, opened 2023-10-30
Nostr Wallet Auth

benthecarman:
https://github.com/nostr-protocol/nips/pull/851

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -457,8 +457,10 @@ impl<S: MutinyStorage> MutinyWallet<S> {
                         .nwc_uri
                 };
 
-                // only should have to submit the NWC if never created locally before
-                subscription_client.submit_nwc(nwc).await?;
+                if let Some(nwc) = nwc {
+                    // only should have to submit the NWC if never created locally before
+                    subscription_client.submit_nwc(nwc).await?;
+                }
             }
             Some(profile) => {
                 if profile.tag != NwcProfileTag::Subscription {
--- a/mutiny-core/src/nostr/mod.rs
+++ b/mutiny-core/src/nostr/mod.rs
@@ -1,6 +1,7 @@
 use crate::logging::MutinyLogger;
 use crate::node::Node;
 use crate::nodemanager::NodeManager;
+use crate::nostr::nip49::{NIP49BudgetPeriod, NIP49URI};
 use crate::nostr::nwc::{
     BudgetPeriod, BudgetedSpendingConditions, NostrWalletConnect, NwcProfile, NwcProfileTag,
     PendingNwcInvoice, Profile, SingleUseSpendingConditions, SpendingConditions,
@@ -15,8 +16,8 @@ use bitcoin::secp256k1::{PublicKey, Secp256k1, Signing};
 use bitcoin::util::bip32::{ChildNumber, DerivationPath, ExtendedPrivKey};
 use futures::{pin_mut, select, FutureExt};
 use futures_util::lock::Mutex;
-use lightning::log_warn;
 use lightning::util::logger::Logger;
+use lightning::{log_error, log_warn};
 use nostr::key::SecretKey;
 use nostr::nips::nip47::*;
 use nostr::prelude::{decrypt, encrypt};
@@ -27,6 +28,7 @@ use std::sync::atomic::Ordering;
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
 
+pub mod nip49;
 pub mod nwc;
 
 const NWC_ACCOUNT_INDEX: u32 = 1;
@@ -103,7 +105,7 @@ impl<S: MutinyStorage> NostrManager<S> {
             .collect()
     }
 
-    pub fn get_nwc_uri(&self, index: u32) -> Result<NostrWalletConnectURI, MutinyError> {
+    pub fn get_nwc_uri(&self, index: u32) -> Result<Option<NostrWalletConnectURI>, MutinyError> {
         let opt = self
             .nwc
             .read()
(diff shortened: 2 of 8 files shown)


benthecarman:
Have this end-to-end working with zapple pay

mutiny-core/src/lib.rs
```diff
@@ -456,6 +456,17 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+                    Some(nwc) => nwc,
+                    None => {
+                        log_error!(
+                            self.node_manager.logger,
+                            "Create NWC without valid connection string"
+                        );
+                        return Err(MutinyError::NostrError);
+                    }
+                };
+
```
TonyGiorgio, line 469: Was this happening here? This really shouldn't be a problem on `ensure_mutiny_nwc_profile`. 

In fact, if you're returning an error here, then it might cause the node not to start up since that happens on startup (or might happen on startup soon).

mutiny-core/src/nostr/mod.rs
```diff
@@ -257,6 +259,83 @@ impl<S: MutinyStorage> NostrManager<S> {
...
+                period: match budget.time_period {
+                    NIP49BudgetPeriod::Daily => BudgetPeriod::Day,
+                    NIP49BudgetPeriod::Weekly => BudgetPeriod::Week,
+                    NIP49BudgetPeriod::Monthly => BudgetPeriod::Month,
+                    NIP49BudgetPeriod::Yearly => BudgetPeriod::Year,
+                },
+            }),
+        };
+
+        let mut profiles = self.nwc.write().unwrap();
```
TonyGiorgio, line 283: unwrap shouldn't be here

mutiny-wasm/src/lib.rs
```diff
@@ -1200,6 +1201,25 @@ impl MutinyWallet {
...
+        &self,
+        name: String,
+        uri: String,
+    ) -> Result<NwcProfile, MutinyJsError> {
+        let uri = NIP49URI::from_str(&uri).map_err(|_| MutinyJsError::InvalidArgumentsError)?;
+        log::info!("Approving NWC auth request: {}", uri);
+        let profile = self
+            .inner
+            .nostr
+            .approve_nostr_wallet_auth(ProfileType::Normal { name }, uri, NwcProfileTag::General)
```
TonyGiorgio, line 1217: might as well remove type from this request if it's only going to be used for `Normal`

mutiny-core/src/nostr/mod.rs
```diff
@@ -257,6 +259,83 @@ impl<S: MutinyStorage> NostrManager<S> {
...
+                    .filter(|&nwc| nwc.profile.index >= USER_NWC_PROFILE_START_INDEX)
+                    .max_by(|a, b| a.profile.index.cmp(&b.profile.index))
+                    .map(|nwc| nwc.profile.index + 1)
+                    .unwrap_or(USER_NWC_PROFILE_START_INDEX);
+
+                debug_assert!(next_index >= USER_NWC_PROFILE_START_INDEX);
+
+                (name, next_index, Some(get_random_bip32_child_index()))
+            }
+        };
```
TonyGiorgio, line 303: Is this copy-pasted from somewhere? Might as well make it a function.

mutiny-core/src/nostr/mod.rs
```diff
@@ -368,6 +448,68 @@ impl<S: MutinyStorage> NostrManager<S> {
...
+        tag: NwcProfileTag,
+    ) -> Result<NwcProfile, MutinyError> {
+        let mut commands = uri.required_commands.clone();
+        commands.extend_from_slice(&uri.optional_commands);
+
+        let secret = uri.secret.clone();
+        let relay = uri.relay_url.to_string();
+        let profile = self.nostr_wallet_auth(profile_type, uri, tag)?;
+
+        let nwc = self.nwc.read().unwrap().iter().find_map(|nwc| {
```
TonyGiorgio, line 467: No unwrap.

mutiny-core/src/nostr/nwc.rs
```diff
@@ -659,13 +717,11 @@ impl NostrWalletConnect {
...
             index: self.profile.index,
+            client_key: self.profile.client_key,
             relay: self.profile.relay.clone(),
             enabled: self.profile.enabled,
             archived: self.profile.archived,
-            nwc_uri: self
-                .get_nwc_uri()
-                .expect("failed to get nwc uri")
-                .to_string(),
+            nwc_uri: self.get_nwc_uri().unwrap().map(|uri| uri.to_string()),
```
TonyGiorgio, line 724: No unwrap

mutiny-wasm/src/lib.rs
```diff
@@ -1273,7 +1293,8 @@ impl MutinyWallet {
     #[wasm_bindgen]
     pub fn get_nwc_uri(&self, index: u32) -> Result<String, MutinyJsError> {
         match self.inner.nostr.get_nwc_uri(index) {
-            Ok(uri) => Ok(uri.to_string()),
+            Ok(Some(uri)) => Ok(uri.to_string()),
+            Ok(None) => Ok("NWA".to_string()),
```
TonyGiorgio, line 1297: We're going to need some changes on the frontend side for this, right? Wouldn't it be better to just leave string as empty?

TonyGiorgio (changes requested):
Could use some tests too

mutiny-wasm/src/lib.rs
benthecarman: We may want to use `approve_nostr_wallet_auth` with other profile types in the future, think it is better to have the type.

benthecarman:
Responded to review and added a test (which caught a bug :) )

TonyGiorgio (approved):

PR #820, opened 2023-11-01
Receive payjoin

DanGould:
Request + receive payjoin v2. This basic functionality requires no changes to mutiny-web because it only changes the presented bip21 uri to include a `&pj=` endpoint. This change does NOT yet include v2 sending, but was tested using the payjoin-cli in the [payjoin v2 pr]([the payjoin v2 implementation](https://github.com/DanGould/rust-payjoin/tree/serverless-payjoin)

In order to become a production feature some changes must be considered:

## Fetch Oblivious HTTP Configuration

This implementation hard codes the payjoin relay's ohttp gateway's configuration (ciphersuite and keys). This information ought to be fetched by mutiny and served over https. If a mutiny client asks for this information directly without proxy then their IP address would be revealed to the payjoin relay, which OHTTP otherwise protects.  It could also be configured by the Mutiny client but that adds complexity and assumes the client takes steps to protect their IP address in an independent fetch.

## Display `&ohttp=` gateway configuration in bip21 uri

By including the ohttp gateway configuration in the bip21 uri, the sender can immediately connect to the payjoin server without fetching the configuration and potentially leaking their IP. This parameter should be included in waila.

It could also be hacked into the pj endpoint as a query parameter, but is information for the sender wallet, not to be passed as query to payjoin relay, and is better suited to a bip21 parameter.

## Run Oblivious HTTP Relay

Mutiny should run a basic HTTP proxy to relay OHTTP requests from client to payjoin relay. If this relay were run by the same party as the payjoin relay, then that party would have enough information to correlate specific payjoin peer pairs with IP addresses. This same server could be trusted to fetch ohttp configurations on behalf of mutiny clients. The worst Mutiny could do by failing to return proper ohttp configurations would be deny service, NOT disrupt privacy (Assuming the payjoin relay is available over https as a known entity run separately from the Mutiny OHTTP relay).

## Link a proper `payjoin` crate release

The [payjoin dependency tested here](https://github.com/DanGould/rust-payjoin/tree/serverless-payjoin-mutiny) varies slightly from that in [the payjoin v2 implementation](https://github.com/DanGould/rust-payjoin/tree/serverless-payjoin). That payjoin dependency even depends on a WIP `ohttp` crate. These all must be published on crates.io for a production release.


mutiny-core/src/nodemanager.rs
```diff
@@ -1052,14 +1055,115 @@ impl<S: MutinyStorage> NodeManager<S> {
             return Err(MutinyError::WalletOperationFailed);
         };
 
+        // If we are in safe mode, we don't create payjoin sessions
```
TonyGiorgio, line 1058: I think we can do this anyways in safe mode given that it's mostly lightning related for that.

mutiny-core/src/nodemanager.rs
```diff
@@ -1052,14 +1055,115 @@ impl<S: MutinyStorage> NodeManager<S> {
...
+                "AQAg7YjKSn1zBziW3LvPCQ8X18hH0dU67G-vOcMHu0-m81AABAABAAM",
+                "http://localhost:8080", // Ignore and pass directly to relay for now
+                                         //Some("c53989e590b0f02edeec42a9c43fd1e4e960aec243bb1e6064324bd2c08ec498")
+            );
+            let pj_uri = format!("{}{}", "http://localhost:8080/", context.subdirectory());
+            log_debug!(self.logger, "{pj_uri}");
+            let wallet = self.wallet.clone();
+            // run await payjoin task in the background as it'll keep polling the relay
+            wasm_bindgen_futures::spawn_local(async move {
+                let wallet = wallet.clone();
```
TonyGiorgio, line 1073: We'll probably want some way to stop this. A stop channel should be available to use but might need to also stop it if one of the payment methods in the bip21 completes. Though we haven't linked them together yet so that might be hard to do currently.

mutiny-core/src/nodemanager.rs
```diff
@@ -1052,14 +1055,115 @@ impl<S: MutinyStorage> NodeManager<S> {
...
+            });
+            Some(pj_uri)
+        };
+
         Ok(MutinyBip21RawMaterials {
             address,
             invoice,
             btc_amount: amount.map(|amount| bitcoin::Amount::from_sat(amount).to_btc().to_string()),
             labels,
+            pj,
```
TonyGiorgio, line 1085: I wonder if we should limit the options or allow user to specify a specific one. My concern is mostly showing both on chain address + lightning + payjoin in each QR code might make it pretty dense. Do you have an example payjoin uri with the new fields added for `ohttp` to get an idea of the size?

mutiny-core/src/nodemanager.rs
```diff
@@ -1052,14 +1055,115 @@ impl<S: MutinyStorage> NodeManager<S> {
...
+                .map_err(|e| anyhow!("parse error {}", e))
+                .unwrap();
+            match proposal {
+                Some(proposal) => return Ok(proposal),
+                None => Self::delay(5000).await.unwrap(),
+            }
+        }
+    }
+
+    async fn delay(millis: u32) -> Result<(), wasm_bindgen::JsValue> {
```
TonyGiorgio, line 1156: We should have this already in the `utils`

TonyGiorgio (commented):
Very cool, appreciate the description and PR! 

I'm curious what ohttp adoption is looking like and what information is gained by running one? I guess the assumption would be that the ohttp proxy and the service are not the same people and not colluding? What, if any, information gained by the payjoin server or ohttp proxy? Trying to think through what it makes sense to host or have others host that we link to in app. 

Would users be able to select their own ohttp proxy and their own payjoin servers? Ideally we don't have to host or hard code it in order for it to work, though potentially willing to if it makes sense.

mutiny-core/src/nodemanager.rs
DanGould: It's an base64-url encoded ~128 bit pubkey + a few bytes to specify ciphersuite from a pre-defined list. We could (should?) encode &pj= and &ohttp= using the bitcoincommons UR encoding + Caps in the address, params, & BOLT 11 for a significantly more compact QR.

`bitcoin:tb1pr3hpfv99fhftn5gdxucgy4rax5mqggu2xc96puxgw5u4wp93tnrqgpcpmx?amount=0.00066588&lightning=lntbs665880n1pj5zdq4sp5n594wqz7cxfyxm8ngm6pjwt5x8ne9mjfjpejs8f9aq2jk42gdafqpp5jfjssftnn2un8e7wv94gp4e6zg65vpcycqcra3a6ccqgeh2myaxqdqqnp4qdn2hj8tfknpuvdg6tz9yrf3e27ltrx9y58c24jh89lnm43yjwfc5xqrrsxcqzzn9qyysgqumnvl5s6xg9hpcl4mvf8qfl4l6axyqj9vqaqrgflc2al690yxr5kpuyut6dfckpvts8fgx3uhqhat5zezwdzxgvjytmavkxsxe2cxkcqd88wvj&pj=http://localhost:8080/A_-kmO0WUPULpAWTCqt5J4IlbdynoTY0NsxkRBtKupoR&ohttp=AQAgwMcjpR7rer1o726dZj4n4-gqCdIVN7Q2bA6gZYwK2CwABAABAAM`

DanGould:
> I'm curious what ohttp adoption is looking like 

[Oblivious HTTP](https://ietf-wg-ohai.github.io/oblivious-http/draft-ietf-ohai-ohttp.html#name-message-ohttp-req-media-typ) is still a *draft*, but is implemented in Chrome, Firefox, Cloudflare and Apple operating systems. It is used for [Oblivious DNS over HTTP](https://blog.cloudflare.com/oblivious-dns/) with a relay deployed by Fastly.

> and what information is gained by running one? I guess the assumption would be that the ohttp proxy and the service are not the same people and not colluding? What, if any, information gained by the payjoin server or ohttp proxy? Trying to think through what it makes sense to host or have others host that we link to in app.

OHTTP relies on 2 distinct servers. The OHTTP Relay Resource sees the client IP and the OHTTP Gateway Resource target IP, behind which the target resource may be accessed. The OHTTP Relay sees an encrypted `ohttp-req` message. It does NOT see the contents of the message.

The OHTTP Gateway sees that the request came from the OHTTP Relay IP, and may decrypt the contents of the request: a second layer of e2ee between sender and receiver and the pubkey associated with that session (a pseudonymous subdirectory identifier for the session)

The threat model in the BIP assumes they do not collude. If they do collude, they may correlate the subdirectory with sender and receiver IP addresses to discover only that 2 IP addresses may have communicated with one another at the Payjoin relay subdirectory at a given time.

> Would users be able to select their own ohttp proxy and their own payjoin servers? Ideally we don't have to host or hard code it in order for it to work, though potentially willing to if it makes sense.

I would suggest users don't select their own proxy for the initial rollout for three reasons.

1. Dependency compatibility: OHTTP's specified Hybrid Pub Key Encryption HMAC-based Extract-and-Expand Key Derivation Function does not support the secp256k1 curve. I've specified a change which uses secp256k1 so that bitcoin software does not need any external crypto in order to use payjoin v2. Secp256k1, sha256, and chacha20poly1305 are all already part of the bitcoin stack and secure for the purposes of such a protocol. There is a formal process to get secp256k1 into the spec Key Encapsulation Mechanisms [via IANA](https://www.iana.org/assignments/hpke/hpke.xhtml) and potentially the browsers/operating systems. However, the spec does not prevent us from building working software today.
2. UX: Any external relay would need to support secp256k1 or potentially be incompatible with a payjoin peer. Any *draft* spec external  OHTTP Relay will not support secp256k1 at time of writing. A Payjoin Relay relied on an OHTTP Gateway configuration using only curves other than secp256k1, which is guaranteed to be available in bitcoin software, some payjoin-enabled wallets may not be able to communicate with them.
3. Privacy: Mutiny clients already reveal their IP address to Mutiny and don't need to reveal it to another service. There is also an anonymity set benefit of having as many users using the same OHTTP relay as possible.

TonyGiorgio:
By running an OHTTP relay, could anyone use that to connect to any other OHTTP gateway? I'm guessing the relay would see what IP is connecting to what gateway, but could there be multiple services behind any given gateway?

mutiny-core/src/nodemanager.rs
TonyGiorgio: https://github.com/MutinyWallet/mutiny-node/blob/e11cdc341e935b8dd1ae7a59f3449f2577184004/mutiny-core/src/nodemanager.rs#L969-L975

DanGould:
> By running an OHTTP relay, could anyone use that to connect to any other OHTTP gateway?

In section 8.2: "The relay will only forward for Oblivious Gateway Resources that it has explicitly configured and allowed."

> I'm guessing the relay would see what IP is connecting to what gateway, but could there be multiple services behind any given gateway?

Yes, and specific path, query, methods, request body, headers are all encrypted from the ohttp relay's view

TonyGiorgio:
Okay thank you, this is very helpful. Just trying to understand the operational side of OHTTP and whether we should stand one up. Regardless of who runs what, I like the approach and the code is clean and makes sense. Appreciate the work there!

DanGould:
FYI standing up a relay is as simple as adding a couple blocks to nginx if you're already running it somewhere

```conf
http {
    # HTTP Server Block
    server {
        listen 80;
        server_name your-server-name.com;

        location /ohttp-relay {
            # Forward requests to the OHTTP Gateway
            proxy_pass https://ohttp-gateway.example.com;

            # Clear the X-Real-IP header
            proxy_set_header X-Real-IP "";
            # Make default proxy_pass behavior explicit for good measure

            access_log /path/to/http-access.log;
            error_log /path/to/http-error.log;
        }
    }

    # HTTPS Server Block
    server {
        listen 443 ssl;
        server_name your-server-name.com;

        ssl_certificate /path/to/your/certificate.crt;
        ssl_certificate_key /path/to/your/private.key;

        location /ohttp-relay {
            # Forward requests to the OHTTP Gateway
            proxy_pass https://ohttp-gateway.example.com;

            # Clear the X-Real-IP header
            proxy_set_header X-Real-IP "";
            # Make default proxy_pass behavior explicit for good measure

            access_log /path/to/https-access.log;
            error_log /path/to/https-error.log;
        }
    }
}
```

mutiny-core/src/nodemanager.rs
DanGould: We should also be able to start this from persistent storage if the wallet goes offline and wants to check once it's back open online

DanGould:
This should now be receiving v2 payjoins, however the persistence and asynchronous completion is still a bit janky in this form factor. It does not seem like `resume_payjoins` is properly resuming sessions on startup.

We've also got to consider how to display pending payjoin in activity. It seems like that UI list is only considering broadcasted transactions and pending lightning payments at the moment. This introduces a new tx pending state where a PSBT has been signed and posted to a payjoin directory but not yet responded to (from a sender) or responded to but not yet broadcast by the sender (for a receiver).

I think v2 payjoin receivers who sign should show transactions pending, since only 1 step remains before the transaction is broadcast, and they've effectively been probed and revealed their utxo. v2 Senders perhaps should not treat a single signature as pending, since they both need to have receiver sign and check the result, at least until BDK allows us to lock UTXOs. Still, a sender in this state could have their original psbt broadcast by a receiver. I'll have to look into how you're handling pending lightning payments to become inspired with a solution regarding pending payjoins.

TonyGiorgio:
I think it's valid to consider it as pending in the UI. I think we just do a filter on `ActivityItems` before returning it. We may have to save these in our DB for it to reflect state accurately.

DanGould:
This is now based on the `payjoin-0.13.0` release, persists receive payjoin sessions, and adds pending transactions to the `wallet` using `insert_tx` and `cancel_tx` before they hit mempool. Though tracked payjoins are not explicitly marked as such. Their PSBTs are just put in the `ActivityItem::Onchain` list.  One way to explicitly mark them as payjoin would be to manually construct payjoin `TransactionDetails`, add an `is_payjoin` field or similar, store them in `MutinyStorage`, and pull from that list in `get_activity()` if this solution seems incomplete.  This solution should work with the existing `add_onchain_labels` since it attaches labels to addresses.

For now, the way the receiver here gets payjoin directory's `ohttp-keys` does reveal IP address to the directory but that will be solved with the next deployment of [ohttp-relay](https://github.com/payjoin/ohttp-relay) @teemie1

This would require the addition of `&ohttp=` bip21 param parsing, ideally in bitcoin-waila, for ui support.

DanGould:
While I've left a few TODOs that are in progress, this change is ready for deeper review

mutiny-core/src/lib.rs
```diff
@@ -172,6 +175,8 @@ pub enum ActivityItem {
     OnChain(TransactionDetails),
     Lightning(Box<MutinyInvoice>),
     ChannelClosed(ChannelClosure),
+    // /// A payjoin proposal is posted to the directory but not yet broadcast from the sender
+    // PendingPayjoin,
```
DanGould, line 179: @benalleng Currently v2 payjoins here are, as you mentioned, handles as on-chain ActivityItems.

This only gets triggered for senders who have signed an original psbt before their payjoin psbt is signed and broadcast by a receiver. We should explore whether or not this is sufficient, what is if not, and document a decision on what to do for the receiver who as of now only will see an ActivityItem once they post a received payjoin psbt that they broadcast

TonyGiorgio:
Sorry for the delay on this. 

I want to be transparent, that we're pretty unsure about what we want to do with the on chain wallet long term. We're really not optimizing for on chain related activities and it has just caused various friction. Ideally long term we get to move towards something like splicing and can do splice outs for on chain payments, but until then, we've talked about removing the on chain wallet and defaulting to on chain going into a fedimint. 

I've seen the discussions [here](https://github.com/fedimint/fedimint/discussions/1764) which are great to see but would imagine it might be complex.

PR #821, opened 2023-11-01
Update wasm bindgen

benthecarman:
New wasm-bindgen supports sending `Vec<T>` for non-async functions. Tried to implement that everywhere we can.


TonyGiorgio:
How does this look on the typescript side? Would it be straightfoward enough to implement? @futurepaul ?

futurepaul:
conceptually it sounds like a win, although in `mutiny_wasm.d.ts` I still see these show up as `any` type (rather than `string[]` which is what I would hope for)

but when I switch to this branch I get some panics... 

Screenshot 2023-11-02 at 12 48 41 PM

Screenshot 2023-11-02 at 12 49 30 PM

TonyGiorgio:
We're not running lightning 117 anymore. Are you correctly updating?

futurepaul (approved):
re-tested this and it's actually very helpful (more types in the frontend!). idk what I was doing wrong before. here's a frontend PR against this: https://github.com/MutinyWallet/mutiny-web/pull/691

TonyGiorgio (approved):

PR #822, opened 2023-11-02
Add session id to logs

benthecarman:
This should hopefully help detect if we are getting multiple instances running on the same device. By giving a session id to a running wallet that will show up in the logs and we could see 2 concurrent versions running


TonyGiorgio (approved):
Good idea

PR #827, opened 2023-11-05
Give InsufficientBalance before trying to send

### LINKED ISSUE #816: add error for when insufficient funds due to channel not being fully open yet

benthecarman:
Closes #816

Before if we had a confirming channel we would still try to send and give a RouteNotFound error which could be confusing.

This fixes by first summing up the balance of our confirmed channels and seeing if we have enough balance. 

We can remove the `is_empty` check because that will sum to `0`

### DIFF
--- a/mutiny-core/src/node.rs
+++ b/mutiny-core/src/node.rs
@@ -1083,9 +1083,24 @@ impl<S: MutinyStorage> Node<S> {
             return Err(MutinyError::NonUniquePaymentHash);
         }
 
-        if self.channel_manager.list_channels().is_empty() {
-            // No channels so routing will always fail
-            return Err(MutinyError::RoutingFailed);
+        // get invoice amount or use amt_sats
+        let send_msats = invoice
+            .amount_milli_satoshis()
+            .or(amt_sats.map(|x| x * 1_000))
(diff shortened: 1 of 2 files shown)


mutiny-core/src/node.rs
```diff
@@ -1083,9 +1083,24 @@ impl<S: MutinyStorage> Node<S> {
...
+        if channels
+            .iter()
+            // only consider channels that are confirmed
+            .filter(|c| c.is_channel_ready)
+            .map(|c| c.balance_msat)
+            .sum::<u64>()
+            < send_msats
+        {
+            // Channels exist but not enough capacity
+            return Err(MutinyError::InsufficientBalance);
```
TonyGiorgio, line 1103: This could be confusing if the user has some channels and balance but the connection to the LSP node is not currently active.

benthecarman: This handles that, is channel ready, only checks if it's been confirmed, not if they have an active connection.

TonyGiorgio: Ohh okay. Then it sounds like it is much better then but doesn't quite resolve the LSP disconnection problem, but I think that's what the UI designs for disconnections will have to fix.

TonyGiorgio (approved):

PR #831, opened 2023-11-07
Allow for 0 reserve in channels

benthecarman:
I dont think this will let the user have 0 reserve but it will at least give them slightly more inbound from the LSP not having any reserve


TonyGiorgio:
Is 0 an allowed value? I think above dust is still required.

benthecarman:
it says

> If the calculated proportional value is less than 1000 sats, it will be treated as 1000 sats instead, which is a safe implementation-specific lower bound

I can change to 1000 sats which is actually the min

TonyGiorgio (approved):

PR #839, opened 2023-11-09
Restore returning the subscription expire time instead of bool

benthecarman:
small fix so we can display when the next mutiny+ invoice is

### DIFF
--- a/mutiny-wasm/src/lib.rs
+++ b/mutiny-wasm/src/lib.rs
@@ -65,6 +65,7 @@ async fn uninit() {
 pub struct MutinyWallet {
     mnemonic: Mnemonic,
     inner: mutiny_core::MutinyWallet<IndexedDbStorage>,
+    subscription_expire_time: Option<u64>,
 }
 
 /// The [MutinyWallet] is the main entry point for interacting with the Mutiny Wallet.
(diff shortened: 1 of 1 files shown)


TonyGiorgio (approved):

mutiny-wasm/src/lib.rs
```diff
@@ -262,7 +261,11 @@ impl MutinyWallet {
         let inner =
             mutiny_core::MutinyWallet::new(storage, config, Some(logger.session_id.clone()))
                 .await?;
-        Ok(MutinyWallet { mnemonic, inner })
+        Ok(MutinyWallet {
+            mnemonic,
+            inner,
+            subscription_expire_time: sub_expired_time,
```
TonyGiorgio, line 267: really? that's such a hack

benthecarman: not sure a better place to put it, maybe the config?

TonyGiorgio: Honestly I was fine with the old way of doing it. Now I'm questioning if you've covered every edge case paul was doing in the frontend previously.

TonyGiorgio: Where do you update this when the user subscribes?

TonyGiorgio: Honestly I would rather revert the subscription changes at this point.

PR #840, opened 2023-11-09
Revert "Detect subscribed in backend"

benthecarman:
Reverts MutinyWallet/mutiny-node#784


TonyGiorgio (approved):
Tested and it works just fine

PR #842, opened 2023-11-10
Windowless sleep 2.0

benthecarman:
Need @futurepaul to test but I _think_ this version will work

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1145,6 +1145,16 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


futurepaul:
This has the same web sys call under the hood which uses window: https://github.com/tomaka/wasm-timer/blob/64c812a69981cbf0bc428d290a7887818f451e2d/src/timer/global/wasm.rs#L27

benthecarman:
Dang it

PR #844, opened 2023-11-10
Properly return IncorrectPassword when failing to decrypt seed

benthecarman:


TonyGiorgio (approved):
Don't have time to test it but it looks like that'll fix this.

PR #846, opened 2023-11-11
release vss lock when stop is called

futurepaul:
was hoping this would fix flaky test. it didn't I don't think, but it's still an improvement (especially when restoring in place)


benthecarman (commented):
Needs `rustfmt`

Logically, this makes sense to me. I fear about merging right now as we're trying to nail down double init errors. This would just add to the confusion.

PR #848, opened 2023-11-11
Use full inbound capacity for LSP check

benthecarman:
Before we were only checking if we had a single channel big enough, this fixes so we use all of our channels inbound.


TonyGiorgio (approved):
Nice catch

PR #849, opened 2023-11-11
Log device lock on startup

benthecarman:
might help finding double init errors


TonyGiorgio (approved):
I suspect one problem with finding these are overwriting logs, but doesn't hurt here.

Issue #826, opened 2023-11-04
Ease the hodl / stuck payment problem


TonyGiorgio:
Background processes, push notifications, longer cltv, etc. Need to do something more to help mitigate force closes due to this reason.

TonyGiorgio:
Handled a lot with this

PR #854, opened 2023-11-14
Block zeus pay node for nwc

benthecarman:
If a user wants to pay to a zeus pay address explicitly we should allow them but often over nwc they user is just randomly sending payments, not knowing where they are going and this can cause force closures. So lets just stop payments to zeus pay over nwc for now.

Helps with #826

### DIFF
--- a/mutiny-core/src/nostr/nwc.rs
+++ b/mutiny-core/src/nostr/nwc.rs
@@ -6,6 +6,7 @@ use crate::storage::MutinyStorage;
 use crate::utils;
 use anyhow::anyhow;
 use bitcoin::hashes::hex::{FromHex, ToHex};
+use bitcoin::secp256k1::PublicKey;
 use bitcoin::secp256k1::{Secp256k1, Signing};
 use bitcoin::util::bip32::ExtendedPrivKey;
 use chrono::{DateTime, Datelike, Duration, NaiveDateTime, Utc};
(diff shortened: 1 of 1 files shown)


benthecarman:
> Are you sure it goes to this node and not an intermediary zaplocker one?

this is the node id I got from the invoice when getting from `https://zeuspay.com/api/lnurl/pay/benthecarman?amount=1000000`

TonyGiorgio (approved):

supertestnet:
I think this is wise. Perhaps, in the future, mobile users will be able to safely pay hodl invoices. But right now they cannot.

Be aware that there is an easy way to generalize this solution to all zaplocker nodes. (As far as I'm aware, only myself and Evan are currently running zaplocker, but I hope more people join us.) Consider this link: https://zeuspay.com/api/lnurl/pay/benthecarman?amount=1000000

The invoice you get back has a key/value pair that says "hodl_invoice: true"

All (i.e. both) zaplocker nodes show this key/value pair. So consider not paying invoices if retrieved from an endpoint that shares that key/value pair.

TonyGiorgio:
> I think this is wise. Perhaps, in the future, mobile users will be able to safely pay hodl invoices. But right now they cannot.
> 
> Be aware that there is an easy way to generalize this solution to all zaplocker nodes. (As far as I'm aware, only myself and Evan are currently running zaplocker, but I hope more people join us.) Consider this link: https://zeuspay.com/api/lnurl/pay/benthecarman?amount=1000000
> 
> The json it serves has a key/value pair that says "hodl_invoice: true"
> 
> All (i.e. both) zaplocker nodes show this key/value pair. So consider not paying invoices if retrieved from an endpoint that shares that key/value pair.

This is not an option because we are not pulling the LNURL directly, the invoice is coming from nostr clients.

PR #856, opened 2023-11-15
LDK in std mode

TonyGiorgio:
Binary size is 9.7MB, I think it went up a little bit but not much. 

Tested almost all functionality before and after the upgrade, no issues.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,23 +53,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ahash"
-version = "0.3.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e8fd72866655d1904d6b0997d0b07ba561047d070fbe29de039031c641b61217"
-
-[[package]]
-name = "ahash"
-version = "0.7.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5a824f2aa7e75a0c98c5a504fceb80649e9c35265d44525b5f94de4771a395cd"
-dependencies = [
- "getrandom",
- "once_cell",
- "version_check",
-]
-
-[[package]]
 name = "aho-corasick"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


benthecarman:
I think with this we can start using `lightning-background-sync`, might be worth doing in a follow up though

TonyGiorgio:
> I think with this we can start using `lightning-background-sync`, might be worth doing in a follow up though

I think the copy/paste was originally for `sync + send` stuff we had to remove.

benthecarman:
> I think the copy/paste was originally for sync + send stuff we had to remove.

Yeah IIRC thats because the stuff we were using didn't have sync + send in no-std

benthecarman:
We also might be able to get rid of the mutex in `utils`

TonyGiorgio:
Getting a lot of errors switching to their background processor: 

```
585  | pub async fn process_events_async<                                                                                                                                                                                                            
     |              -------------------- required by a bound in this function                                                                                                                                                                        
...                                                                                                                                                                                                                                                  
603  |     CM: 'static + Deref<Target = ChannelManager<CW, T, ES, NS, SP, F, R, L>> + Send + Sync,                                                                                                                                                   
     |                                                                                ^^^^ required by this bound in `process_events_async`                                                                                                          
help: consider further restricting this bound                                                                                                                                                                                                        
     |                                                                                                                                                                                                                                               
177  | impl<S: MutinyStorage + std::marker::Sync> Node<S> {                                                                                                                                                                                          
     |                       +++++++++++++++++++                                                                                                                                                                                                     
                                                                                                                                                                                                                                                     
error[E0277]: `S` cannot be sent between threads safely                                                                                                                                                                                              
    --> mutiny-core/src/node.rs:493:21                                                                                                                                                                                                               
     |                                                                                                                                                                                                                                               
489  |                 if let Err(e) = process_events_async(                                                                                                                                                                                         
     |                                 -------------------- required by a bound introduced by this call                                                                                                                                              
...                                                                                                                                                                                                                                                  
493  |                     background_processor_channel_manager.clone(),                                                                                                                                                                             
     |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `S` cannot be sent between threads safely                                                                                                                                    
     |                                                                                                                            
```

Which comes from our own trait: 
```
pub trait MutinyStorage: Clone + Sized + 'static {
```

and I suppose we have probably done quite a bit of that everywhere. Was that due to the Mutex type we were using or something? I thought really comes from the fact that we're limited to single threaded in wasm anyways.

benthecarman:
Okay let's just ignore it for now then, can attempt it later.

benthecarman (approved):

PR #864, opened 2023-11-17
Unfilter force cloes balance

### LINKED ISSUE #862: Unconfirmed Force closes do not show up in the balance

benthecarman:
Closes #862

This was a bug from #825 turns out this counts towards both mutual closes and force closes. https://github.com/lightningdevkit/rust-lightning/issues/2738 will allow us to truly fix.

Force closes are more common (and scary) in Mutiny so for now would be better to have the bug be around mutual closes.

Don't think we can write a test for this without mocking `chain_monitor` and `channel_manager`

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -37,7 +37,6 @@ use bitcoin::{Address, Network, OutPoint, Transaction, Txid};
 use core::time::Duration;
 use esplora_client::Builder;
 use futures::{future::join_all, lock::Mutex};
-use lightning::chain::channelmonitor::Balance;
 use lightning::chain::Confirm;
 use lightning::events::ClosureReason;
 use lightning::ln::channelmanager::{ChannelDetails, PhantomRouteHints};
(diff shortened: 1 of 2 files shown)


TonyGiorgio (approved):
True, it's more rare when a normal close happens anyways for the time being. And when force closes can result in potential on chain transactions waiting to confirm at previous fee rates, it is necessary to show _something_ rather than nothing.

though are transaction activity items not showing up either then?

benthecarman:
> though are transaction activity items not showing up either then?

those still show up

mutiny-core/src/nodemanager.rs
```diff
@@ -37,7 +37,6 @@ use bitcoin::{Address, Network, OutPoint, Transaction, Txid};
 use core::time::Duration;
 use esplora_client::Builder;
 use futures::{future::join_all, lock::Mutex};
-use lightning::chain::channelmonitor::Balance;
```
TonyGiorgio, line 40: wrong commit appended to :P

benthecarman: dammit

PR #865, opened 2023-11-17
Return proper NWC errors on skipped invoices

### LINKED ISSUE #863: Send proper NWC response for skipped hodl invoices

benthecarman:
Now we will return a proper error response to the relay when we get an invoice we won't pay over nwc. This way the relay can safely delete them and we don't have to replay them everytime. 

Closes #863 also fixes for expired invoices and invoices with no amount.

### DIFF
--- a/mutiny-core/src/nostr/nwc.rs
+++ b/mutiny-core/src/nostr/nwc.rs
@@ -323,6 +323,28 @@ impl NostrWalletConnect {
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


benthecarman:
damn need to update tests

TonyGiorgio (approved):
Could use a test for the hodl invoice skip, since you have all the others. Otherwise lgtm

PR #867, opened 2023-11-19
Add fedimint

TonyGiorgio:
TODO: 
- [x] payment storage in gluesql
- [x] Get labels to work properly with fedimint payments
- [x] NWC Invoice Payer

Follow up: 
- [ ] LNURL out of fedimint

### DIFF
--- a/mutiny-core/Cargo.toml
+++ b/mutiny-core/Cargo.toml
@@ -46,6 +46,7 @@ jwt-compact = { version = "0.8.0-beta.1", features = ["es256k"] }
 argon2 = { version = "0.5.0", features = ["password-hash", "alloc"] }
 payjoin = { version = "0.10.0", features = ["send", "base64"] }
 gluesql = { version = "0.15", default-features = false, features = ["memory-storage"] }
+gluesql-core = "0.15.0"
 bincode = "1.3.3"
 hex = "0.4.3"
 
--- a/mutiny-core/Cargo.toml
+++ b/mutiny-core/Cargo.toml
@@ -46,6 +46,14 @@ jwt-compact = { version = "0.8.0-beta.1", features = ["es256k"] }
 argon2 = { version = "0.5.0", features = ["password-hash", "alloc"] }
 payjoin = { version = "0.10.0", features = ["send", "base64"] }
 
+fedimint-client = { git = "https://github.com/fedimint/fedimint.git", rev = "51d09c76b8a15615114d4f2c2663c738d8ff3135" }
+fedimint-core = { git = "https://github.com/fedimint/fedimint.git", rev = "51d09c76b8a15615114d4f2c2663c738d8ff3135" }
+fedimint-wallet-client = { git = "https://github.com/fedimint/fedimint.git", rev = "51d09c76b8a15615114d4f2c2663c738d8ff3135" }
+fedimint-mint-client = { git = "https://github.com/fedimint/fedimint.git", rev = "51d09c76b8a15615114d4f2c2663c738d8ff3135" }
+fedimint-ln-client = { git = "https://github.com/fedimint/fedimint.git", rev = "51d09c76b8a15615114d4f2c2663c738d8ff3135" }
+fedimint-bip39 = { git = "https://github.com/fedimint/fedimint.git", rev = "51d09c76b8a15615114d4f2c2663c738d8ff3135" }
+fedimint-ln-common = { git = "https://github.com/fedimint/fedimint.git", rev = "51d09c76b8a15615114d4f2c2663c738d8ff3135" }
+
 base64 = "0.13.0"
 pbkdf2 = "0.11"
 aes-gcm = "0.10.1"
--- a/mutiny-core/Cargo.toml
+++ b/mutiny-core/Cargo.toml
@@ -45,6 +45,9 @@ aes = { version = "0.8" }
 jwt-compact = { version = "0.8.0-beta.1", features = ["es256k"] }
 argon2 = { version = "0.5.0", features = ["password-hash", "alloc"] }
 payjoin = { version = "0.10.0", features = ["send", "base64"] }
+gluesql = { version = "0.15", default-features = false, features = ["memory-storage"] }
+bincode = "1.3.3"
+hex = "0.4.3"
 
 fedimint-client = { git = "https://github.com/fedimint/fedimint.git", rev = "51d09c76b8a15615114d4f2c2663c738d8ff3135" }
 fedimint-core = { git = "https://github.com/fedimint/fedimint.git", rev = "51d09c76b8a15615114d4f2c2663c738d8ff3135" }
@@ -81,6 +84,7 @@ js-sys = "0.3.65"
 gloo-net = { version = "0.4.0" }
 instant = { version = "0.1", features = ["wasm-bindgen"] }
 getrandom = { version = "0.2", features = ["js"] }
+gluesql = { version = "0.15", default-features = false, features = ["idb-storage"] }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokio = { version = "1", features = ["rt"] }
--- a/mutiny-core/src/federation.rs
+++ b/mutiny-core/src/federation.rs
@@ -116,6 +116,10 @@ pub struct FederationStorage {
 pub struct FederationIdentity {
     pub uuid: String,
     pub federation_id: FederationId,
+    // https://github.com/fedimint/fedimint/tree/master/docs/meta_fields
+    pub federation_name: Option<String>,
+    pub federation_expiry_timestamp: Option<String>,
+    pub welcome_message: Option<String>,
 }
 
 // This is the FederationIndex reference that is saved to the DB
@@ -356,10 +360,15 @@ impl<S: MutinyStorage> FederationClient<S> {
         }
     }
 
-    pub async fn get_mutiny_federation_identity(&self) -> FederationIdentity {
+    pub fn get_mutiny_federation_identity(&self) -> FederationIdentity {
         FederationIdentity {
             uuid: self.uuid.clone(),
             federation_id: self.fedimint_client.federation_id(),
+            federation_name: self.fedimint_client.get_meta("federation_name"),
+            federation_expiry_timestamp: self
+                .fedimint_client
+                .get_meta("federation_expiry_timestamp"),
+            welcome_message: self.fedimint_client.get_meta("welcome_message"),
         }
     }
 }
--- a/mutiny-core/src/federation.rs
+++ b/mutiny-core/src/federation.rs
@@ -4,9 +4,9 @@ use crate::{
     logging::MutinyLogger,
     nodemanager::MutinyInvoice,
     onchain::coin_type_from_network,
-    sql::glue::GlueDB,
+    sql::{glue::GlueDB, ApplicationStore},
     storage::MutinyStorage,
-    utils::sleep,
+    utils::{self, sleep},
     ActivityItem, HTLCStatus, DEFAULT_PAYMENT_TIMEOUT,
 };
 use bip39::Mnemonic;
@@ -19,6 +19,7 @@ use bitcoin::{
 };
 use fedimint_bip39::Bip39RootSecretStrategy;
 use fedimint_client::{
+    db::ChronologicalOperationLogKey,
     derivable_secret::DerivableSecret,
     get_config_from_db,
     oplog::{OperationLogEntry, UpdateStreamOrOutcome},
@@ -41,8 +42,8 @@ use fedimint_ln_common::LightningCommonInit;
 use fedimint_mint_client::MintClientInit;
 use fedimint_wallet_client::WalletClientInit;
 use futures::future::{self};
-use futures_util::{pin_mut, stream, StreamExt};
-use lightning::{log_debug, log_info, log_warn, util::logger::Logger};
+use futures_util::{pin_mut, StreamExt};
+use lightning::{log_debug, log_error, log_info, log_trace, log_warn, util::logger::Logger};
 use lightning_invoice::Bolt11Invoice;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use std::{
@@ -141,6 +142,7 @@ pub(crate) struct FederationClient<S: MutinyStorage> {
     pub(crate) fedimint_client: ClientArc,
     stopped_components: Arc<RwLock<Vec<bool>>>,
     storage: S,
+    g: GlueDB,
     network: Network,
     pub(crate) logger: Arc<MutinyLogger>,
     stop: Arc<AtomicBool>,
@@ -204,19 +206,34 @@ impl<S: MutinyStorage> FederationClient<S> {
             fedimint_client,
             stopped_components,
             storage,
+            g,
             network,
             logger,
             stop,
         })
     }
 
-    pub(crate) async fn get_invoice(&self, amount: u64) -> Result<MutinyInvoice, MutinyError> {
+    pub(crate) async fn get_invoice(
+        &self,
+        amount: u64,
+        labels: Vec<String>,
+    ) -> Result<MutinyInvoice, MutinyError> {
         let lightning_module = self
             .fedimint_client
             .get_first_module::<LightningClientModule>();
         let (_id, invoice) = lightning_module
             .create_bolt11_invoice(Amount::from_sats(amount), String::new(), None, ())
             .await?;
+
+        // persist the invoice
+        let mut stored_payment: MutinyInvoice = invoice.clone().into();
+        stored_payment.inbound = true;
+        stored_payment.labels = labels;
+
+        log_trace!(self.logger, "Persiting payment");
+        self.g.save_payment(stored_payment).await?;
+        log_trace!(self.logger, "Persisted payment");
+
         Ok(invoice.into())
     }
 
@@ -226,92 +243,215 @@ impl<S: MutinyStorage> FederationClient<S> {
     }
 
     pub async fn get_activity(&self) -> Result<Vec<ActivityItem>, MutinyError> {
-        let operations = self
-            .fedimint_client
-            .operation_log()
-            .list_operations(FEDIMINT_OPERATIONS_LIST_MAX, None)
-            .await;
+        log_trace!(self.logger, "Getting activity");
+        let payments = self.g.list_payments().await?;
+
+        let mut payments_map: HashMap<sha256::Hash, MutinyInvoice> = HashMap::new();
+        let mut pending_invoices: Vec<&MutinyInvoice> = Vec::new();
+
+        for payment in payments.iter() {
+            payments_map.insert(payment.payment_hash, payment.clone());
+            if matches!(payment.status, HTLCStatus::InFlight | HTLCStatus::Pending) {
+                pending_invoices.push(payment);
+            }
+        }
+
+        let operations = if !pending_invoices.is_empty() {
+            log_trace!(self.logger, "pending invoices, going to list operations");
+            self.fedimint_client
+                .operation_log()
+                .list_operations(FEDIMINT_OPERATIONS_LIST_MAX, None)
+                .await
+        } else {
+            vec![]
+        };
 
         let lightning_module = Arc::new(
             self.fedimint_client
                 .get_first_module::<LightningClientModule>(),
         );
 
-        let activity_stream = stream::iter(operations.into_iter());
-
-        let activity_items = activity_stream
-            .then(move |(key, entry)| {
-                let logger = self.logger.clone();
-                let lightning_module = Arc::clone(&lightning_module);
-                async move {
-                    if entry.operation_module_kind() == LightningCommonInit::KIND.as_str() {
-                        Some(
-                            handle_ln_meta(logger, &entry, key.operation_id, &lightning_module)
-                                .await,
-                        )
-                    } else {
-                        log_warn!(
-                            logger,
-                            "Unsupported module: {}",
-                            entry.operation_module_kind()
-                        );
-                        None
+        let mut operation_map: HashMap<
+            sha256::Hash,
+            (ChronologicalOperationLogKey, OperationLogEntry),
+        > = HashMap::new();
+        log_trace!(
+            self.logger,
+            "About to go through {} operations",
+            operations.len()
+        );
+        for (key, entry) in operations {
+            if entry.operation_module_kind() == LightningCommonInit::KIND.as_str() {
+                let lightning_meta: LightningOperationMeta = entry.meta();
+                match lightning_meta.variant {
+                    LightningOperationMetaVariant::Pay(pay_meta) => {
+                        operation_map.insert(*pay_meta.invoice.payment_hash(), (key, entry));
+                    }
+                    LightningOperationMetaVariant::Receive { invoice, .. } => {
+                        operation_map.insert(*invoice.payment_hash(), (key, entry));
                     }
                 }
+            }
+        }
+
+        log_trace!(
+            self.logger,
+            "Going through {} pending invoices to extract status",
+            pending_invoices.len()
+        );
+        for invoice in pending_invoices {
+            let hash = invoice.payment_hash;
+            if let Some((key, entry)) = operation_map.get(&hash) {
+                if let Some(updated_invoice) = extract_invoice_from_entry(
+                    self.logger.clone(),
+                    entry,
+                    &hash,
+                    key.operation_id,
+                    &lightning_module,
+                )
+                .await
+                {
+                    self.maybe_update_after_checking_fedimint(updated_invoice.clone())
+                        .await?;
+                    payments_map.insert(hash, updated_invoice);
+                }
+            }
+        }
+
+        let updated_payments = payments_map.into_values().collect::<Vec<_>>();
+
+        let activity_items = updated_payments
+            .into_iter()
+            .filter_map(|invoice| {
+                if !invoice
+                    .bolt11
+                    .as_ref()
+                    .is_some_and(|b| b.would_expire(utils::now()))
+                    && matches!(invoice.status, HTLCStatus::Succeeded | HTLCStatus::InFlight)
+                {
+                    Some(ActivityItem::Lightning(Box::new(invoice)))
+                } else {
+                    None
+                }
             })
-            .filter_map(|item| async move { item })
-            .collect::<Vec<ActivityItem>>()
-            .await;
+            .collect::<Vec<_>>();
 
         Ok(activity_items)
+    }
+
+    async fn maybe_update_after_checking_fedimint(
+        &self,
+        updated_invoice: MutinyInvoice,
+    ) -> Result<(), MutinyError> {
+        if matches!(
+            updated_invoice.status,
+            HTLCStatus::Succeeded | HTLCStatus::Failed
+        ) {
+            log_debug!(self.logger, "Saving updated payment");
+            self.g
+                .update_payment_status(
+                    &updated_invoice.payment_hash,
+                    updated_invoice.status.clone(),
+                )
+                .await?;
+            self.g
+                .update_payment_fee(&updated_invoice.payment_hash, updated_invoice.fees_paid)
+                .await?;
+            self.g
+                .update_payment_preimage(
+                    &updated_invoice.payment_hash,
+                    updated_invoice.preimage.clone(),
+                )
+                .await?;
+        }
+        Ok(())
     }
 
     pub async fn get_invoice_by_hash(
         &self,
         hash: &sha256::Hash,
     ) -> Result<MutinyInvoice, MutinyError> {
-        let operations = self
-            .fedimint_client
-            .operation_log()
-            .list_operations(FEDIMINT_OPERATIONS_LIST_MAX, None)
-            .await;
+        log_trace!(self.logger, "get_invoice_by_hash");
 
-        let lightning_module = self
-            .fedimint_client
-            .get_first_module::<LightningClientModule>();
+        // Try to get the invoice from storage first
+        let invoice = match self.g.get_payment(hash).await {
+            Ok(i) => i,
+            Err(e) => {
+                log_error!(self.logger, "could not get invoice by hash: {e}");
+                return Err(e);
+            }
+        };
 
-        for (key, entry) in operations {
-            if entry.operation_module_kind() == LightningCommonInit::KIND.as_str() {
-                if let Some(invoice) = extract_invoice_from_entry(
-                    self.logger.clone(),
-                    &entry,
-                    hash,
-                    key.operation_id,
-                    &lightning_module,
-                )
-                .await
-                {
-                    return Ok(invoice);
+        if let Some(invoice) = invoice {
+            log_trace!(self.logger, "retrieved invoice by hash");
+
+            if matches!(invoice.status, HTLCStatus::InFlight | HTLCStatus::Pending) {
+                log_trace!(self.logger, "invoice still in flight, getting operations");
+                // If the invoice is InFlight or Pending, check the operation log for updates
+                let lightning_module = self
+                    .fedimint_client
+                    .get_first_module::<LightningClientModule>();
+
+                let operations = self
+                    .fedimint_client
+                    .operation_log()
+                    .list_operations(FEDIMINT_OPERATIONS_LIST_MAX, None)
+                    .await;
+
+                log_trace!(
+                    self.logger,
+                    "going to go through {} operations",
+                    operations.len()
+                );
+                for (key, entry) in operations {
+                    if entry.operation_module_kind() == LightningCommonInit::KIND.as_str() {
+                        if let Some(updated_invoice) = extract_invoice_from_entry(
+                            self.logger.clone(),
+                            &entry,
+                            hash,
+                            key.operation_id,
+                            &lightning_module,
+                        )
+                        .await
+                        {
+                            self.maybe_update_after_checking_fedimint(updated_invoice.clone())
+                                .await?;
+                            return Ok(updated_invoice);
+                        }
+                    } else {
+                        log_warn!(
+                            self.logger,
+                            "Unsupported module: {}",
+                            entry.operation_module_kind()
+                        );
+                    }
                 }
             } else {
-                log_warn!(
-                    self.logger,
-                    "Unsupported module: {}",
-                    entry.operation_module_kind()
-                );
+                // If the invoice is not InFlight or Pending, return it directly
+                log_trace!(self.logger, "returning final invoice");
+                return Ok(invoice);
             }
         }
 
+        log_debug!(self.logger, "could not find invoice");
         Err(MutinyError::NotFound)
     }
 
     pub(crate) async fn pay_invoice(
         &self,
         invoice: Bolt11Invoice,
+        labels: Vec<String>,
     ) -> Result<MutinyInvoice, MutinyError> {
+        // Save before sending
+        let mut stored_payment: MutinyInvoice = invoice.clone().into();
+        stored_payment.inbound = false;
+        stored_payment.labels = labels;
+        self.g.save_payment(stored_payment.clone()).await?;
+
         let lightning_module = self
             .fedimint_client
             .get_first_module::<LightningClientModule>();
+
         let outgoing_payment = lightning_module
             .pay_bolt11_invoice(invoice.clone(), ())
             .await?;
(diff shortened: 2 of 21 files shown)


mutiny-core/Cargo.toml
```diff
@@ -41,6 +41,12 @@ aes = { version = "0.8" }
 jwt-compact = { version = "0.8.0-beta.1", features = ["es256k"] }
 argon2 = { version = "0.5.0", features = ["password-hash", "alloc"] }
 
+fedimint-client = { git = "https://github.com/fedimint/fedimint", branch = "master" }
+fedimint-core = { git = "https://github.com/fedimint/fedimint", branch = "master" }
+fedimint-wallet-client = { git = "https://github.com/fedimint/fedimint", branch = "master"}
```
benthecarman, line 46: We might want to pick a commit and stick to it until 0.2.0, otherwise we might get random compile errors anytime they change an api

mutiny-core/src/fedimint.rs
```diff
@@ -0,0 +1,139 @@
...
+    xprivkey: ExtendedPrivKey,
+    fedimint_index: &FedimintIndex,
+) -> Result<DerivableSecret, MutinyError> {
+    let context = Secp256k1::new();
+    let xpriv = xprivkey.derive_priv(
+        &context,
+        &DerivationPath::from(vec![
+            ChildNumber::from_hardened_idx(1)?,
+            ChildNumber::from_hardened_idx(fedimint_index.child_index)?,
+        ]),
```
benthecarman, line 134: we should incorporate the network into the secret, so mainnet is different than testnet

mutiny-core/src/fedimint.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub uuid: String,
+    pub federation_id: FederationId,
+}
+
+// This is the FedimintIndex reference that is saved to the DB
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone, Default)]
+pub struct FedimintIndex {
+    pub child_index: u32,
+    pub federation_code: String,
+    pub archived: Option<bool>,
```
benthecarman, line 44: dont think this needs to be optional, we only did that on node index for backwards compatibility

mutiny-core/src/nodemanager.rs
```diff
@@ -708,6 +714,35 @@ impl<S: MutinyStorage> NodeManager<S> {
...
+                logger.clone(),
+            )
+            .await?;
+
+            let id = fedimint.fedimint_client.federation_id();
+
+            fedimint_map.insert(id, Arc::new(fedimint));
+        }
+
+        let fedimints = Arc::new(Mutex::new(fedimint_map));
```
benthecarman, line 744: should these be in lib instead of nodemanager? Imo that'd make more sense

mutiny-core/src/fedimint.rs
TonyGiorgio: Shouldnt we really be doing that at the highest level instead? We should want this everywhere, not just for fedimint.

mutiny-core/src/nodemanager.rs
TonyGiorgio: moved, might still clean it up a bit in lib though.

mutiny-core/src/fedimint.rs
benthecarman: We could but really wouldn't be following the BIPs anymore, normally for derivation paths it is `m/Purpose/Network/Account/Internal/Index`

mutiny-core/src/fedimint.rs
```diff
@@ -0,0 +1,271 @@
...
+        let mut inv: MutinyInvoice = invoice.clone().into();
+        match outgoing_payment.payment_type {
+            fedimint_ln_client::PayType::Internal(pay_id) => {
+                // TODO merge the two as much as we can
+                let pay_outcome = lightning_module
+                    .subscribe_internal_pay(pay_id)
+                    .await
+                    .map_err(|_| MutinyError::ConnectionFailed)?; // TODO resume if we error here?
+
+                let mut inv: MutinyInvoice = invoice.clone().into();
```
benthecarman, line 197: you already did this above on 188

benthecarman, line 197: you already did this above on 188

mutiny-core/src/fedimint.rs
```diff
@@ -0,0 +1,271 @@
...
+pub struct FedimintIdentity {
+    pub uuid: String,
+    pub federation_id: FederationId,
+}
+
+// This is the FedimintIndex reference that is saved to the DB
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone, Default)]
+pub struct FedimintIndex {
+    pub child_index: u32,
+    pub federation_code: String,
```
benthecarman, line 91: You should be able to make this `InviteCode` so you don't need to parse it later

benthecarman: looks like there is a bunch of places where you are using `String` would be nice to use `InviteCode` instead

mutiny-core/src/nodemanager.rs
```diff
@@ -1662,75 +1604,72 @@ impl<S: MutinyStorage> NodeManager<S> {
...
-                "could not find internal node {self_node_pubkey}"
-            );
-            Err(MutinyError::NotFound)
-        }
+        let nodes = self.nodes.lock().await;
+        let node = match self_node_pubkey {
+            Some(pubkey) => nodes.get(pubkey),
+            None => nodes.iter().next().map(|(_, node)| node),
+        };
+        let node = node.ok_or(MutinyError::NotFound)?;
```
benthecarman, line 1669: should make this a function, you do this a lot

mutiny-core/src/lib.rs
```diff
@@ -359,6 +418,128 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+                    // Check if the federation has enough balance
+                    let balance = fedimint_client.get_balance().await?;
+                    if balance >= invoice_amount / 1_000 {
+                        // Try to pay the invoice using the federation
+                        return fedimint_client.pay_invoice(inv).await;
+                    }
+                }
+                // If invoice amount is None or balance is not sufficient, fall through to node_manager payment
+            }
+            // If federation client is not found, fall through to node_manager payment
```
benthecarman, line 452: should also iterate through the fedimint clients if first fails

mutiny-core/src/lib.rs
```diff
@@ -359,6 +418,128 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+            labels,
+        })
+    }
+
+    /// Gets the current balance of the wallet.
+    /// This includes both on-chain and lightning funds.
+    ///
+    /// This will not include any funds in an unconfirmed lightning channel.
+    pub async fn get_balance(&self) -> Result<MutinyBalance, MutinyError> {
+        Ok(MutinyBalance::new(self.node_manager.get_balance().await?))
```
benthecarman, line 540: fedimint balance should be added

benthecarman (commented):
looking really good! just a couple nits

mutiny-core/src/sql/glue.rs
```diff
@@ -0,0 +1,566 @@
+#![allow(dead_code, unused_variables, unused_imports, unreachable_code)] // TODO remove
```
TonyGiorgio, line 1: @elsirion could you check the approach for this `glue.rs` file whenever you have a chance?

The approach I've taken for the fedimint traits is to have a KV table on top of gluesql (w/ indexeddb under the hood) where each fedimint client will have their own row. I'm just doing a dump of the values on `commit` like you are with webimint. I tried a few different approaches but came back to that one being the safest while still adhering to the trait requirements.

TonyGiorgio:
I'm thinking about going ahead and implementing `MutinyStorage` for `GlueDB`, but as another underlying `GlueKV` struct that is specifically just for KV data like fedimint and our existing data. Then `GlueKV` can pull in `VSSClient` and handle all of the versioning and existing data that we're storing as a KV anyways (all of it currently).

If I do that, then at that point we can just do a clean migration from IndexedDB to Glue. 

Then at a later point, we can start to add structured SQL queries for data, like payments and NWC things. However, I'm still a bit unsure of how we should best handle the structured SQL to KV after we migrate to that. 

I'm also wondering if we can now get rid of the in memory preload of all of the data. We could do the `GlueKV` `VSS` integration always doing the remote version check for `get` and then retrieve if there's a difference. Would speed up initial load times. 

Lastly, I'm wondering how much farther we should go with this PR before we ship an early alpha of fedimint. I'd like to get VSS integrations in there first but that opens all of these cans of worms. It's definitely possible we go without it and just warn our early alpha users that these funds are subject to complete loss (which is to be expected whether we have VSS or not for this testing period). 

Thoughts @benthecarman ?

benthecarman:
If that's the plan I think it'd make sense to merge this before vss. 

I have been thinking about if we can remove the memory preload. I think it'd be possible because we can use async storage pretty much everywhere. If anything still needs it we could just preload that.

TonyGiorgio:
I think one of the other last things to do is making sure we can do NWC stuff out of the fedimint. I should probably do that before getting into complicated storage stuff.

benthecarman:
Yeah I think we should hold off on touching the rest of storage until we have fedimint working in production

mutiny-core/src/lib.rs
TonyGiorgio: No fedimint funds exist at this stage in the commit but yes federation balances have been added. I'd rather not worry about the details of multi federation yet.

mutiny-wasm/src/models.rs
```diff
@@ -417,17 +418,88 @@ impl MutinyBalance {
...
+            federation: m.federation,
             force_close: m.force_close,
         }
     }
 }
 
+#[derive(Serialize, Deserialize, Clone, Eq, PartialEq)]
+#[wasm_bindgen]
+pub struct FederationBalance {
+    identity: FedimintIdentity,
```
TonyGiorgio, line 436: Rename all `Fedimint` references to `Federation` for consistency.

mutiny-core/src/fedimint.rs
TonyGiorgio: Not going to be able to comply exactly with their derivations but the prefix essentially does once is goes into the fedimint client builder. I've also implemented network into this.

TonyGiorgio:
Could use another review. Last thing I need to hook up is the zap stuff, but otherwise I've cleaned this up a ton and handled all previous PR feedback. Tested pretty thoroughly too.

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+use lightning::{log_warn, util::logger::Logger};
+use lightning_invoice::Bolt11Invoice;
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::fmt::Debug;
+use std::{
+    collections::HashMap,
+    sync::{atomic::AtomicBool, Arc, RwLock},
+};
+
+const FEDIMINT_CLIENT_NONCE: &[u8] = b"Fedimint Client Salt";
```
benthecarman, line 47: is this supposed to be a fixed value? would be nice to have a comment explaining what this does/is for

benthecarman: maybe we should put the name mutiny in it to so we don't clash with other clients? unless we want to?

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+            LnPayState::UnexpectedError { .. } => HTLCStatus::Failed,
+        }
+    }
+}
+
+// This is the FederationStorage object saved to the DB
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
+pub struct FederationStorage {
+    pub federations: HashMap<String, FederationIndex>,
+    #[serde(default)]
```
benthecarman, line 94: shouldn't need this `#[serde(default)]`, only needed that for the node storage because we had ones without a version.

Might be better to just use `VersionedValue` instead so we don't need to add another case for serialization

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+    pub(crate) fedimint_client: ClientArc,
+    stopped_components: Arc<RwLock<Vec<bool>>>,
+    storage: S,
+    network: Network,
+    pub(crate) logger: Arc<MutinyLogger>,
+    stop: Arc<AtomicBool>,
+}
+
+// TODO remove
+#[allow(dead_code)]
```
benthecarman, line 139: think you can remove this now

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+        g: GlueDB,
+        network: Network,
+        logger: Arc<MutinyLogger>,
+    ) -> Result<Self, MutinyError> {
+        log_info!(logger, "initializing a new federation client: {uuid}");
+
+        // a list of components that need to be stopped and whether or not they are stopped
+        let stopped_components = Arc::new(RwLock::new(vec![]));
+
+        let stop = Arc::new(AtomicBool::new(false));
```
benthecarman, line 157: shouldn't we be getting this from `lib.rs`?

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+        );
+
+        let activity_stream = stream::iter(operations.into_iter());
+
+        let activity_items = activity_stream
+            .then(move |(key, entry)| {
+                let logger = self.logger.clone();
+                let lightning_module = Arc::clone(&lightning_module);
+                async move {
+                    if entry.operation_module_kind() == "ln" {
```
benthecarman, line 244: is there a way to get around this magic string, can we match on the type instead?

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+            .operation_log()
+            .list_operations(100, None)
+            .await;
+
+        let lightning_module = self
+            .fedimint_client
+            .get_first_module::<LightningClientModule>();
+
+        for (key, entry) in operations {
+            if entry.operation_module_kind() == "ln" {
```
benthecarman, line 281: same here

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+    }
+
+    pub async fn get_invoice_by_hash(
+        &self,
+        hash: &sha256::Hash,
+    ) -> Result<MutinyInvoice, MutinyError> {
+        let operations = self
+            .fedimint_client
+            .operation_log()
+            .list_operations(100, None)
```
benthecarman, line 273: i think this only looks back at the last hundred, we'll need to add handling if you've done over 100 txs

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+
+    pub(crate) async fn pay_invoice(
+        &self,
+        invoice: &Bolt11Invoice,
+    ) -> Result<MutinyInvoice, MutinyError> {
+        let lightning_module = self
+            .fedimint_client
+            .get_first_module::<LightningClientModule>();
+        let outgoing_payment = lightning_module
+            .pay_bolt11_invoice(invoice.clone(), ())
```
benthecarman, line 313: if we're just immediately cloning then no need to pass in a reference

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+        let inv = match outgoing_payment.payment_type {
+            fedimint_ln_client::PayType::Internal(pay_id) => {
+                match lightning_module.subscribe_internal_pay(pay_id).await {
+                    Ok(o) => {
+                        process_outcome(
+                            o,
+                            process_pay_state_internal,
+                            invoice.clone(),
+                            true,
+                            30_000,
```
benthecarman, line 326: should just pull in `DEFAULT_PAYMENT_TIMEOUT * 1_000` from `node.rs`  instead of this magic value

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+            }
+            fedimint_ln_client::PayType::Lightning(pay_id) => {
+                match lightning_module.subscribe_ln_pay(pay_id).await {
+                    Ok(o) => {
+                        process_outcome(
+                            o,
+                            process_pay_state_ln,
+                            invoice.clone(),
+                            false,
+                            30_000,
```
benthecarman, line 342: same here

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+}
+
+fn network_to_bip44_chain(network: Network) -> u32 {
+    match network {
+        Network::Bitcoin => 0,
+        Network::Testnet => 1,
+        Network::Signet => 1,
+        Network::Regtest => 1,
+    }
+}
```
benthecarman, line 399: we are doing this in `onchain.rs`, should move this function to utils and call it in both places

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+        LightningOperationMetaVariant::Pay(pay_meta) => {
+            if pay_meta.invoice.payment_hash() == hash {
+                match lightning_module.subscribe_ln_pay(operation_id).await {
+                    Ok(o) => Some(
+                        process_outcome(
+                            o,
+                            process_pay_state_ln,
+                            pay_meta.invoice,
+                            false,
+                            30,
```
benthecarman, line 420: this 30 number is here again but not multiplied by 1k, probably a bug? at the very least should be a `const`

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+                }
+            } else {
+                None
+            }
+        }
+        LightningOperationMetaVariant::Receive { invoice, .. } => {
+            if invoice.payment_hash() == hash {
+                match lightning_module.subscribe_ln_receive(operation_id).await {
+                    Ok(o) => Some(
+                        process_outcome(o, process_receive_state, invoice, true, 30, logger).await,
```
benthecarman, line 435: same here

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+    operation_id: OperationId,
+    lightning_module: &LightningClientModule,
+) -> ActivityItem {
+    let lightning_meta: LightningOperationMeta = entry.meta();
+
+    let invoice = match lightning_meta.variant {
+        LightningOperationMetaVariant::Pay(pay_meta) => {
+            match lightning_module.subscribe_ln_pay(operation_id).await {
+                Ok(o) => {
+                    process_outcome(o, process_pay_state_ln, pay_meta.invoice, false, 30, logger)
```
benthecarman, line 458: same here

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+                        .await
+                }
+                Err(_) => pay_meta.invoice.into(),
+            }
+        }
+        LightningOperationMetaVariant::Receive { invoice, .. } => {
+            match lightning_module.subscribe_ln_receive(operation_id).await {
+                Ok(o) => process_outcome(o, process_receive_state, invoice, true, 30, logger).await,
```
benthecarman, line 466: same here

mutiny-core/src/lib.rs
```diff
@@ -355,6 +532,201 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+        if inv.network() != self.config.network {
+            return Err(MutinyError::IncorrectNetwork(inv.network()));
+        }
+
+        // Try each federation first
+        let federation_ids = self.list_federations().await?;
+        for federation_id in federation_ids {
+            if let Some(fedimint_client) = self.federations.lock().await.get(&federation_id) {
+                // Check the amount specified in the invoice
+                if let Some(invoice_amount) = inv.amount_milli_satoshis() {
```
benthecarman, line 554: should do this check before `list_federations` otherwise we'll have to loop through each time for no reason

mutiny-core/src/lib.rs
```diff
@@ -355,6 +532,201 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+    ///
+    /// If the server returns a status of 500 with a different error message,
+    /// a [`MutinyError::LspGenericError`] is returned.
+    pub async fn create_bip21(
+        &self,
+        amount: Option<u64>,
+        labels: Vec<String>,
+    ) -> Result<MutinyBip21RawMaterials, MutinyError> {
+        let invoice = if self.config.safe_mode {
+            None
```
benthecarman, line 621: we can probably enable fedimint in safe mode? probably a better discussion for the future however

mutiny-core/src/lib.rs
```diff
@@ -630,6 +994,231 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+
+            total_balance += balance;
+        }
+
+        Ok(total_balance)
+    }
+
+    pub async fn get_federation_balances(&self) -> Result<FederationBalances, MutinyError> {
+        let federation_ids = self.list_federations().await?;
+        let mut balances = Vec::new();
```
benthecarman, line 1090: nit: can do `let mut balances = Vec::with_capacity(federation_ids.len());`

benthecarman (commented):
Overall looking good, mostly nits

Would be good to try and document at least all the public functions so it easier to grok as well.

mutiny-core/src/lib.rs
TonyGiorgio: Nah, in case something is going wrong with fedimint, it would be good for now to leave that unaccessible for now. Until we test more at least.

mutiny-core/src/federation.rs
TonyGiorgio: Idk about `VersionedValue`. We're only currently doing it for the channel manager, and I think that's mostly so that we can compare it's internal value with the value stored, right? Why don't we use it for channel monitor, nodes, locks, etc? 

Also, I would rather be passing around `FederationStorage` to multiple places instead of passing around and parsing a generic `Value`. We've got to parse it somewhere, and should be using real values. Plus we never know what we're going to add to this later.

mutiny-core/src/federation.rs
TonyGiorgio: If someone derrives their `m/1'/X`keys, they could use it in theory with other fedimint clients. So it's a feature that might not ever be necessary but could be a bonus for interoperability. I added a comment about it.

mutiny-core/src/nodemanager.rs
```diff
@@ -1461,7 +1326,7 @@ impl<S: MutinyStorage> NodeManager<S> {
             .map(|bal| bal.claimable_amount_satoshis())
             .sum();
 
-        Ok(MutinyBalance {
+        Ok(OnChainLnBalance {
```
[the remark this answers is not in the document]
benthecarman: Maybe `NodesBalance`

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,612 @@
...
+        }
+
+        Err(MutinyError::NotFound)
+    }
+
+    pub(crate) async fn pay_invoice(
+        &self,
+        invoice: &Bolt11Invoice,
+    ) -> Result<MutinyInvoice, MutinyError> {
+        let lightning_module = self
```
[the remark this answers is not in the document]
TonyGiorgio: There's a couple ways we can go about this and I'm unsure of how we should do it. 

Unfortunately, we save a ton of things inside of `LdkStorage` and this includes all payment information. Another unfortunate thing is we append the node_id to all of the keys. So, if we want to store and recheck if we've made a payment, there's a few options: 
1. Ignore if we possibly paid it through the self custodial node
2. Pull in `storage` and save it similarly on the fedimint side in the KV store (only checking fedimint)
3. Move payments to GlueSQL for fedimint or also for node payments
4. Do nothing for now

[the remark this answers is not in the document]
benthecarman: Maybe a combo of 1 and 4 would be good here. Start storing fedimint payments in gluesql and we can plan to move the node ones over in a future update

benthecarman:
We can pull out the `Move log and stop outside of nodemanager ` commit separately and just merge that for now if we want, not sure how close we are getting here

mutiny-core/src/lib.rs
```diff
@@ -525,6 +526,31 @@ impl<S: MutinyStorage> MutinyWallet<S> {
             return Err(MutinyError::IncorrectNetwork(inv.network()));
         }
 
+        // Try each federation first
```
benthecarman, line 552: this comment needs to be moved down

mutiny-core/src/federation.rs
```diff
@@ -0,0 +1,216 @@
...
+        &DerivationPath::from(vec![
+            ChildNumber::from_hardened_idx(1)?,
+            ChildNumber::from_hardened_idx(coin_type_from_network(network))?,
+            ChildNumber::from_hardened_idx(federation_index.child_index)?,
+        ]),
+    )?;
+    let secret =
+        DerivableSecret::new_root(&xpriv.private_key.secret_bytes(), FEDIMINT_CLIENT_NONCE);
+    Ok(secret)
+}
```
benthecarman, line 400: Was thinking about this more, do we need the `federation_index`? Don't see why we need different keys for each federation. I don't think there is a privacy implication since it is all ecash, could be wrong though.

If we got rid of it it'd make things easier where if the user has their seed they don't also need to know the index for each federation.

mutiny-core/src/lib.rs
```diff
@@ -876,6 +893,182 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+            federations_guard.remove(&federation_id);
+        } else {
+            return Err(MutinyError::NotFound);
+        }
+
+        Ok(())
+    }
+}
+
+async fn create_federation<S: MutinyStorage>(
```
benthecarman, line 1113: nit: this should be named `create_federations`

mutiny-core/src/lib.rs
```diff
@@ -876,6 +893,182 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+    // Begin with a mutex lock so that nothing else can
+    // save or alter the federation list while it is about to
+    // be saved.
+    let mut federation_mutex = federation_storage.lock().await;
+
+    // Get the current federations and their bip32 indices
+    // so that we can create another federation with the next.
+    // Always get it from our storage, the federation_mutex is
+    // mostly for read only and locking.
+    let mut existing_federations = storage.get_federations()?;
```
benthecarman, line 1177: should we check that the new federation isn't in our current list. Could see where you archive one and then add it back, we'd have it twice in the list, could just unarchive in that case, if it is active then can throw an error.

mutiny-core/src/lib.rs
```diff
@@ -876,6 +893,182 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+        network,
+        logger.clone(),
+        stop,
+    )
+    .await;
+
+    let new_federation = match new_federation_res {
+        Ok(new_federation) => new_federation,
+        Err(e) => return Err(e),
+    };
```
benthecarman, line 1220: pretty sure this is the exact same as `new_federation_res?`

mutiny-core/src/lib.rs
```diff
@@ -876,6 +893,182 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+
+    let federation_id = new_federation.fedimint_client.federation_id();
+    federations
+        .clone()
```
benthecarman, line 1224: dont think you need this clone

mutiny-core/src/lib.rs
```diff
@@ -655,7 +668,28 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+        // Get federation activities and merge them with node manager activities
+        let federation_ids = self.list_federations().await?;
+        for fed_id in federation_ids {
+            let federation_activities = self
+                .federations
+                .lock()
+                .await
+                .get(&fed_id)
+                .ok_or(MutinyError::NotFound)?
+                .get_activity()
```
benthecarman, line 693: this seems inefficient, we get the ids and individually get each client, can we not do this?

```rust 
let federations = self.federations.lock().await;

for (fed_id, federation) in federations {
	let federation_activities= federation.get_activity().await?;
	activities.extend(federation_activities);
}

```

mutiny-core/src/lib.rs
```diff
@@ -670,7 +704,23 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+        // If not found, search in federations
+        // FIXME: this is incredibly inefficient without a KV mapping operationid:hash
+        let federation_ids = self.list_federations().await?;
+        for fed_id in federation_ids {
+            if let Some(fed_client) = self.federations.lock().await.get(&fed_id) {
+                if let Ok(invoice) = fed_client.get_invoice_by_hash(hash).await {
+                    return Ok(invoice);
+                }
+            }
+        }
```
benthecarman, line 731: same here

mutiny-core/src/sql/glue.rs
```diff
@@ -0,0 +1,609 @@
...
+    #[tokio::test]
+    async fn create_glue_storage_tests() {
+        // TODO reactivate
+        // create_glue_and_fedimint_storage().await;
+    }
+
+    #[tokio::test]
+    async fn create_glue_storage_value_tests() {
+        // TODO reactivate
+        // create_glue_storage_value().await;
```
benthecarman, line 577: can we enable these now?

mutiny-core/src/sql/glue.rs
```diff
@@ -0,0 +1,609 @@
...
+        Ok(Self {
+            db,
+            federation_id,
+            fedimint_memory: Arc::new(fedimint_memory),
+            logger,
+        })
+    }
+}
+
+fn key_id(federation_id: String) -> String {
```
benthecarman, line 209: nit could make this a `&str` so you don't need clones

benthecarman (commented):
Went through commit by commit this time. Mostly nits again, what is left before we can merge this?

mutiny-core/src/sql/glue.rs
TonyGiorgio: No, glue can't run on a server. I may need to just delete these.

mutiny-core/src/federation.rs
TonyGiorgio: This is a valid point. I can remove it and just use the master seed but be aware, both fedimint and our LN nodes will be using `m/0'/` as the first part of the prefix. I'm afraid of potential clashes.

TonyGiorgio: These secondary indexes should be hardcoded and stored somewhere.

mutiny-core/src/federation.rs
TonyGiorgio: Not handling for now, will worry about this as we tighten up more storage stuff.

TonyGiorgio:
I got payment status working though labels don't for some reason. Going to clean up and get that working tomorrow.

mutiny-core/src/nostr/nwc.rs
```diff
@@ -1295,13 +1316,21 @@ mod wasm_test {
...
+            .unwrap()
+            .persister
             .persist_payment_info(invoice.payment_hash().as_inner(), &payment_info, false)
             .unwrap();
         let event = create_nwc_request(&uri, invoice.to_string());
-        let result = nwc.handle_nwc_request(event, &node, &nostr_manager).await;
+        let result = nwc.handle_nwc_request(event, &mw, &nostr_manager).await;
         assert_eq!(result.unwrap(), None);
-        check_no_pending_invoices(&storage);
+        // FIXME: Not working
```
TonyGiorgio, line 1332: @benthecarman I couldn't get a couple of these NWC asserts to work and I've tried for quite a bit. I tested budgetting and auto approval pretty thoroughly for the node & fedimint payer, so I don't think I've introduced a bug here. TBH, I'm a bit unsure how how these tests actually work since I don't think these nodes have funds and don't attempt to pay? 

We can either fix in a follow up or if you can investigate this further, go ahead.

TonyGiorgio:
This is ready @benthecarman 

I say we handle LNURL pay/withdraw in a follow up along with you fixing a couple of the NWC tests in a follow up too. I don't think they'd require any foundational change, mostly a bit of manual work that's not worth blocking on.

mutiny-core/src/lib.rs
```diff
@@ -93,15 +93,14 @@ use mockall::{automock, predicate::*};
...
+pub trait LnPayer {
     fn logger(&self) -> &MutinyLogger;
     fn skip_hodl_invoices(&self) -> bool;
-    fn get_outbound_payment_status(&self, payment_hash: &[u8; 32]) -> Option<HTLCStatus>;
+    async fn get_outbound_payment_status(&self, payment_hash: &[u8; 32]) -> Option<HTLCStatus>;
     async fn pay_invoice_with_timeout(
         &self,
         invoice: &Bolt11Invoice,
         amt_sats: Option<u64>,
-        timeout_secs: Option<u64>,
```
benthecarman: since you removed the timeout can rename this to just `pay_invoice`

mutiny-core/src/lib.rs
```diff
@@ -440,6 +439,7 @@ impl<S: MutinyStorage> MutinyWallet<S> {
     pub(crate) async fn start_nostr_wallet_connect(&self, from_node: PublicKey) {
         let nostr = self.nostr.clone();
         let nm = self.node_manager.clone();
```
benthecarman, line 441: i think we're only getting the node manager for the logger now, could just replace with self.logger

benthecarman (commented):
only 2 nits

mutiny-core/src/lib.rs
TonyGiorgio: Glad you caught this. It's more involved because `node` is being retrieved and called upon actually. Not everywhere was it using the `LnPayer` trait to find the status of a payment. But I should be able to get this tonight.

mutiny-core/src/lib.rs
```diff
@@ -653,33 +660,9 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
-            } else {
-                // Fallback to node_manager invoice creation if no federation is found
-                let inv = self
-                    .node_manager
-                    .create_invoice(amount, labels.clone())
-                    .await?;
-                Some(inv.bolt11.ok_or(MutinyError::WalletOperationFailed)?)
+            match self.create_lightning_invoice(amount, labels.clone()).await {
+                Ok(invoice) => Some(invoice), // Assuming create_lightning_invoice returns a Result
+                Err(_) => None, // You can handle the error more specifically if needed
```
benthecarman, line 665: We were returning an error here before, should probably keep that functionality at least for now.
otherwise this is the exact same as `ok()`

TonyGiorgio:
@benthecarman please review the last 3 commits.
I've done the following: 
- Removed more references to `node_manager` and `node` inside of `nwc`
- Added another method to the trait in order to handle invoice creation (for one time nwc)
- I've renamed `LnPayer` to `InvoiceHandler` since it now generates, checks, and pays invoices

mutiny-core/src/lib.rs
TonyGiorgio: We didn't return error here, when we create bip21, we return the address that's created but without the invoice, and display a warning.

mutiny-core/src/lib.rs
```diff
@@ -689,12 +672,40 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+                    .await
+                {
+                    return Ok(inv);
+                }
+            }
+        }
+
+        // Fallback to node_manager invoice creation if no federation invoice created
+        self.node_manager
+            .create_invoice(amount, labels.clone())
```
benthecarman, line 705: nit: unneeded clone

mutiny-core/src/lib.rs
TonyGiorgio: This is what it was before this PR: 

```
        } else {
            match self.create_lightning_invoice(amount, labels.clone()).await {
                Ok(invoice) => Some(invoice),
                Err(_) => None,
            }
        };
```

I was incorrect in making it return an error when I introduced fedimint.

benthecarman (commented):
one tiny nit, otherwise lgtm.

You can just take my commit and merge it into your last one if you want, would make the commits a little cleaner.

mutiny-core/src/lib.rs
benthecarman: Okay, we can at least clean this up a bit with `ok()` if you want

TonyGiorgio:
> one tiny nit, otherwise lgtm.
> 
> You can just take my commit and merge it into your last one if you want, would make the commits a little cleaner.

Done

mutiny-core/src/lib.rs
TonyGiorgio: Doing `ok().and_then()` here now.

benthecarman (approved):

benthecarman:
:ship: it 🚀

PR #868, opened 2023-11-20
Device Lock Improvements

benthecarman:
Broken up into 3 commits. 

First allows us to get the remaining time in the device lock.

Second makes it so we only write the lock to VSS instead of to both VSS and indexeddb. This should let us reduce our  timer between writes because we won't be running into limits with browser storage.

Last commit reduces the timer to 30 seconds, this way the user only has to max wait 60 seconds to startup instead of 120.

### DIFF
--- a/mutiny-wasm/src/lib.rs
+++ b/mutiny-wasm/src/lib.rs
@@ -33,7 +33,7 @@ use mutiny_core::lnurlauth::AuthManager;
 use mutiny_core::nostr::nwc::{BudgetedSpendingConditions, NwcProfileTag, SpendingConditions};
 use mutiny_core::redshift::RedshiftManager;
 use mutiny_core::redshift::RedshiftRecipient;
-use mutiny_core::storage::MutinyStorage;
+use mutiny_core::storage::{DeviceLock, MutinyStorage, DEVICE_LOCK_KEY};
 use mutiny_core::utils::sleep;
 use mutiny_core::vss::MutinyVssClient;
 use mutiny_core::{labels::LabelStorage, nodemanager::NodeManager};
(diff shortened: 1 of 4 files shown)


mutiny-wasm/src/lib.rs
```diff
@@ -264,6 +264,70 @@ impl MutinyWallet {
         NodeManager::has_node_manager(storage)
     }
 
+    /// Returns the number of remaining seconds until the device lock expires.
+    #[wasm_bindgen]
+    pub async fn get_device_lock_remaining_secs(
+        password: Option<String>,
```
TonyGiorgio, line 270: Requiring password here? Is that double password requirement then?

TonyGiorgio: Is it going to require the user to type in the password twice then? Once for the wallet unlock attempt and then again for the estimation about how long to wait? Why not just return the error with the amount of remaining seconds required?

benthecarman: > Is it going to require the user to type in the password twice then? Once for the wallet unlock attempt and then again for the estimation about how long to wait?

No we can cache it in the front end

> Why not just return the error with the amount of remaining seconds required?

because we can't put params in the error for wasm

TonyGiorgio (approved):
If you've tested the encryption part of this, then LGTM

benthecarman:
> If you've tested the encryption part of this, then LGTM

Tested it manually and it works, not sure the best way to implement it in the front end though, will leave @futurepaul for that

TonyGiorgio:
I see that in your PR on mutiny-web, you're letting it fall back to 60s if it failed to parse things, so I think that's acceptable here. Think we can merge and cut a release if you want.

PR #870, opened 2023-11-22
Auto tag lnurl payments

benthecarman:
This makes it so if you are paying an lnurl or lightning address we search your contacts for it and try to tag the transaction.

to test:

1. sync nostr contact, make sure you follow: `npub1qf8e8cvfp60ywrah984zgsn8veggcrsv2cvttdr47tgz05ypf5yszwx308`
2. pay to `mutiny@lnurl-staging.mutinywallet.com` with no tags
3. See that in your payment history it is tagged

### DIFF
--- a/mutiny-core/src/labels.rs
+++ b/mutiny-core/src/labels.rs
@@ -76,6 +76,21 @@ impl Contact {
         };
         init.update_with_metadata(metadata)
     }
+
+    /// Checks if the contact has the given lnurl as either a lnurl or a lightning address
+    pub fn has_lnurl(&self, lnurl: &LnUrl) -> bool {
+        if self.lnurl.as_ref().is_some_and(|l| l == lnurl) {
+            return true;
+        }
+
+        if let Some(ln_address) = self.ln_address.as_ref() {
+            if lnurl.lightning_address().as_ref() == Some(ln_address) {
+                return true;
+            }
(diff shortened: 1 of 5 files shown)


mutiny-core/src/labels.rs
```diff
@@ -131,6 +146,16 @@ pub trait LabelStorage {
     fn edit_contact(&self, id: impl AsRef<str>, contact: Contact) -> Result<(), MutinyError>;
     /// Gets all the existing tags (labels and contacts)
     fn get_tag_items(&self) -> Result<Vec<TagItem>, MutinyError>;
+    /// Finds a contact that has the given lnurl as either a lnurl or a lightning address
+    fn get_contact_for_lnurl(&self, lnurl: &LnUrl) -> Result<Option<String>, MutinyError> {
+        let contacts = self.get_contacts()?;
+        for (id, contact) in contacts {
```
TonyGiorgio, line 152: And if there's hundreds of contacts? That feels pretty inefficient. Probably needs a lookup or something.

benthecarman: We'd have to change how we store contacts. The payment will always be the longest part of this so figured the slight in efficiency here isn't the end of the world

TonyGiorgio (approved):

PR #871, opened 2023-11-22
Create contact for Mutiny+ subscription

benthecarman:
Makes it so Mutiny+ will have a nice contact icon in the activity. Just used the icon we have on our nostr profile for now. Not sure if @futurepaul has a better image we can use

image


TonyGiorgio (approved):
Slick!

PR #877, opened 2023-11-27
Generalize LSP config to an enum

benthecarman:
Pre requisite for LSPS


TonyGiorgio (approved):
Will this overwrite the LSP config on first load to translate it to the new enum version?

PR #889, opened 2023-12-05
Panic if lock has changed before reconnections

TonyGiorgio:
Still need to test it first, but pretty sure this is another edge case causing problems: 

Phone1: Connect Peer, Open Channel, minimizing app for 60s
Phone2: Opens, Pulls down state & channel, makes payment, closes app for 60s
Phone1: Reopens app, lock gets checked in reconnect loop, is allowed to proceed with previous state, connects to peer and force closes

It's just not safe to reconnect after lock id had changed. Longer term we can do a full storage refresh but that's a more complex procedure to bubble back up from the node reconnect logic.

### DIFF
--- a/mutiny-core/src/peermanager.rs
+++ b/mutiny-core/src/peermanager.rs
@@ -353,11 +353,19 @@ pub(crate) async fn connect_peer_if_necessary<S: MutinyStorage, P: PeerManager>(
         Ok(())
     } else {
         // make sure we have the device lock before connecting
-        // otherwise we could cause force closes
+        // otherwise we could cause force closes.
+        // If we didn't have the lock last, we need to panic because
+        // the state could have changed.
         if let Some(lock) = storage.fetch_device_lock().await? {
             let id = storage.get_device_id()?;
-            if lock.is_locked(&id) {
-                return Err(MutinyError::AlreadyRunning);
+            if !lock.is_last_locker(&id) {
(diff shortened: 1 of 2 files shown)


benthecarman (approved):
Makes sense to me!

TonyGiorgio:
I still need to test

TonyGiorgio:
Been thinking about this a little bit. One thing I think we might need instead is either a continuous lock checker that starts when the wallet starts up and then panic's if the lock has changed. Something maybe that checks every 3 seconds? Maybe it's too late at that point? 

Or we change VSS to be aware of a lock ID and if a state change was attempted without having the current lock then it fails the write?

benthecarman (approved):
Yeah this is definitely an improvement over the previous version.

We can probably do further restrictions in the future

benthecarman (approved):

TonyGiorgio:
Tested this a couple times on different browsers. 

First time it stopped it correctly. 
Screenshot 2023-12-28 at 11 17 03 PM

The second time it didn't catch it for some reason. It's possible it's due to weirdness with browser testing tolls to turn off network. It might not be disconnecting an active socket. At least the panic we put in LDK catches it. 
Screenshot 2023-12-28 at 11 21 17 PM

Over the next few days I can finish the websocket lock feature I started a few weeks ago.

PR #890, opened 2023-12-05
Fix lock problems found in #885

benthecarman:
These are pulled out of #885

These refactors were needed to get mutiny to compile with `Send + Sync`. These likely are indicative of deeper underlying problems with holding a lock longer than we should so pulling those out to get in now.


TonyGiorgio:
We've already got this now, right? Or does it need a rebase?

benthecarman:
Oh yeah this can be closed now

PR #891, opened 2023-12-06
Update secp256k1-sys in Cargo.lock

benthecarman:
old version was causing build issues in my IDE and made me have to use nix (see https://github.com/rust-bitcoin/rust-secp256k1/issues/667), this update fixes.

Also pinned the nightly version for the CI that was causing the panic


.github/workflows/test.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Install nightly toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly
+          toolchain: nightly-2023-10-24
```
TonyGiorgio, line 27: Pretty sure there's way more places you need to set this if you're going to set it at all. For instance, it should be kept in lockstep with nix too.

TonyGiorgio (changes requested):

.github/workflows/test.yml
benthecarman: yeah i am just trying to get this job working

benthecarman:
@TonyGiorgio got it working now

TonyGiorgio (approved):

PR #894, opened 2023-12-08
Add LSPS support via `lightning-liquidity` crate

johncantrell97:
Adds support for both Voltage and LSPS style LSP's.

If lsp_url configuration option is set, it adds voltage style lsp's as it does today.

If lsp_connection_string option is set, it adds lsps style lsp's. If the lsps lsp requires a token then lsp_token must be set and the index of the token must match the index of the corresponding lsp_connection_string. It's a little janky. One idea could be to incorporate the token into the connection_string... something like pubkey:token@ip:port and let it be optional when parsing. This isn't standardized anywhere but might be a nice way to do it? Though it does go against the typical connection string someone might copy and paste from other software for their LSP (amboss/mempool.space/etc). though maybe if someone's willing to go add a custom LSP they can handle this whereas most users will just use one of the predefined one's you guys populate in a list somewhere. I'll let you guys make final call on how to handle this.

The compute_opening_fee fn should be exported from the lightning-liquidity crate soon enough so we can remove the copy-pasted version here.

An alpha release of the crate is also expected to land shortly, not sure if we want to wait for that before merging something like this or if you guys are fine with keeping a dependency on a specific git commit for now.

### DIFF
--- a/mutiny-core/Cargo.toml
+++ b/mutiny-core/Cargo.toml
@@ -32,6 +32,7 @@ lightning-invoice = { version = "0.26.0", features = ["serde"] }
 lightning-rapid-gossip-sync = { version = "0.0.118" }
 lightning-background-processor = { version = "0.0.118", features = ["futures"] }
 lightning-transaction-sync = { version = "0.0.118", features = ["esplora-async-https"] }
+lightning-liquidity = { git = "https://github.com/lightningdevkit/lightning-liquidity.git", rev = "e00d917a8bb17e29493497538c7a4dda00bef151" }
 chrono = "0.4.22"
 futures-util = { version = "0.3", default-features = false }
 reqwest = { version = "0.11", default-features = false, features = ["json"] }
--- a/mutiny-core/src/lsp/lsps.rs
+++ b/mutiny-core/src/lsp/lsps.rs
@@ -1,0 +1,502 @@
+use async_trait::async_trait;
+use bitcoin::hashes::{sha256, Hash};
+use bitcoin::secp256k1::{PublicKey, Secp256k1};
+use bitcoin::Network;
+use futures::channel::oneshot;
+use lightning::ln::channelmanager::MIN_FINAL_CLTV_EXPIRY_DELTA;
+use lightning::ln::PaymentHash;
+use lightning::routing::gossip::RoutingFees;
+use lightning::routing::router::{RouteHint, RouteHintHop};
+use lightning::util::logger::Logger;
+use lightning::{log_debug, log_error};
+use lightning_invoice::{Bolt11Invoice, InvoiceBuilder};
+use lightning_liquidity::events;
+use lightning_liquidity::lsps2::{LSPS2Event, OpeningFeeParams};
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc, Mutex};
+
+use crate::{
+    error::MutinyError,
+    keymanager::PhantomKeysManager,
+    ldkstorage::PhantomChannelManager,
+    logging::MutinyLogger,
+    lsp::{FeeRequest, InvoiceRequest, Lsp, LspConfig},
+    node::{parse_peer_info, LiquidityManager},
+    storage::MutinyStorage,
+    utils,
+};
+
+use super::FeeResponse;
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
+pub struct LspsConfig {
+    pub connection_string: String,
+    pub token: Option<String>,
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct JitChannelInfo {
+    pub channel_id: u128,
+    pub fee_params: OpeningFeeParams,
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct GetInfoResponse {
+    pub jit_channel_id: u128,
+    pub opening_fee_params_menu: Vec<OpeningFeeParams>,
+}
+
+pub(crate) struct PendingPaymentInfo {
+    pub expected_fee_msat: Option<u64>,
+    pub fee_params: OpeningFeeParams,
+}
+
+type PendingFeeRequestSender = oneshot::Sender<Result<GetInfoResponse, MutinyError>>;
+type PendingBuyRequestSender = oneshot::Sender<Result<Bolt11Invoice, MutinyError>>;
+
+#[derive(Clone)]
+pub struct LspsClient<S: MutinyStorage> {
+    pub pubkey: PublicKey,
+    pub connection_string: String,
+    pub token: Option<String>,
+    liquidity_manager: Arc<LiquidityManager<S>>,
+    channel_manager: Arc<PhantomChannelManager<S>>,
+    keys_manager: Arc<PhantomKeysManager<S>>,
+    network: Network,
+    logger: Arc<MutinyLogger>,
+    pending_fee_requests: Arc<Mutex<HashMap<u128, PendingFeeRequestSender>>>,
+    pending_buy_requests: Arc<Mutex<HashMap<u128, PendingBuyRequestSender>>>,
+    pending_channel_info: Arc<Mutex<HashMap<u128, JitChannelInfo>>>,
+    pending_payments: Arc<Mutex<HashMap<PaymentHash, PendingPaymentInfo>>>,
+    stop: Arc<AtomicBool>,
+}
+
+impl<S: MutinyStorage> LspsClient<S> {
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn new(
+        lsp_connection_string: String,
+        token: Option<String>,
+        liquidity_manager: Arc<LiquidityManager<S>>,
+        channel_manager: Arc<PhantomChannelManager<S>>,
+        keys_manager: Arc<PhantomKeysManager<S>>,
+        network: Network,
+        logger: Arc<MutinyLogger>,
+        stop: Arc<AtomicBool>,
+    ) -> Result<Self, MutinyError> {
+        let (lsp_pubkey, _) = parse_peer_info(&lsp_connection_string)?;
+
+        let client = LspsClient {
+            pubkey: lsp_pubkey,
+            connection_string: lsp_connection_string,
+            token,
+            liquidity_manager,
+            channel_manager,
+            keys_manager,
+            network,
+            logger,
+            pending_fee_requests: Arc::new(Mutex::new(HashMap::new())),
+            pending_buy_requests: Arc::new(Mutex::new(HashMap::new())),
+            pending_channel_info: Arc::new(Mutex::new(HashMap::new())),
+            pending_payments: Arc::new(Mutex::new(HashMap::new())),
+            stop,
+        };
+
+        let events_client = client.clone();
+        utils::spawn(async move {
+            events_client.handle_events().await;
+        });
+
+        Ok(client)
+    }
+
+    pub(crate) async fn handle_event(&self, event: events::Event) {
+        match event {
+            events::Event::LSPS2(LSPS2Event::GetInfoResponse {
+                jit_channel_id,
+                opening_fee_params_menu,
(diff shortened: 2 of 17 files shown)


mutiny-core/src/nodemanager.rs
```diff
@@ -637,31 +614,34 @@ impl<S: MutinyStorage> NodeManager<S> {
...
-            }
+            Some(lsp_urls) if !lsp_urls.is_empty() && !c.safe_mode => lsp_urls
+                .split(',')
+                .map(|s| LspConfig::new_voltage_flow(s.trim().to_string()))
+                .collect(),
             _ => Vec::new(),
         };
 
+        // load lsps flow configs, if any
+        if let Some(lsp_connection_string) = c.lsp_connection_string.clone() {
```
TonyGiorgio, line 629: I wonder if we should remove the "multi lsp at the same time" option. It's really not practical to be interfacing with two at once. @benthecarman ?

TonyGiorgio (commented):
First pass looks good! Mostly suggestions around parameters and the differences between voltage & lsps. Still want to do a bunch of testing though. Is there a signet lsps currently?

mutiny-core/src/node.rs
```diff
@@ -321,27 +334,55 @@ impl<S: MutinyStorage> Node<S> {
...
+                }),
+            }),
+            channel_manager.clone(),
+            None,
+            None,
+        ));
+
+        let mutiny_message_handler = Arc::new(MutinyMessageHandler {
+            liquidity: liquidity_manager.clone(),
+        });
```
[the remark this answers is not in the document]
johncantrell97: yeah good point, will fix

mutiny-core/src/nodemanager.rs
benthecarman: ack from me

mutiny-core/src/node.rs
```diff
@@ -801,92 +839,135 @@ impl<S: MutinyStorage> Node<S> {
...
-                .sum();
-            let min_amount_sat = if inbound_capacity_msat > amount_sat * 1_000 {
-                1
-            } else {
-                utils::min_lightning_amount(self.network)
-            };
-            if amount_sat < min_amount_sat {
-                return Err(MutinyError::BadAmountError);
-            }
+        let user_channel_id: u128 = utils::now().as_secs().into();
```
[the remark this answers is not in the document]
johncantrell97: yeah it's just ignored in voltage case. I can make all non-shared parameters Options and add comments.  Well LSPS uses it to track the whole flow from initial fee negotiation through channel open.

mutiny-core/src/node.rs
```diff
@@ -801,92 +839,135 @@ impl<S: MutinyStorage> Node<S> {
...
+                                }
+                            };
+                            Ok(lsp_invoice)
+                        }
+                    }
+                }
+            }
+            None => Ok(self
+                .create_internal_invoice(amount_sat, None, labels, route_hints)
+                .await?),
```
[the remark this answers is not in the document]
johncantrell97: Yeah I restructured it a bit because it was getting too confusing for me to leave it how it was.  Basically you can think of the flow like 

-- is there an LSP configured?
    --No, then create normal invoice
    --Yes,
         --- do shared setup that all LSPs need
         --- match on type of LSP and do custom logic each one needs

mutiny-wasm/src/lib.rs
```diff
@@ -87,6 +87,8 @@ impl MutinyWallet {
         user_esplora_url: Option<String>,
         user_rgs_url: Option<String>,
         lsp_url: Option<String>,
+        lsp_connection_string: Option<String>,
+        lsp_token: Option<String>,
```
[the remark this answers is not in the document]
johncantrell97: well you can check the mutiny-web `lsps` branch to see how I did it for now.

right now the logic is:

if `lsp_url` is set then it's `voltage` flow.
if `lsp_connection_string` is set then it's `lsps` flow.

But yeah, it's _kind_ of awkward this way. There's no spec here so we're kind of making it up as we go.  I think as for how it should look in an ideal state for an advanced user who wants to switch LSPs is to support a dropdown that has mutiny pre-configured lsps (voltage, c=, whoever) and the last option is "Custom LSPS LSP". When that last option is selected it exposes two textboxes for providing connection string and token.

You could come up with some url scheme to combine them into one but it's not some standard. Could have pubkey:token@ip:port as a standard way to combine it, using the token almost like a basic auth password and pubkey as the username.

I'm not familiar with what the voltage style LSP_URL is but I imagine in theory you could combine these schemes. I don't think a user will ever need to enter a manual alternative voltage flow LSP though since that won't really exist.

We can talk about this on the call later today

mutiny-core/Cargo.toml
```diff
@@ -30,6 +30,9 @@ esplora-client = { version = "0.5", default-features = false }
 lightning = { version = "0.0.118", default-features = false, features = ["max_level_trace", "grind_signatures", "std"] }
 lightning-invoice = { version = "0.26.0", features = ["serde"] }
 lightning-rapid-gossip-sync = { version = "0.0.118" }
+lightning-background-processor = { version = "0.0.118", features = ["futures"] }
+lightning-transaction-sync = { version = "0.0.118", features = ["esplora-async-https"] }
+lightning-liquidity = { git = "https://github.com/lightningdevkit/lightning-liquidity.git", rev = "e00d917a8bb17e29493497538c7a4dda00bef151" }
```
TonyGiorgio, line 35: note to me: take a look at this repo

mutiny-core/src/lsp/lsps.rs
```diff
@@ -0,0 +1,391 @@
...
+            events_client.handle_events().await;
+        });
+
+        Ok(client)
+    }
+
+    pub(crate) async fn handle_events(&self) {
+        loop {
+            for event in self.liquidity_manager.get_and_clear_pending_events() {
+                match event {
```
benthecarman, line 110: Would be nice to move this event handling into a separate function that returns `Result` so we don't need all these `unwraps` and can properly handle the errors

mutiny-core/src/lsp/lsps.rs
```diff
@@ -0,0 +1,391 @@
...
+                                    base_msat: 0,
+                                    proportional_millionths: 0,
+                                },
+                                cltv_expiry_delta: cltv_expiry_delta.try_into().unwrap(),
+                                htlc_minimum_msat: None,
+                                htlc_maximum_msat: None,
+                            }]);
+
+                            let mut invoice = InvoiceBuilder::new(self.network.into())
+                                .description("Coins pls!".into())
```
benthecarman, line 164: make empty string

mutiny-core/src/lsp/lsps.rs
```diff
@@ -0,0 +1,391 @@
...
+                        counterparty_node_id,
+                        payment_size_msat,
+                        ..
+                    }) => {
+                        let mut pending_buy_requests = self.pending_buy_requests.lock().unwrap();
+
+                        if let Some(buy_response_sender) =
+                            pending_buy_requests.remove(&user_channel_id)
+                        {
+                            let invoice_expiry_delta_secs = 60 * 60 * 24 * 3;
```
benthecarman, line 145: We use `3600` in other parts of the code, should just do the same here

mutiny-core/src/lsp/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+#[derive(Serialize, Deserialize)]
+pub struct FeeRequest {
+    pub pubkey: String,
+    pub amount_msat: u64,
+    pub user_channel_id: u128,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct FeeResponse {
+    pub id: String,
```
benthecarman, line 73: Probably should make this `Option<String>` but not the end of the world

mutiny-core/src/node.rs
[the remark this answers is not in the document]
benthecarman: We would need to dynmically change our type above, might not be easily doable

mutiny-core/src/lsp/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+use std::sync::Arc;
+use voltage::LspClient;
+
+pub mod lsps;
+pub mod voltage;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub enum LspConfig {
+    VoltageFlow(String),
+    LspsFlow(LspsConfig),
```
benthecarman, line 23: I named it `VoltageFlow` because Flow is the name of the product, I think the LSPS would can just be named `LSPS`

benthecarman (changes requested):

mutiny-core/src/node.rs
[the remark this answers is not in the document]
johncantrell97: oh right, also a good point. hm. could probably make `liquidity` be Option<LiquidityManager> and ignore messages if it's None or something in the MutinyMessageHandler

mutiny-core/src/lsp/lsps.rs
```diff
@@ -0,0 +1,391 @@
...
+                                .description("Coins pls!".into())
+                                .payment_hash(sha256::Hash::from_slice(&payment_hash.0).unwrap())
+                                .payment_secret(payment_secret)
+                                .duration_since_epoch(utils::now())
+                                .min_final_cltv_expiry_delta(MIN_FINAL_CLTV_EXPIRY_DELTA.into())
+                                .private_route(lsp_route_hint);
+
+                            if let Some(payment_size_msat) = payment_size_msat {
+                                invoice = invoice.amount_milli_satoshis(payment_size_msat)
+                            };
```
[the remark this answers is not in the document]
johncantrell97: yeah the lib returns an error to the LSP (https://github.com/lightningdevkit/lightning-liquidity/blob/d89690914577a8e820c63fe0510b51107c890197/src/lsps2/channel_manager.rs#L270C7-L270C81)

and the LSP is supposed to fail the intercepted htlc back.

yeah can add a check here I guess but it's actually impossible for this to be `None` at the moment. the mutiny create_invoice flow enforces an amount:

```
// LSP requires an amount:
let amount_sat = amount_sat.ok_or(MutinyError::BadAmountError)?;
```

TonyGiorgio:
We've merged the send + sync change so this can be rebased

mutiny-core/src/lsp/lsps.rs
johncantrell97: did this but doesn't need to return Result itself because really it's sending the Result over the channel to be handled. there's nothing to be done here other than log.

mutiny-core/src/lsp/lsps.rs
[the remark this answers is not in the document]
johncantrell97: i ended up just forcing this to be Some here and erroring otherwise

mutiny-core/src/lsp/mod.rs
johncantrell97: oh whoops, I'll change it

johncantrell97:
ok, since @benthecarman ack'd it I went ahead and added a commit that removes multi-lsp support. i wasn't entirely sure about the NodeIndex related changes but it's probably not even necessary anymore? Though maybe it is 🤷 , I think the logic I left is equivalent to what was there.  In the case the saved lsp config does not match the provided config it returns None, which is what would have happened if the saved config was not in any of the provided configs before.

TonyGiorgio:
> i wasn't entirely sure about the NodeIndex related changes but it's probably not even necessary anymore

You mean how the LSP info is saved at the `NodeIndex` level? I believe this is still going to be required even if we don't have the ability to have multiple LSPs at the same time. We need to know when a new user restores their wallet which LSP they were connected to and not try to assign defaults there.

benthecarman (approved):

TonyGiorgio:
@futurepaul brings up a good point here: https://github.com/MutinyWallet/mutiny-web/pull/767

We store the connected LSP with the node in our storage. So we'll probably have to expose some sort of "switch LSP" APIs or something to handle this but that can be done in a follow up. Do you think I have assumption correct @benthecarman ?

PR #898, opened 2023-12-13
Rework wasm apis for Contacts

benthecarman:
Requested by @futurepaul 

Removed the `Contact` struct and just using the `TagItem` struct instead, they contain the same info besides `TagItem` has the `id`, which was the main problem with `Contact`

### DIFF
--- a/mutiny-wasm/src/lib.rs
+++ b/mutiny-wasm/src/lib.rs
@@ -26,16 +26,18 @@ use futures::lock::Mutex;
 use gloo_utils::format::JsValueSerdeExt;
 use lightning::routing::gossip::NodeId;
 use lightning_invoice::Bolt11Invoice;
+use lnurl::lightning_address::LightningAddress;
 use lnurl::lnurl::LnUrl;
 use mutiny_core::auth::MutinyAuthClient;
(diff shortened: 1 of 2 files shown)


futurepaul (changes requested):
looks like you hardcoded image_url to None, need to add that to TagItem

TonyGiorgio:
Code looks good. Is this going to work with existing tags / contacts though? Should double check a before and after.

benthecarman:
> Code looks good. Is this going to work with existing tags / contacts though? Should double check a before and after.

it should, this is just the api layer on top of the internal logic. Lets wait for paul to finish building on top before we merge though

benthecarman:
> looks like you hardcoded image_url to None, need to add that to TagItem

fixed

futurepaul (approved):
worked for me!

futurepaul:
can we merge this and cut an npm release? want to start using this for the new send stuff. (this allows me to look up a single contact by id)

benthecarman:
> can we merge this and cut an npm release? want to start using this for the new send stuff. (this allows me to look up a single contact by id)

sgtm maybe get in #904 as well so we can have the multi lsp stuff working too

PR #901, opened 2023-12-16
Update to LDK 119, bdk alpha2, bitcoin 0.30

benthecarman:
Updates all of our major deps

this will effect fedimint and DLCs, we'll need to update the deps in those as well.

Looks like bdk changes aren't backwards compatible so we'll need to wipe that and do a full resync


TonyGiorgio:
Not upgrading to 119

PR #902, opened 2023-12-18
Move log and stop outside of nodemanager

TonyGiorgio:


benthecarman (approved):

DanGould:
Why does this instantiated logger exist, and why use ldk's logger? I've had no trouble with env_logger and the `log` crate I'm used to in bitmask, another wasm web wallet based on bdk & ldk.

TonyGiorgio:
> Why does this instantiated logger exist, and why use ldk's logger? I've had no trouble with env_logger and the `log` crate I'm used to in bitmask, another wasm web wallet based on bdk & ldk.

We created our own logger called `MutinyLogger` to occasionally write to disk. We have to implement LDK's logger anyways so we just use the same struct for that.

PR #888, opened 2023-12-01
DLCs

benthecarman:
Opening for now so I can better document changes I need to make

Currently can open and close DLCs with my dummy function `test_send_dlc_offer`.

Has some current problems

- [ ] Privacy leak same keys used across all DLCs (need changes to `rust_dlc`)
- [ ] Same nostr key for DLCs
- [ ] Need to do proper coin selection, not just use them all
- [ ] Currently using damus relay with random kind, need to do this properly

Some other todos for usability

- [ ] Make functions for decoding oracle announcements and attestations
- [ ] Better wasm model

Longer term todos

- [ ] nostr market place
- [ ] local social tagging
- [ ] dlc "zaps"

Issue #895, opened 2023-12-10
Improve NWC support (return response from send payment, get_balance, make_invoice/lookup_invoice)


rolznz:
Hi, we added a mutiny wallet connector to Bitcoin Connect so you can connect to your wallet in any lightning webapp and seamlessly execute lightning functionality, as long as the wallet is open on a second device (e.g. keep the Mutiny wallet open on your PC and then use Bitcoin Connect to connect to it on your phone).

Unfortunately there are some issues: 
- Send Payment works, but it seems Mutiny Wallet never sends a response event, or at least not within 60 seconds. I had the wallet open on my phone and it marked the payment as sent within 20 seconds or so.
- Get Balance and Make invoice just don't seem to be supported at all?

I created an issue here: https://github.com/getAlby/bitcoin-connect/issues/150

benthecarman:
The response event is a problem with our relay. 

What is the filter you're using to try and get it?

rolznz:
@benthecarman we use this code:

```js
const event = await this.signEvent(unsignedEvent);
const sub = this.relay.sub([
  {
    kinds: [23195],
    authors: [this.walletServicePubkey], 
    "#e": [event.id],
  },
]);
```

benthecarman:
Ah okay, looks like we were only allowing filters that had the `authors` and `#p` defined. Have a change ready to fix that: https://github.com/MutinyWallet/blastr/pull/31

benthecarman:
> Get Balance and Make invoice just don't seem to be supported at all?

We haven't done `get_balance` because of the privacy concerns. I think we can add `make_invoice` without any issues though

TonyGiorgio:
Is there any `info` or `capabilities` feature of NWC? Now that the new NWC extensions are going to add a ton more API calls, some we may or may not implement. I wonder what's the best way to indicate who supports what.

rolznz:
@TonyGiorgio good point, it would be nice to add a `methods` property in the get_info RPC call at https://github.com/nostr-protocol/nips/pull/685/files

That is currently how WebLN does it: https://www.webln.guide/building-lightning-apps/webln-reference/webln.getinfo - I think this makes sense because it can be per-connection (not per-wallet-service as per the NIP-47 info event)

I will add a suggestion there.

benthecarman:
There is the info event that you create when you create the NWC, but yeah adding to the `get_info` event makes sense

rolznz:
[...]

PR #905, opened 2023-12-21
Update rust-nostr to v0.26

benthecarman:
This is needed for #888 and figured it'd be better to pull out on its own


benthecarman:
Fixes a bug now too https://github.com/MutinyWallet/mutiny-node/issues/895#issuecomment-1870866052

TonyGiorgio (approved):

PR #907, opened 2023-12-21
Fixes for LSPS

benthecarman:
Fixes a couple issues.

1. Makes sure we await node storage being written, this way we are sure it is done when we change LSP and restart
2. Makes `get_lsp_invoice` so we aren't serializing and deserializing when we don't need to 
3. Fixes a bug where I put the wrong pubkey in the invoice
4. Fixes so we persist invoices created from LSPS so we can properly track it

### DIFF
--- a/mutiny-core/src/lsp/lsps.rs
+++ b/mutiny-core/src/lsp/lsps.rs
@@ -228,12 +228,13 @@ impl<S: MutinyStorage> LspsClient<S> {
                         }
                     };
 
+                    let secp = Secp256k1::new();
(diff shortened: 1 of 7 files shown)


TonyGiorgio (approved):

TonyGiorgio:
Was there anything else that needed verification here?

benthecarman:
Never got a payment to go through was getting rate limited.

TonyGiorgio:
Needs rebase

benthecarman:
Got this working now!

TonyGiorgio (approved):

PR #913, opened 2023-12-23
Remove redshift

TonyGiorgio:
Redshift will probably look completely different when we ship something that users can use more easily, let's go ahead and stop supporting this code.


benthecarman (approved):
Been thinking the same thing 
RIP

PR #915, opened 2023-12-25
Switch to builder methodology

TonyGiorgio:
TODO: 
- [x] MutinyWalletConfig
- [x] MutinyWallet
- [x] NodeManager
- [x] Node

Will probably get rid of MutinyWalletConfigBuilder once I make my way through it all and just use MutinyWalletBuilder instead. But baby steps.

This will make it a lot easier to do specific things with `MutinyWallet` but not init the whole thing.

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -46,7 +46,6 @@ pub use crate::gossip::{GOSSIP_SYNC_TIME_KEY, NETWORK_GRAPH_KEY, PROB_SCORER_KEY
 pub use crate::keymanager::generate_seed;
 pub use crate::ldkstorage::{CHANNEL_MANAGER_KEY, MONITORS_PREFIX_KEY};
 
-use crate::logging::LOGGING_KEY;
 use crate::nodemanager::{
     ChannelClosure, MutinyBip21RawMaterials, MutinyInvoice, TransactionDetails,
 };
@@ -63,6 +62,7 @@ use crate::{
     nodemanager::NodeBalance,
     sql::glue::GlueDB,
 };
+use crate::{logging::LOGGING_KEY, nodemanager::NodeManagerBuilder};
 use crate::{nodemanager::NodeManager, nostr::ProfileType};
 use crate::{nostr::NostrManager, utils::sleep};
 use ::nostr::key::XOnlyPublicKey;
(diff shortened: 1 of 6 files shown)


mutiny-core/src/node.rs
```diff
@@ -244,17 +365,11 @@ impl<S: MutinyStorage> Node<S> {
             .replace(chain_monitor.clone());
 
         // read channelmonitor state from disk
-        let channel_monitors = if empty_state {
```
TonyGiorgio, line 247: FYI: we've never set `empty_state` to `true` so I went ahead and deleted these references.

TonyGiorgio:
I won't touch `mutiny-wasm` for now since I don't know how that'll effect the frontend yet. Also, in follow ups I can work on removing the places where `config` is passed around, and instead build things with independent variables. For instance, `NodeManager` is partially built with `MutinyWalletConfig` and I want to remove that, but I think this PR is in a good place as is without needing to restructure that yet.

mutiny-core/src/lib.rs
```diff
@@ -247,54 +246,128 @@ pub struct MutinyWalletConfig {
...
+            do_not_connect_peers: self.do_not_connect_peers,
+            skip_device_lock: self.skip_device_lock,
+            safe_mode: self.safe_mode,
+            skip_hodl_invoices: self.skip_hodl_invoices,
+        }
     }
 }
 
+#[derive(Clone)]
+pub struct MutinyWalletConfig {
```
benthecarman, line 352: since we want people to use the builder this should be `pub(crate)`

mutiny-core/src/lib.rs
```diff
@@ -487,7 +544,29 @@ impl<S: MutinyStorage> MutinyWallet<S> {
 
         Ok(mw)
     }
+}
 
+#[derive(Clone)]
+/// MutinyWallet is the main entry point for the library.
+/// It contains the NodeManager, which is the main interface to manage the
+/// bitcoin and the lightning functionality.
+pub struct MutinyWallet<S: MutinyStorage> {
```
benthecarman, line 550: nit: can we have the comment on top of the `#[derive(..)]` this always bothered me

mutiny-core/src/node.rs
benthecarman: oh yeah i think that was from the SCB stuff

mutiny-core/src/node.rs
```diff
@@ -166,55 +166,176 @@ impl PubkeyConnectionInfo {
...
+    wallet: Option<Arc<OnChainWallet<S>>>,
+    esplora: Option<Arc<AsyncClient>>,
     #[cfg(target_arch = "wasm32")]
-    websocket_proxy_addr: String,
+    websocket_proxy_addr: Option<String>,
+
+    // optional
+    lsp_config: Option<LspConfig>,
+    logger: Option<Arc<MutinyLogger>>,
+    network: Network,
```
benthecarman, line 187: i think we should require network, can definitely see some tricky bugs coming because someone forgot to set the network somewhere

benthecarman (changes requested):
overall looks good, this will definitely clean up some code

just a couple nits

mutiny-core/src/lib.rs
TonyGiorgio: The builder returns `MutinyWalletConfig`, so if we have it behind `crate`, it would be unable to be used. Having the members non `pub` should eliminate the ability to init `MutinyWalletConfig` directly.

TonyGiorgio:
Handled and made sure network is required for all of the builders.

benthecarman (approved):
nice work!

PR #916, opened 2023-12-27
Increase on-chain stop gap

benthecarman:
We keep getting support requests for missing txs, this should help


TonyGiorgio (changes requested):
I don't think we should be scanning for 100 each time. IMO, only clicking the resync button should trigger this. And I'd like to see it go up to 150 because I know at least one person needed it to go that high.

TonyGiorgio (approved):

PR #918, opened 2023-12-29
Optional federations

TonyGiorgio:
The init and restore/delete flow should behave back to normal if no federation had been added. Feel more comfortable shipping the release after this.

The workaround to any safari federation user that tries to `restore_mnemonic` is to do `delete_all` first until we get a fix in place.

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -485,26 +485,27 @@ impl<S: MutinyStorage> MutinyWalletBuilder<S> {
             stop.clone(),
         )?);
 
-        // create gluedb storage
-        let glue_db = GlueDB::new(
-            #[cfg(target_arch = "wasm32")]
-            None,
-            logger.clone(),
-        )
-        .await?;
-
-        // create federation library
+        // create federation module if any exist
         let federation_storage = self.storage.get_federations()?;
-        let federations = if federation_storage.federations.len() > 0 {
-            create_federations(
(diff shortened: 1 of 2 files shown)


mutiny-core/src/lib.rs
```diff
@@ -1157,7 +1166,10 @@ impl<S: MutinyStorage> MutinyWallet<S> {
     /// Deletes all the storage and gluedb data
     pub async fn delete_all(&self) -> Result<(), MutinyError> {
         self.storage.delete_all().await?;
-        self.glue_db.delete_all().await
+        if let Some(glue_db) = self.glue_db.clone() {
```
benthecarman, line 1169: think you can just use `as_ref` instead of `clone` here

mutiny-core/src/lib.rs
```diff
@@ -1181,7 +1193,10 @@ impl<S: MutinyStorage> MutinyWallet<S> {
         storage.set_data(LOGGING_KEY.to_string(), logs, None)?;
 
         // Delete all of glue_db storage
-        glue_db.delete_all().await?;
+        // FIXME: on Safari this will clash with previous storage setup
+        if let Some(glue_db) = glue_db.clone() {
```
benthecarman, line 1197: same here

benthecarman (approved):
2 nits but lgtm

benthecarman (approved):

Issue #917, opened 2023-12-29
Refactor tests to run in many environments


TonyGiorgio:
I've started this with the GlueDB testing. Write the testing functions once and call it from both wasm32 and non-wasm32. 

Although additionally, since we want to move towards mutiny working in a worker as well, we should start running tests in browser workers too. 

To do this, we need to create seperate test.rs files, unfortunately. 

https://docs.rs/wasm-bindgen-test/latest/src/wasm_bindgen_test/lib.rs.html#38

PR #922, opened 2023-12-30
Add some web worker tests

benthecarman:
Does some of #917

Main thing I needed to do was add a random id to the gluesql tests so the browser and worker tests don't clash

### DIFF
--- a/mutiny-core/src/federation.rs
+++ b/mutiny-core/src/federation.rs
@@ -728,12 +728,32 @@ mod tests {
 
 #[cfg(test)]
 #[cfg(target_arch = "wasm32")]
-mod wasm_tests {
+mod browser_tests {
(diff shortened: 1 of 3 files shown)


TonyGiorgio (changes requested):
I suspect that this isn't actually working as intended. If you check the docs on using `run_in_*`, it specifies that you cannot get away with just using a different `mod`. When I was testing out the sql library that needed to run in a worker, it was failing when I seperated it into a different `mod` but in the same file as another that called `run_in_browser`. The `run_in_browser` took priority since it can only run once. Once I commented out that `mod`, it would run correctly in a worker. You can test out if this is working or not by calling `js_sys::global().dyn_into::<web_sys::WorkerGlobalScope>() ` to see if it errors or not.

benthecarman:
dang it, yeah

PR #924, opened 2024-01-02
Wrap fedimint around an async rwlock

TonyGiorgio:


benthecarman (approved):
lgtm, I presume this is needed because before we couldn't read at the same time but with this we can

TonyGiorgio:
> lgtm, I presume this is needed because before we couldn't read at the same time but with this we can

Correct, it was blocking the activity list while a fedimint payment was being made.

benthecarman:
There's probably other places in the codebase we can do this too, would be a good followup

PR #926, opened 2024-01-03
Use AddressIndex::LastUnused when sweeping spendable outputs

benthecarman:
We had to bump the stop gap for the on-chain wallet because people had large gaps, I suspect this is one of the causes. This should help reduce the number of on-chain addresses we generate that don't end up ever being used.

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -800,12 +800,10 @@ impl<S: MutinyStorage> NodeManager<S> {
     }
 
     /// Gets a new bitcoin address from the wallet.
-    /// Will generate a new address on every call.
-    ///
-    /// It is recommended to create a new address for every transaction.
+    /// Will generate the last unused address in our bdk wallet.
     pub fn get_new_address(&self, labels: Vec<String>) -> Result<Address, MutinyError> {
         if let Ok(mut wallet) = self.wallet.wallet.try_write() {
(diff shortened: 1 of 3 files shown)


TonyGiorgio:
Can we use that everywhere? 

I don't like that every invoice generates an on chain address that will probably never be used.

benthecarman:
Yeah good idea. Did it in most places. Still having it generating new ones for channels because those are one time only

mutiny-core/src/nodemanager.rs
```diff
@@ -800,12 +800,10 @@ impl<S: MutinyStorage> NodeManager<S> {
...
     /// Gets a new bitcoin address from the wallet.
-    /// Will generate a new address on every call.
-    ///
-    /// It is recommended to create a new address for every transaction.
+    /// Will generate the last unused address in our bdk wallet.
     pub fn get_new_address(&self, labels: Vec<String>) -> Result<Address, MutinyError> {
         if let Ok(mut wallet) = self.wallet.wallet.try_write() {
-            let address = wallet.get_address(AddressIndex::New).address;
+            let address = wallet.get_address(AddressIndex::LastUnused).address;
             self.set_address_labels(address.clone(), labels)?;
```
TonyGiorgio, line 807: Is it going to clash with previous labels? Or does it overwrite?

benthecarman: it overwrites

PR #927, opened 2024-01-03
Return last fedimint error if node manager fails

TonyGiorgio:
Eventually I think we can stop making assumptions about the node manager being there at all, but that refactor for another day.

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -741,6 +741,7 @@ impl<S: MutinyStorage> MutinyWallet<S> {
 
         // Try each federation first
         let federation_ids = self.list_federation_ids().await?;
+        let mut last_federation_error = None;
         for federation_id in federation_ids {
             if let Some(fedimint_client) = self.federations.read().await.get(&federation_id) {
                 // Check if the federation has enough balance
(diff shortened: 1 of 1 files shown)


mutiny-core/src/lib.rs
```diff
@@ -776,6 +779,22 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+                    // InsufficientBalance for the node manager might just mean there is
+                    // no lightning node balance at all, something that might be common for
+                    // federation only users. Attempt to take a federation error if that
+                    // was configured and InsufficientBalance is returned.
+                    MutinyError::InsufficientBalance => {
+                        if let Some(federation_error) = last_federation_error {
+                            federation_error
+                        } else {
+                            e
+                        }
```
benthecarman, line 793: this can be simplified as `last_federation_error.unwrap_or(e)`

mutiny-core/src/lib.rs
```diff
@@ -776,6 +779,22 @@ impl<S: MutinyStorage> MutinyWallet<S> {
         self.node_manager
             .pay_invoice(None, inv, amt_sats, labels)
```
benthecarman, line 780: should we just check the node manager balance first and see if we could even pay it?

TonyGiorgio: seems like an unnecessary call that'll happen anyways and produce the same exact error

benthecarman (approved):

mutiny-core/src/lib.rs
TonyGiorgio: Decided it was an okay idea to go ahead and do. It'll do the call mostly twice but whatever.

benthecarman (approved):

PR #931, opened 2024-01-03
Refactor things in node manager to be in wallet

TonyGiorgio:


benthecarman (approved):
looks simple enough

TonyGiorgio:
Found a couple other smaller areas where `node_manager` was being called. Fix them and tested both ln address and subscriptions, all of that works now.

PR #932, opened 2024-01-03
Properly return error for create_bip21

benthecarman:
Closes https://github.com/MutinyWallet/mutiny-web/issues/792

I think this is from  #867

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -830,10 +830,12 @@ impl<S: MutinyStorage> MutinyWallet<S> {
         let invoice = if self.safe_mode {
             None
         } else {
-            self.create_lightning_invoice(amount, labels.clone())
-                .await
-                .ok()
-                .and_then(|invoice| invoice.bolt11)
+            Some(
+                self.create_lightning_invoice(amount, labels.clone())
+                    .await?
+                    .bolt11
+                    .ok_or(MutinyError::InvoiceCreationFailed)?,
+            )
         };
 
         let Ok(address) = self.node_manager.get_new_address(labels.clone()) else {


mutiny-core/src/lib.rs
```diff
@@ -831,9 +831,8 @@ impl<S: MutinyStorage> MutinyWallet<S> {
             None
         } else {
             self.create_lightning_invoice(amount, labels.clone())
-                .await
-                .ok()
-                .and_then(|invoice| invoice.bolt11)
+                .await?
+                .bolt11
```
TonyGiorgio, line 835: this was the code before: 

```
        // If we are in safe mode, we don't create invoices
        let invoice = if self.safe_mode {
            None
        } else {
            let inv = self.create_invoice(amount, labels.clone()).await?;
            Some(inv.bolt11.ok_or(MutinyError::WalletOperationFailed)?)
        };
```

needs an `ok_or` and wrapped in `Some` since it is an option

TonyGiorgio (changes requested):

mutiny-core/src/lib.rs
benthecarman: Brought back, made the error `InvoiceCreationFailed` because that seemed more correct

TonyGiorgio: Well, is the frontend parsing the error and deciding which one to do?

benthecarman: So we just display it

TonyGiorgio: No, it does switch on the error type, but in this case not this specific error so it's fine. 

```js
        } catch (e) {
            console.error(e);
            if (e === "Satoshi amount is invalid") {
                setError(i18n.t("receive.error_under_min_lightning"));
            } else {
                setError(i18n.t("receive.error_creating_unified"));
            }
        }
```

TonyGiorgio (approved):

mutiny-core/src/lib.rs
benthecarman: That is for checking the min lightning amount, I thought we displayed this error in the toast message

TonyGiorgio: No, this code is based on the return of this API call. It parses the error we return here and it is not taking it raw.

PR #937, opened 2024-01-05
Call process events after a disconnect

benthecarman:
These fix the connection issues with the C= LSP for me


TonyGiorgio (commented):
Oh man, this is touching code that's very delicate and has never needed to be touched. Why is an empty message being sent?

benthecarman:
I am just copying what ldk does in their tokio crate. Seems like the issue is fixed with just the second commit if you'd be more comfortable with just that

https://github.com/lightningdevkit/rust-lightning/blob/15b7f667b05b09a25af8ab2791ea4f0752fcc2ab/lightning-net-tokio/src/lib.rs#L171-L249

TonyGiorgio:
> Seems like the issue is fixed with just the second commit

yeah this one makes more sense. I don't believe empty bytes should be passed in.

benthecarman:
Okay changed to just that fix

TonyGiorgio (approved):

PR #940, opened 2024-01-05
Handle payjoin errors according to BIP 78

DanGould:
"The receiver is allowed to return implementation specific errors which
may assist the sender to diagnose any issue.

However, it is important that error codes that are not well-known and 
that the message do not appear on the sender's software user interface.
Such error codes or messages could be used maliciously to phish a non-
technical user. Instead those errors or messages can only appear in
debug logs.

It is advised to hard code the description of the well known error
codes into the sender's software."

See:
https://github.com/bitcoin/bips/blob/master/bip-0078.mediawiki#receivers-well-known-errors

This commit displays templates based on a particular [`ResponseError`](https://docs.rs/payjoin/latest/payjoin/send/enum.ResponseError.html)
and debugs the rest, according to this specification.

result:
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2957,13 +2957,14 @@ checksum = "de3145af08024dea9fa9914f381a17b8fc6034dfb00f3a84013f7ff43f29ed4c"
 
 [[package]]
 name = "payjoin"
-version = "0.10.0"
+version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


benthecarman (approved):

benthecarman:
needs rebase

DanGould:
I'm not sure why these platform-specific errors are happening in CI after rebase since they don't seem to relate to my changes. @benthecarman any hints?

benthecarman:
Maybe just try rebasing? seems like a spurious failure

DanGould:
yep it was spurious

benthecarman (approved):

PR #942, opened 2024-01-05
Less aggressive peer reconnection

TonyGiorgio:
One problem is LSP peer connection thread spawns and then this spawns and it tries to reconnect to LSP twice immediately after starting up. This should allow for some time to actually do the handshake. I don't care about waiting up to 30s before starting reconnect logic because if a connection fails immeidately upon startup, there's probably other factors at play. 

I've also moved from 5 second reconnect minimums to 10s.

### DIFF
--- a/mutiny-core/src/node.rs
+++ b/mutiny-core/src/node.rs
@@ -1981,7 +1981,8 @@ async fn start_reconnection_handling<S: MutinyStorage>(
     let stop_copy = stop.clone();
     utils::spawn(async move {
         // Now try to connect to the client's LSP
-        if let Some(lsp) = lsp_client_copy {
+        // This is here in case the LSP client node info has not saved to storage yet
+        if let Some(lsp) = lsp_client_copy.as_ref() {
             let node_id = NodeId::from_pubkey(&lsp.get_lsp_pubkey());
 
             let connect_res = connect_peer_if_necessary(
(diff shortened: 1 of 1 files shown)


benthecarman:
Overall looks good. I think this might break for users who have channels to peers that aren't their LSP. It would only connect to them after 30 seconds. Maybe a better approach is to combine this thread with the LSP thread.

TonyGiorgio:
> Overall looks good. I think this might break for users who have channels to peers that aren't their LSP. It would only connect to them after 30 seconds. Maybe a better approach is to combine this thread with the LSP thread.

Good point. I don't know why the seperation is there.

TonyGiorgio:
I still need to test it but this does the following: 

1. tries to connect to LSP node
2. tries to connect to non-lsp nodes
3. then after 30s will try to reconnect to any nodes that might have been disconnected

mutiny-core/src/node.rs
```diff
@@ -2014,29 +2015,94 @@ async fn start_reconnection_handling<S: MutinyStorage>(
...
+                    node_pubkey.to_hex(),
+                );
+                return;
+            }
+            sleep(1_000).await;
+        }
+
         loop {
-            for _ in 0..5 {
+            for _ in 0..10 {
```
benthecarman, line 2095: should make this `0..INITIAL_RECONNECTION_DELAY`

mutiny-core/src/node.rs
```diff
@@ -1981,7 +1981,8 @@ async fn start_reconnection_handling<S: MutinyStorage>(
     let stop_copy = stop.clone();
     utils::spawn(async move {
         // Now try to connect to the client's LSP
-        if let Some(lsp) = lsp_client_copy {
+        // This is here in case the LSP client node info has not saved to storage yet
+        if let Some(lsp) = lsp_client_copy.clone() {
```
benthecarman, line 1985: nit: think you can use `as_ref()` here instead

benthecarman (changes requested):
looks like it has a clippy issue too

TonyGiorgio:
Tested with/without an LSP and with/without another peer connected and both the init process and reconnecting after I switch off/on the wifi seemed to work.

benthecarman (approved):

PR #944, opened 2024-01-05
Show gateway_fee in FederationIdentity

TonyGiorgio:


benthecarman (approved):

TonyGiorgio:
Will wait for a comment from @futurepaul , this is for him

futurepaul (commented):
this looks good to me data-wise, although I don't see any code in mutiny-wasm so not sure it will show up for the frontend?

also, if we could add a convenience method called `get_estimated_fee` and I can just put in an mount and get an `Option<u64>` of your best guess, I can display that as "estimated fee"

TonyGiorgio:
> although I don't see any code in mutiny-wasm so not sure it will show up for the frontend?

Good catch, I assumed it was pulling it for conversion but it's not.

benthecarman:
Done in #995

PR #949, opened 2024-01-09
Refactor fedimint outcome processing

TonyGiorgio:
Functionally not different but less error prone and more idiomatic

### DIFF
--- a/mutiny-core/src/federation.rs
+++ b/mutiny-core/src/federation.rs
@@ -569,33 +569,28 @@ async fn extract_invoice_from_entry(
     }
 }
 
-fn process_pay_state_internal(pay_state: InternalPayState) -> (HTLCStatus, Option<String>) {
-    let status: HTLCStatus = pay_state.clone().into();
-
-    let p = if let InternalPayState::Preimage(preimage) = pay_state {
+fn process_pay_state_internal(pay_state: InternalPayState, invoice: &mut MutinyInvoice) {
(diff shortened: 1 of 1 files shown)


mutiny-core/src/federation.rs
```diff
@@ -569,33 +569,28 @@ async fn extract_invoice_from_entry(
     }
 }
 
-fn process_pay_state_internal(pay_state: InternalPayState) -> (HTLCStatus, Option<String>) {
-    let status: HTLCStatus = pay_state.clone().into();
+fn process_pay_state_internal(pay_state: InternalPayState, invoice: &mut MutinyInvoice) {
+    invoice.status = pay_state.clone().into();
```
benthecarman, line 573: nit: if you do this after the `if let` you don't need to `clone`

mutiny-core/src/federation.rs
```diff
@@ -569,33 +569,28 @@ async fn extract_invoice_from_entry(
...
         None
     };
-
-    (status, p)
 }
 
-fn process_pay_state_ln(pay_state: LnPayState) -> (HTLCStatus, Option<String>) {
-    let status: HTLCStatus = pay_state.clone().into();
+fn process_pay_state_ln(pay_state: LnPayState, invoice: &mut MutinyInvoice) {
+    invoice.status = pay_state.clone().into();
```
benthecarman, line 583: same here

benthecarman (approved):

PR #950, opened 2024-01-09
Add comments to zaps & lnurl pay

benthecarman:


TonyGiorgio (changes requested):
I'd prefer if you added it after labels in order to have less of a breaking change on the frontend.

TonyGiorgio (approved):
Thanks. Now it doesn't matter if the frontend updates or not, it'll treat it as None.

TonyGiorgio:
what is the intended use case for this? changes look fine, I'm just curious what spurred this change?

benthecarman:
Right now we have the `What's it for?` text on the new send screen, but it only labels it locally. With this we could send it in a zap/lnurl pay as well

TonyGiorgio:
> Right now we have the `What's it for?` text on the new send screen, but it only labels it locally. With this we could send it in a zap/lnurl pay as well

Okay, that makes sense when differentiated between private & public zaps, so makes sense for some of the things we've discussed.

PR #952, opened 2024-01-09
Remove empty claimable balances check

benthecarman:
I think this isn't behaving how we expect and seems like it will be empty for channels when it is still viable. I don't think this optimization did much either way.


TonyGiorgio:
what determines when we stop watching it then? Because something should stop watching old channels.

benthecarman:
I think ldk will internally, this just makes sure we don't prematurely stop

TonyGiorgio:
Please confirm that in some way. There's plenty of things LDK _should_ do that we can't assume they actually do.

benthecarman:
this didn't fix @futurepaul's issue so just going to close

PR #953, opened 2024-01-09
Move channel config load after watch_channel

benthecarman:
We weren't following the docs for the channel manager where we shouldn't change the channel config until after we've called `watch_channel`. This should fix the `unreachable` @futurepaul was running into

https://docs.rs/lightning/latest/lightning/ln/channelmanager/struct.ChannelManagerReadArgs.html

### DIFF
--- a/mutiny-core/src/ldkstorage.rs
+++ b/mutiny-core/src/ldkstorage.rs
@@ -197,6 +197,7 @@ impl<S: MutinyStorage> MutinyNodePersister<S> {
     pub(crate) async fn read_channel_manager(
         &self,
         network: Network,
+        accept_underpaying_htlcs: bool,
         chain_monitor: Arc<ChainMonitor<S>>,
         mutiny_chain: Arc<MutinyChain<S>>,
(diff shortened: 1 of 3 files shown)


futurepaul (approved):
worked!

TonyGiorgio (changes requested):
I don't see why these would be triggering at all. There's some bug here then because we haven't changed the channel config in a very long time. I agree we should move the code, however something is wrong with this code if it is being triggered without a config change.

benthecarman:
The config changed with the LSPS changes

TonyGiorgio:
It has? What has it changed to? The accept under pay thing? I guess that makes sense.  Was this @futurepaul 's first time using the wallet since that update? 

How was this not caught in testing? I used an existing wallet when I tested.

benthecarman:
Added a commit to only set `accept_underpaying_htlcs` with LSPS

TonyGiorgio:
Okay makes sense. I think it happened with someone else last week too. Was having a random panic.

TonyGiorgio (approved):

PR #959, opened 2024-01-11
update to latest liquidity lib

johncantrell97:
- updates to latest lib
- needs to use fork to patch ldk back to 118 (same as c=)
- fixes a bug where we weren't calling peer manager process events, causing delays in message sending

### DIFF
--- a/mutiny-core/Cargo.toml
+++ b/mutiny-core/Cargo.toml
@@ -32,7 +32,7 @@ lightning-invoice = { version = "0.26.0", features = ["serde"] }
 lightning-rapid-gossip-sync = { version = "0.0.118" }
 lightning-background-processor = { version = "0.0.118", features = ["futures"] }
 lightning-transaction-sync = { version = "0.0.118", features = ["esplora-async-https"] }
-lightning-liquidity = { git = "https://github.com/lightningdevkit/lightning-liquidity.git", rev = "e00d917a8bb17e29493497538c7a4dda00bef151" }
+lightning-liquidity = { git = "https://github.com/johncantrell97/ldk-lsp-client.git", rev = "9e01757d20c04aa31c28de8c4ffab5442d547edc" }
 chrono = "0.4.22"
 futures-util = { version = "0.3", default-features = false }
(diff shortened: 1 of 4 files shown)


benthecarman:
Can't create an invoice when running this, did the token change again?

johncantrell97:
yeah it would need to be live on our node first, we haven't even merged it yet. there's an underlying protocol change in this that both nodes need to have

benthecarman:
Okay once it is deployed there and you add a note so I can test again

johncantrell97:
This should be deployed to our mutinynet node now

TonyGiorgio:
Okay, we just cut a release to mutiny-node (not mutiny-web) and don't plan on another one for awhile. Is this good to go ahead and merge and cut another release before we freeze mutiny-web for a little bit?

johncantrell97:
From my perspective it's good to go. I think ben wanted to test it end-to-end just to make sure it works though

benthecarman (approved):
works for me!

mutiny-core/Cargo.toml
```diff
@@ -32,7 +32,7 @@ lightning-invoice = { version = "0.26.0", features = ["serde"] }
 lightning-rapid-gossip-sync = { version = "0.0.118" }
 lightning-background-processor = { version = "0.0.118", features = ["futures"] }
 lightning-transaction-sync = { version = "0.0.118", features = ["esplora-async-https"] }
-lightning-liquidity = { git = "https://github.com/lightningdevkit/lightning-liquidity.git", rev = "e00d917a8bb17e29493497538c7a4dda00bef151" }
+lightning-liquidity = { git = "https://github.com/johncantrell97/ldk-lsp-client.git", rev = "9e01757d20c04aa31c28de8c4ffab5442d547edc" }
```
TonyGiorgio, line 35: Does this branch not need this commit? 

https://github.com/johncantrell97/ldk-lsp-client/compare/118-patch...lightningdevkit%3Alightning-liquidity%3Amain

johncantrell97: na, that's for LSPS1 which we aren't using.

PR #962, opened 2024-01-13
Change archive contact to delete contact

benthecarman:
Previous iteration just archived which made things ugly for the front end afterwards. This is cleaner and makes thing simpler for everyone


benthecarman:
This doesn't work well with fedimint labels, not sure what the status on storage for fedimint stuff is cc @TonyGiorgio

TonyGiorgio:
> This doesn't work well with fedimint labels, not sure what the status on storage for fedimint stuff is cc @TonyGiorgio

Not sure. I'll look into it more later.

benthecarman:
can revisit this now that fedimint has moved to indexed db

TonyGiorgio (approved):
I think this will work just fine with the fedimint stuff. Just rebase and test but code looks functional.

benthecarman:
Tested and this now works with fedimint payments

PR #965, opened 2024-01-14
Set invoice labels after fedimint payment

### LINKED ISSUE #963: Making a payment to a contact from fedimint balance doesn't update contact's `last_updated`

benthecarman:
Closes #963

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -823,7 +823,10 @@ impl<S: MutinyStorage> MutinyWallet<S> {
                         .pay_invoice(inv.clone(), labels.clone())
                         .await;
                     match payment_result {
-                        Ok(r) => return Ok(r),
+                        Ok(r) => {
+                            self.storage.set_invoice_labels(inv.clone(), labels)?;
+                            return Ok(r);
+                        }
                         Err(e) => match e {
                             MutinyError::PaymentTimeout => return Err(e),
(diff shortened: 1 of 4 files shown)


mutiny-core/src/lib.rs
```diff
@@ -816,7 +816,10 @@ impl<S: MutinyStorage> MutinyWallet<S> {
                         .pay_invoice(inv.clone(), labels.clone())
                         .await;
                     match payment_result {
-                        Ok(r) => return Ok(r),
+                        Ok(r) => {
+                            self.storage.set_invoice_labels(inv.clone(), labels)?;
```
TonyGiorgio, line 820: If it happens inside of `pay_invoice` for `node_manager` then it should happen inside `pay_invoice` for `fedimint_client`, or vice versa.

TonyGiorgio (changes requested):

mutiny-core/src/lib.rs
benthecarman: That's what I wanted but the fedimint client doesn't have an instance of storage.

TonyGiorgio: Okay then they can both happen outside for now, that's probably better anyways so they both behave the same.

benthecarman: Good call, did that for invoice creation as well. Cleaned up code a lot

PR #967, opened 2024-01-15
DM requests for invoices

benthecarman:
Think this is all the functionality needed for @futurepaul to start

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -69,7 +69,7 @@ use crate::{nostr::NostrManager, utils::sleep};
 use ::nostr::key::XOnlyPublicKey;
 use ::nostr::nips::nip57;
 use ::nostr::prelude::ZapRequestData;
-use ::nostr::{JsonUtil, Kind};
+use ::nostr::{Event, JsonUtil, Kind};
 use async_lock::RwLock;
 use bdk_chain::ConfirmationTime;
 use bip39::Mnemonic;
@@ -421,6 +421,31 @@ pub struct MutinyWalletConfigBuilder {
     skip_device_lock: bool,
     pub safe_mode: bool,
     skip_hodl_invoices: bool,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct DirectMessage {
+    pub from: XOnlyPublicKey,
+    pub to: XOnlyPublicKey,
+    pub message: String,
+    pub date: u64,
+}
+
+impl PartialOrd for DirectMessage {
+    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for DirectMessage {
+    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+        // order by date, then the message, the keys
+        self.date
+            .cmp(&other.date)
+            .then_with(|| self.message.cmp(&other.message))
+            .then_with(|| self.from.cmp(&other.from))
+            .then_with(|| self.to.cmp(&other.to))
+    }
 }
 
 impl MutinyWalletConfigBuilder {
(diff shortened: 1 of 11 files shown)


mutiny-core/src/lib.rs
```diff
@@ -629,7 +629,7 @@ impl<S: MutinyStorage> MutinyWalletBuilder<S> {
 pub struct MutinyWallet<S: MutinyStorage> {
     xprivkey: ExtendedPrivKey,
     config: MutinyWalletConfig,
-    pub(crate) storage: S,
+    pub storage: S,
```
TonyGiorgio, line 632: Pass storage to things that need storage.

TonyGiorgio (changes requested):

mutiny-core/src/storage.rs
```diff
@@ -268,6 +273,28 @@ pub trait MutinyStorage: Clone + Sized + Send + Sync + 'static {
         self.get_data(MNEMONIC_KEY)
     }
 
+    /// Insert a nsec into the storage
+    async fn insert_nsec(&self, nsec: SecretKey) -> Result<(), MutinyError> {
```
TonyGiorgio, line 277: Hard NACK. Nsec should not be pasted into a browser site.

TonyGiorgio (changes requested):

mutiny-core/src/storage.rs
benthecarman: Should discuss IRL but paul and I's thinking was we already doing worse with a bitcoin seed

TonyGiorgio: Huge difference in a mutiny seed being used for mutiny things and importing another seed/private key from somewhere else for other purposes. 

Find another way, there's others that exist. Maybe ask Pablo...

TonyGiorgio: There's a few things I think we can do here but I would absolutely not allow nsec pasting here. 

1. Nostr plugins for web browsers that already have them
2. Create an nsec from the derivation path and use that for users that don't already have nostr
  2a. Some delegated nostr stuff so you can use the mutiny derivation seed as if it were the user's main key (don't think this really exists yet)
3. nsec bunker for everyone else

benthecarman:
Changed to not import but be instead be part of the building phase. We can then instead store it somewhere more secure

TonyGiorgio:
Needs clippy fixes

mutiny-core/src/nostr/mod.rs
```diff
@@ -91,14 +106,17 @@ impl<S: MutinyStorage> NostrManager<S> {
             .map(|x| x.profile.relay.clone())
             .collect();
 
+        // add relay to pull DMs from
+        relays.push("wss://relay.primal.net".to_string());
```
TonyGiorgio, line 110: add damus

mutiny-core/src/storage.rs
```diff
@@ -27,6 +27,7 @@ use uuid::Uuid;
 
 pub const KEYCHAIN_STORE_KEY: &str = "bdk_keychain";
 pub const MNEMONIC_KEY: &str = "mnemonic";
+pub const NSEC_KEY: &str = "nsec";
```
TonyGiorgio, line 30: delete this

TonyGiorgio (changes requested):

mutiny-core/src/lib.rs
```diff
@@ -1420,6 +1462,64 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+    /// Get dm conversation between us and given npub
+    /// Returns a vector of messages sorted by newest first
+    pub async fn get_dm_conversation(
+        &self,
+        primal_url: Option<&str>,
+        npub: XOnlyPublicKey,
+        limit: u64,
+        until: Option<u64>,
+    ) -> Result<Vec<DirectMessage>, MutinyError> {
+        let url = primal_url.unwrap_or("https://primal-cache.mutinywallet.com/api");
```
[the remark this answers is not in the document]
benthecarman: Currently we have it an an ENV in the front end and pass it for the individual requests, probably could refactor to have this as part of init, maybe good for a follow up so we can do it with the sync nostr contacts stuff too

mutiny-core/src/nostr/mod.rs
```diff
@@ -1011,14 +1183,33 @@ impl<S: MutinyStorage> NostrManager<S> {
...
         stop: Arc<AtomicBool>,
     ) -> Result<Self, MutinyError> {
         let context = Secp256k1::new();
 
-        // generate the default primary key
-        let primary_key = Self::derive_nostr_key(&context, xprivkey, 0, None, None)?;
+        // use provided nsec, otherwise generate it from seed
+        let (primary_key, public_key) = match key_source {
+            NostrKeySource::Derived => {
+                let keys = Self::derive_nostr_key(&context, xprivkey, 0, None, None)?;
```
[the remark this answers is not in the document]
benthecarman: Yeah we're using `0` to comply with nip6, will make it an const

https://github.com/nostr-protocol/nips/blob/master/06.md

mutiny-core/src/lib.rs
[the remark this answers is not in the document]
TonyGiorgio: We're already changing the mutiny wallet init parameters, might as well add it in this PR

TonyGiorgio (approved):

benthecarman:
Was missing send functionality and an efficient way to get future messages. Added that. Cleaned up some of our nostr client handling as well so we don't need to make so many new connections

TonyGiorgio:
Tests are panicing

TonyGiorgio (approved):

Issue #947, opened 2024-01-08
Missing fedimint activity items


TonyGiorgio:
Not sure what triggers it, never get it locally/in signet but in prod I've been missing a lot of fedimint transactions, sometimes seconds after they happen.

benthecarman:
had a user report this happened with a nwc payment, maybe it happens when its labeled?

ekzyis:
I was able to reliably (not always but often enough) reproduce it by zapping people using NWC. It seems to only show the latest zap in most cases. See video:


https://github.com/MutinyWallet/mutiny-node/assets/27162016/624e4e4f-4944-4a67-85ea-650be953d63c

PR #970, opened 2024-01-18
Fedimint indexeddb

TonyGiorgio:
Fixes https://github.com/MutinyWallet/mutiny-node/issues/947


mutiny-core/src/storage.rs
```diff
@@ -653,6 +663,99 @@ impl MutinyStorage for () {
...
+    inbound: bool,
+) -> std::io::Result<()> {
+    let key = payment_key(inbound, payment_hash);
+    storage
+        .set_data(key, payment_info, None)
+        .map_err(std::io::Error::other)
+}
+
+pub(crate) fn get_payment_info<S: MutinyStorage>(
+    storage: S,
```
benthecarman, line 695: should make this `storage: &S` so we don't need to clone everywhere

mutiny-core/src/storage.rs
```diff
@@ -653,6 +663,99 @@ impl MutinyStorage for () {
...
+
+    // if no inbound check outbound
+    match read_payment_info(storage.clone(), payment_hash.as_inner(), false, logger) {
+        Some(payment_info) => Ok((payment_info, false)),
+        None => Err(MutinyError::NotFound),
+    }
+}
+
+pub(crate) fn read_payment_info<S: MutinyStorage>(
+    storage: S,
```
benthecarman, line 714: same here

mutiny-core/src/storage.rs
```diff
@@ -653,6 +663,99 @@ impl MutinyStorage for () {
...
+                map.into_values().next()
+            } else {
+                None
+            }
+        }
+    }
+}
+
+pub(crate) fn list_payment_info<S: MutinyStorage>(
+    storage: S,
```
benthecarman, line 735: same here

mutiny-core/src/storage.rs
```diff
@@ -653,6 +663,99 @@ impl MutinyStorage for () {
...
+    };
+    let map: HashMap<String, PaymentInfo> = storage.scan(prefix, None)?;
+
+    // convert keys to PaymentHash
+    Ok(map
+        .into_iter()
+        .map(|(key, value)| {
+            let payment_hash_str = key
+                .trim_start_matches(prefix)
+                .splitn(2, '_') // To support the old format that had `_{node_id}` at the end
```
benthecarman, line 750: does this handle if there is no `_` suffix?

mutiny-wasm/src/indexed_db.rs
```diff
@@ -296,7 +296,9 @@ impl IndexedDbStorage {
                 let results = futures::future::join_all(futs).await;
                 for result in results {
                     if let Some((key, value)) = result? {
-                        map.set_data(key, value, None)?;
+                        // save to memory and local storage so we don't have to pull again
+                        map.set_data(key.clone(), value.clone(), None)?;
+                        Self::save_to_indexed_db(indexed_db, &[(key, value)]).await?;
```
benthecarman, line 301: would be better to do these all in one batch rather than individual writes

mutiny-core/src/nodemanager.rs
```diff
@@ -157,6 +157,40 @@ impl From<Bolt11Invoice> for MutinyInvoice {
     }
 }
 
+impl From<MutinyInvoice> for PaymentInfo {
```
benthecarman, line 160: not a huge deal but should we move `MutinyInvoice` to be defined somewhere else now that node manager doesn't even list invoices

mutiny-core/src/lib.rs
```diff
@@ -1456,6 +1465,39 @@ impl<S: MutinyStorage> MutinyWallet<S> {
         Ok(FederationBalances { balances })
     }
 
+    /// Starts a background process that will watch for nostr wallet connect events
```
benthecarman, line 1468: comment is wrong

mutiny-core/src/federation.rs
```diff
@@ -238,15 +263,44 @@ impl FederationClient {
...
+                    // TODO labels
+                    let mutiny_invoice = MutinyInvoice::from(i.clone(), h, false, vec![]).ok();
+
+                    // filter out expired invoices
+                    mutiny_invoice.filter(|invoice| {
+                        !invoice.bolt11.as_ref().is_some_and(|b| b.would_expire(now))
+                            || matches!(i.status, HTLCStatus::Succeeded | HTLCStatus::InFlight)
+                    })
+                }),
+        );
```
benthecarman, line 299: this seems suspect to me

The `.filter` seems wrong to me, we don't care about successful payments because those are already completed. I think you'd want to get `InFlight` and `Pending` here, also don't want to filter out expired invoices because we should check if they've been paid in the mean time, if not then we should set to failed.

I assume this will effect the logic below

benthecarman (changes requested):

mutiny-core/src/storage.rs
TonyGiorgio: Yes, it's a max split of 2 so if there's none then it's just the whole string as the first vector. Always guaranteed to have 1 entry so the `[0]` is safe.

mutiny-core/src/federation.rs
TonyGiorgio: Good catch, I now just filter out `Succeeded` and `InFlight`. Testing it now but all the other code changes have been handled.

mutiny-core/src/storage.rs
```diff
@@ -653,6 +663,97 @@ impl MutinyStorage for () {
...
+        format!(
+            "{}{}",
+            PAYMENT_OUTBOUND_PREFIX_KEY,
+            payment_hash.to_hex().as_str()
+        )
+    }
+}
+
+pub(crate) fn persist_payment_info<S: MutinyStorage>(
+    storage: S,
```
benthecarman, line 683: can be ref here too

mutiny-core/src/lib.rs
```diff
@@ -1458,6 +1616,39 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+                                Some(fedimint_client) => {
+                                    let _ = fedimint_client.check_activity().await.map_err(|e| {
+                                        log_error!(logger, "error checking activity: {e}")
+                                    });
+                                }
+                                None => todo!(),
+                            }
+                        }
+                    }
+                    Err(_) => todo!(),
```
TonyGiorgio, line 1646: forgot these

benthecarman (approved):
lgtm

PR #980, opened 2024-01-22
Add util function for hodl invoice detection

benthecarman:
No reason to keep the disction between test and non test for the list, just keeping the `ONE_KEY` in there because no one will ever have it anyways

### DIFF
--- a/mutiny-wasm/src/lib.rs
+++ b/mutiny-wasm/src/lib.rs
@@ -1705,6 +1705,14 @@ impl MutinyWallet {
         let npub = parse_npub_or_nip05(&npub).await?;
         Ok(npub.to_bech32().expect("bech32"))
     }
+
+    /// If the invoice is from a node that gives hodl invoices
+    #[wasm_bindgen]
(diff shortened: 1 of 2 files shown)


mutiny-wasm/src/lib.rs
```diff
@@ -1705,6 +1705,14 @@ impl MutinyWallet {
         let npub = parse_npub_or_nip05(&npub).await?;
         Ok(npub.to_bech32().expect("bech32"))
     }
+
+    /// If the invoice is from a node that gives hodl invoices
+    #[wasm_bindgen]
+    pub async fn is_potential_hodl_invoice(invoice: String) -> Result<bool, MutinyJsError> {
```
TonyGiorgio, line 1711: Why is this needed? Don't we already warn users when paying zeus that it is a hodl invoice? Something is already doing that logic somewhere based on the list of pubkeys.

benthecarman: In the front end we only warn for lightning address by checking for @zuespay.com

TonyGiorgio (approved):

PR #981, opened 2024-01-23
Timestamp NWC filters

benthecarman:
This is needed to do NWC on relays that aren't on blastr. Otherwise we may pull down events multiple times and keep reprocessing them

### DIFF
--- a/mutiny-core/src/nostr/mod.rs
+++ b/mutiny-core/src/nostr/mod.rs
@@ -132,14 +132,27 @@ impl<S: MutinyStorage> NostrManager<S> {
         relays
     }
 
-    fn get_nwc_filters(&self) -> Vec<Filter> {
-        self.nwc
+    fn get_nwc_filters(&self) -> Result<Vec<Filter>, MutinyError> {
+        // if we haven't synced before, use now and save to storage
+        let time_stamp = match self.storage.get_nwc_sync_time()? {
+            None => {
+                let now = Timestamp::now();
+                self.storage.set_nwc_sync_time(now.as_u64())?;
+                now
+            }
+            Some(time) => Timestamp::from(time),
+        };
+
+        let vec = self
+            .nwc
             .read()
             .unwrap()
             .iter()
             .filter(|x| x.profile.active())
-            .map(|nwc| nwc.create_nwc_filter())
-            .collect()
+            .map(|nwc| nwc.create_nwc_filter(time_stamp))
+            .collect();
+
+        Ok(vec)
     }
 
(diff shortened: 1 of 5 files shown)


mutiny-core/src/nostr/mod.rs
```diff
@@ -98,14 +98,27 @@ impl<S: MutinyStorage> NostrManager<S> {
         relays
     }
 
-    pub fn get_nwc_filters(&self) -> Vec<Filter> {
-        self.nwc
+    pub fn get_nwc_filters(&self) -> Result<Vec<Filter>, MutinyError> {
+        // if we haven't synced before, use now and save to storage
+        let time_stamp = match self.storage.get_nwc_sync_time()? {
```
TonyGiorgio, line 103: Wasn't dm sync time also added? Wouldn't that be the same there too?

benthecarman: That's in the other pr, not sure we'd want to have them use the same timestamp. We could process a dm but not a nwc and miss it.

TonyGiorgio: But NWCs are DM's? If we treat it all the same as sync all the DM's before marking the sync time, wouldn't that be the same?

TonyGiorgio: The worry is that we sync the DMs but not sync NWCs to completion and then set the sync time? So perhaps we wait for them all to sync before marking the DMs? Afterall, syncing the DMs is so we can get all of the payment requests anyways, there's functionally no difference in why we are syncing them.

benthecarman: It's more so because they exist on different relays and filters so the ordering isn't guaranteed to be the same for both. The Damus relays could be giving us dms from a week ago but we get an nwc from the mutiny relays from today, we'll process the nwc and have the potential to miss all the dms if we close before we finish processing

TonyGiorgio: This is a symptom of a bigger problem then and it's syncing events across relays across nostr keys. Your same argument applies to getting events from one NWC and processing that NWC but it still has other NWCs to go through.

benthecarman: Yeah the way we do it on blastr is best but we don't be able to do that with real relays

TonyGiorgio: I guess we need something, can try it out and see how it goes.

TonyGiorgio (approved):

PR #990, opened 2024-01-25
Reuse reqwest client across codebase

### LINKED ISSUE #973: Unify `reqwest` clients across codebase
We initialize a few of these across the code base and it'd be better if one shared one was used everywhere. Would make it so we open less connections and do less TLS requests

benthecarman:
Closes #973

If we create new reqwest clients everywhere we need to create new connections and redo TLS verification for every new client. This makes it so we reuse one across the codebase for pretty much everywhere. Should make things more efficient .

As part of this, moved the bitcoin price code to mutiny wallet instead of the node manager, made more sense here and made the refactor easier.

### DIFF
--- a/mutiny-core/src/auth.rs
+++ b/mutiny-core/src/auth.rs
@@ -24,7 +24,7 @@ struct CustomClaims {
 
 pub struct MutinyAuthClient {
     pub auth: AuthManager,
-    lnurl_client: Arc<LnUrlClient>,
+    pub lnurl_client: Arc<LnUrlClient>,
     url: String,
     http_client: Client,
     jwt: RwLock<Option<String>>,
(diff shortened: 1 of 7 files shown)


mutiny-core/src/auth.rs
```diff
@@ -24,7 +24,7 @@ struct CustomClaims {
 
 pub struct MutinyAuthClient {
     pub auth: AuthManager,
-    lnurl_client: Arc<LnUrlClient>,
+    pub lnurl_client: Arc<LnUrlClient>,
```
TonyGiorgio, line 27: nack

not only should you not be reusing an internal client to `MutinyAuth` for this, but you should be using an internal client to an internal lnurl client library for this either. either have some generic http client that everyone uses at the top level or let things be as is.

TonyGiorgio (changes requested):

benthecarman:
Turns out this doesn't matter for wasm, the browser handles it all for you. Will still leave the issue open in case someone wants to pick it up for mutiny-server or whatever

PR #991, opened 2024-01-25
Fix expected lsp amount when amount is 1

TonyGiorgio:


benthecarman (approved):
tested with 1, 10, 100, 1k, 10k, 100k, 1m, 10m

PR #992, opened 2024-01-25
Create nostr profile

benthecarman:
Allows you to create a nostr profile. Added a `sync_nostr` function that'll sync up their profile with primal, that way we can fetch changes made outside the app

### DIFF
--- a/mutiny-core/src/labels.rs
+++ b/mutiny-core/src/labels.rs
@@ -155,6 +155,20 @@ pub trait LabelStorage {
         }
         Ok(None)
     }
+    /// Finds a contact that has the given npub
+    fn get_contact_for_npub(
+        &self,
+        npub: XOnlyPublicKey,
+    ) -> Result<Option<(String, Contact)>, MutinyError> {
+        // todo this is not efficient, we should have a map of npub to contact
+        let contacts = self.get_contacts()?;
+        for (id, contact) in contacts {
+            if contact.npub == Some(npub) {
+                return Ok(Some((id, contact)));
+            }
+        }
+        Ok(None)
+    }
 }
 
 impl<S: MutinyStorage> LabelStorage for S {
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -1612,7 +1612,7 @@ impl<S: MutinyStorage> MutinyWallet<S> {
                 if profile.tag != NwcProfileTag::Subscription {
                     let mut nwc = profile.clone();
                     nwc.tag = NwcProfileTag::Subscription;
-                    self.nostr.edit_profile(nwc)?;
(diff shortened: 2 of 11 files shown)


mutiny-wasm/src/lib.rs
```diff
@@ -1597,6 +1597,50 @@ impl MutinyWallet {
...
     }
 
+    /// Returns the user's nostr profile data
+    #[wasm_bindgen]
+    pub fn get_nostr_profile(&self) -> Result<JsValue, MutinyJsError> {
+        let profile = self.inner.nostr.get_profile()?;
+        Ok(JsValue::from_serde(&profile)?)
+    }
+
+    /// Returns the user's nostr profile data
```
futurepaul, line 1607: dupe comment from above

TonyGiorgio:
needs rebase

futurepaul:
Can you rebase to master plz

TonyGiorgio:
Can you manually cut a release without merging here? Should be able to from the command line, might as well make it 0.6.0-rc1

mutiny-core/src/labels.rs
```diff
@@ -155,6 +155,20 @@ pub trait LabelStorage {
         }
         Ok(None)
     }
+    /// Finds a contact that has the given npub
+    fn get_contact_for_npub(
+        &self,
+        npub: XOnlyPublicKey,
+    ) -> Result<Option<(String, Contact)>, MutinyError> {
+        // todo this is not efficient, we should have a map of npub to contact
```
TonyGiorgio, line 163: Does this todo need to happen here?

mutiny-core/src/lib.rs
```diff
@@ -1637,6 +1639,42 @@ impl<S: MutinyStorage> MutinyWallet<S> {
         Ok(())
     }
 
+    /// Uploads a profile pic to nostr.build and returns the uploaded file's URL
+    pub async fn upload_profile_pic(&self, image_bytes: Vec<u8>) -> Result<String, MutinyError> {
+        let client = reqwest::Client::new();
+
+        let form = Form::new().part("fileToUpload", Part::bytes(image_bytes));
+        let res: NostrBuildResult = client
+            .post("https://nostr.build/api/v2/upload/profile")
```
TonyGiorgio, line 1648: Do we need to reach out to do this properly?

TonyGiorgio (approved):
Good to merge, just some questions if we want to handle them in follow ups or not.

mutiny-core/src/labels.rs
benthecarman: It's okay for now, not terribly inefficient, but we should fix before going to prod with this

mutiny-core/src/lib.rs
benthecarman: Talked to their team already and this is what they recommened that we do

PR #993, opened 2024-01-25
Use strfry and blastr for sending

ekzyis:


mutiny-core/src/nostr/mod.rs
```diff
@@ -122,7 +122,7 @@ impl<S: MutinyStorage> NostrManager<S> {
         relays.push("wss://relay.primal.net".to_string());
         relays.push("wss://relay.damus.io".to_string());
 
-        // add blastr for default sending
+        // add strfry for default sending
         relays.push("wss://relay.mutinywallet.com".to_string());
```
benthecarman, line 126: tbh we should add blastr back, can you do that

ekzyis: you mean to have both relays? yes, can do

ekzyis: you mean to have both relays? yes, can do

benthecarman (approved):

TonyGiorgio (approved):
awesome, thank you!

PR #995, opened 2024-01-26
Fedimint v0.2.2 Changes

### LINKED ISSUE #911: Max fedimint limit
Probably 1M sats

### LINKED ISSUE #912: Swap to lightning from fedimint
Probably similar to the swap to ln from on chain screen

### LINKED ISSUE #975: Scan for fedimint ecash
On restore, we should scan for fedimint ecash to make sure we have everything. Could make it a button like we do for resyncing on chain too.

### LINKED ISSUE #976: Go through vetted gateways first and then fall back to trying others

### LINKED ISSUE #988: Remove fedimint KV data when removing a federation

### LINKED ISSUE #997: Batched VSS remote writer
Based on the discussions here: https://github.com/fedimint/fedimint/discussions/4151

It sounds like doing some ~50ms batched writer for the fire and forget vss mode should be really helpful in minimizing the amount of writes that happen for the fedimint KV. Otherwise, even in fire and forget mode, these can be ~40 requests being clogged up in the browser.

TonyGiorgio:
Fixes #988
Fixes #975
Fixes #976
Fixes #997
Fixes #911
Fixes #912

### DIFF
--- a/mutiny-core/src/federation.rs
+++ b/mutiny-core/src/federation.rs
@@ -533,6 +533,17 @@ impl<S: MutinyStorage> FederationClient<S> {
     pub async fn delete_fedimint_storage(&self) -> Result<(), MutinyError> {
         self.fedimint_storage.delete_store().await
     }
+
+    pub async fn recover_backup(&self) -> Result<(), MutinyError> {
+        self.fedimint_client
+            .restore_from_backup(None)
+            .await
+            .map_err(|e| {
+                log_error!(self.logger, "Could not restore from backup: {}", e);
+                e
+            })?;
+        Ok(())
+    }
 }
 
 // A federation private key will be derived from
--- a/mutiny-core/src/federation.rs
+++ b/mutiny-core/src/federation.rs
@@ -159,6 +159,7 @@ pub(crate) struct FederationClient<S: MutinyStorage> {
     pub(crate) uuid: String,
     pub(crate) fedimint_client: ClientArc,
     storage: S,
+    #[allow(dead_code)]
     fedimint_storage: FedimintStorage<S>,
     pub(crate) logger: Arc<MutinyLogger>,
 }
@@ -570,6 +571,8 @@ impl<S: MutinyStorage> FederationClient<S> {
         }
     }
 
+    // delete_fedimint_storage is not suggested at the moment due to the lack of easy restores
+    #[allow(dead_code)]
     pub async fn delete_fedimint_storage(&self) -> Result<(), MutinyError> {
         self.fedimint_storage.delete_store().await
     }
--- a/mutiny-core/src/federation.rs
+++ b/mutiny-core/src/federation.rs
@@ -58,7 +58,7 @@ use hex::FromHex;
 use lightning::{
     ln::PaymentHash, log_debug, log_error, log_info, log_trace, log_warn, util::logger::Logger,
 };
-use lightning_invoice::Bolt11Invoice;
+use lightning_invoice::{Bolt11Invoice, RoutingFees};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use std::{collections::HashMap, fmt::Debug, sync::Arc};
 use std::{
@@ -143,6 +143,22 @@ pub struct FederationIdentity {
     pub federation_name: Option<String>,
     pub federation_expiry_timestamp: Option<String>,
     pub welcome_message: Option<String>,
+    pub gateway_fees: Option<GatewayFees>,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
+pub struct GatewayFees {
+    pub base_msat: u32,
+    pub proportional_millionths: u32,
+}
+
+impl From<RoutingFees> for GatewayFees {
+    fn from(val: RoutingFees) -> Self {
+        GatewayFees {
+            base_msat: val.base_msat,
+            proportional_millionths: val.proportional_millionths,
+        }
+    }
 }
 
 // This is the FederationIndex reference that is saved to the DB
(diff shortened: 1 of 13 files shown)


benthecarman (approved):

TonyGiorgio:
Need to not delete the fedimint data itself, in case the user adds the same federation again later.

mutiny-core/src/federation.rs
```diff
@@ -76,6 +79,12 @@ const FEDIMINT_OPERATIONS_LIST_MAX: usize = 100;
 
 pub const FEDIMINTS_PREFIX_KEY: &str = "fedimints/";
 
+// Default signet/mainnet federation gateway info
+const SIGNET_GATEWAY: &str = "0256f5ef1d986e9abf559651b7167de28bfd954683cd0f14703be12d1421aedc55";
+const MAINNET_GATEWAY: &str = "025b9f090d3daab012346701f27d1c220d6d290f6b498255cddc492c255532a09d";
+const SIGNET_FEDERATION: &str = "c8d423964c7ad944d30f57359b6e5b260e211dcfdb945140e28d4df51fd572d2";
+const MAINNET_FEDERATION: &str = "c36038cce5a97e3467f03336fa8e7e3410960b81d1865cda2a609f70a8f51efb";
```
benthecarman, line 86: nit instead of like this might be cleaner if it was an array:

```rust
const PREF_GATEWAYS [(&str, &str); 2] = [
("c8d423964c7ad944d30f57359b6e5b260e211dcfdb945140e28d4df51fd572d2", "0256f5ef1d986e9abf559651b7167de28bfd954683cd0f14703be12d1421aedc55"), // signet
("c36038cce5a97e3467f03336fa8e7e3410960b81d1865cda2a609f70a8f51efb", "025b9f090d3daab012346701f27d1c220d6d290f6b498255cddc492c255532a09d") //mainnet
]
```

this way it is simpler to add more in the future

mutiny-core/src/lib.rs
```diff
@@ -1126,18 +1127,16 @@ impl<S: MutinyStorage> MutinyWallet<S> {
         amount: Option<u64>,
         labels: Vec<String>,
     ) -> Result<MutinyInvoice, MutinyError> {
-        let federation_ids = self.list_federation_ids().await?;
+        let amt = amount.map_or(Err(MutinyError::InvalidArgumentsError), Ok)?;
```
benthecarman, line 1221: if we're throwing an error if this is None, should jut change the param to not be optional

benthecarman (commented):
looks like you moved over the commit from #944 but did not do @futurepaul's comment: https://github.com/MutinyWallet/mutiny-node/pull/944#pullrequestreview-1836203673

otherwise lgtm

mutiny-core/src/federation.rs
TonyGiorgio: No thanks, I have no intention on using this after fedimint fixes their vetted stuff.

TonyGiorgio:
> looks like you moved over the commit from #944 but did not do @futurepaul's comment: [#944 (review)](https://github.com/MutinyWallet/mutiny-node/pull/944#pullrequestreview-1836203673)

Sure I'll add some stuff there today.

mutiny-core/src/lib.rs
```diff
@@ -1280,7 +1275,54 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+
+        let federation_id = &federation_ids[0];
+        let federation_lock = self.federations.read().await;
+        let fedimint_client = federation_lock
+            .get(federation_id)
+            .ok_or(MutinyError::NotFound)?;
+
+        // if the user provided amount, this is easy
+        if let Some(amt) = amount {
+            let (_inv, fee) = self.node_manager.create_invoice(amt).await?;
```
benthecarman, line 1306: if we do it this way then we'll fill up the storage with unused invoices. Can't we just call `get_lsp_fee_msat` instead

mutiny-core/src/lib.rs
```diff
@@ -1067,6 +1082,66 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+            a
+        } else if let Some(i) = inv {
+            if let Some(a) = i.amount_milli_satoshis() {
+                a
+            } else {
+                return Err(MutinyError::BadAmountError);
+            }
+        } else {
+            return Err(MutinyError::BadAmountError);
+        };
```
benthecarman, line 1105: could simplify this as
```rust
let amt = amt_sats
            .or(inv.and_then(|i| i.amount_milli_satoshis()))
            .ok_or(MutinyError::BadAmountError)?;
```

mutiny-core/src/lib.rs
```diff
@@ -2072,14 +2288,143 @@ pub(crate) async fn create_new_federation<S: MutinyStorage>(
         federation_name,
         federation_expiry_timestamp,
         welcome_message,
+        gateway_fees,
     })
 }
 
+// max amount that can be spent through a gateway
+fn max_spendable_amount(amount_sat: u64, routing_fees: GatewayFees) -> Option<u64> {
```
benthecarman, line 2296: if you make this take a `&GatewayFees` you can get rid of some clones

mutiny-core/src/lib.rs
```diff
@@ -2072,14 +2288,143 @@ pub(crate) async fn create_new_federation<S: MutinyStorage>(
...
+        return Some((initial_max / 1_000.0).floor() as u64);
+    }
+
+    let mut new_max = initial_max;
+    while new_max < amount_msat {
+        let new_check = new_max + 1.0;
+
+        let prop_fee_sat = (new_check * routing_fees.proportional_millionths as f64) / 1_000_000.0;
+
+        let new_amt = new_check + routing_fees.base_msat as f64 + prop_fee_sat;
```
benthecarman, line 2317: why is the fees added here?

benthecarman: some comments in this function would help explain the logic of everything

benthecarman (changes requested):

mutiny-core/src/lib.rs
TonyGiorgio: because there's a new amount, so it needs a new fee

TonyGiorgio:
I added swap labels but they only work on the outgoing payment.

benthecarman (approved):

PR #999, opened 2024-01-30
Create index for sorted activity

benthecarman:
Now keeps a sorted index of our activity at all time. This should make our activity much faster with only a slight overhead increase.

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -42,20 +42,19 @@ pub mod vss;
 mod test_utils;
 
 use crate::cashu::CashuHttpClient;
+use crate::federation::GatewayFees;
 pub use crate::gossip::{GOSSIP_SYNC_TIME_KEY, NETWORK_GRAPH_KEY, PROB_SCORER_KEY};
 pub use crate::keymanager::generate_seed;
-pub use crate::ldkstorage::{CHANNEL_MANAGER_KEY, MONITORS_PREFIX_KEY};
+pub use crate::ldkstorage::{CHANNEL_CLOSURE_PREFIX, CHANNEL_MANAGER_KEY, MONITORS_PREFIX_KEY};
+use crate::storage::{
+    list_payment_info, IndexItem, MutinyStorage, DEVICE_ID_KEY, EXPECTED_NETWORK_KEY,
+    NEED_FULL_SYNC_KEY, ONCHAIN_PREFIX, PAYMENT_INBOUND_PREFIX_KEY, PAYMENT_OUTBOUND_PREFIX_KEY,
+};
 use crate::{auth::MutinyAuthClient, logging::MutinyLogger};
 use crate::{error::MutinyError, nostr::ReservedProfile};
 use crate::{
     event::{HTLCStatus, MillisatAmount, PaymentInfo},
     onchain::FULL_SYNC_STOP_GAP,
-};
-use crate::{
-    federation::GatewayFees,
-    storage::{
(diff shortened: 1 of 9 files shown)


benthecarman:
Changed to `limit`/`offset` instead of `start`/`until` as we talked irl

mutiny-core/src/lib.rs
```diff
@@ -1369,39 +1429,94 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+                        let user_channel_id_str = item
+                            .key
+                            .trim_start_matches(CHANNEL_CLOSURE_PREFIX)
+                            .splitn(2, '_') // Channel closures have `_{node_id}` at the end
+                            .collect::<Vec<&str>>()[0];
+                        let user_channel_id: [u8; 16] = FromHex::from_hex(user_channel_id_str)?;
+                        closure.user_channel_id = Some(user_channel_id);
+                    }
+                    activities.push(ActivityItem::ChannelClosed(closure));
+                }
```
TonyGiorgio, line 1510: We have `list_channel_closures` which is now no longer used or accessed, we should move the logic here into a method that lists channel closes which can be reusable, and remove the old one.

mutiny-core/src/lib.rs
```diff
@@ -1369,39 +1429,94 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+            match (limit, offset) {
+                (None, None) => vec,
+                (Some(limit), Some(offset)) => {
+                    let start = offset;
+                    let end = start.saturating_add(limit).min(vec.len());
+                    vec[start..end].to_vec()
+                }
+                (None, Some(offset)) => vec[offset..].to_vec(),
+                (Some(limit), None) => vec[..limit].to_vec(),
+            }
```
TonyGiorgio, line 1478:
```suggestion
            let (start, end) = match (limit, offset) {
                (None, None) => (0, vec.len()),
                (Some(limit), Some(offset)) => {
                    let start = offset.min(vec.len());
                    let end = start.saturating_add(limit).min(vec.len());
                    (start, end)
                }
                (None, Some(offset)) => (offset.min(vec.len()), vec.len()),
                (Some(limit), None) => (0, limit.min(vec.len())),
            };

            vec[start..end].to_vec()
```

I think you can make this easier to read and safer by limiting the scope of what happens in the match. Just make it return the start and end and only need to call `to_vec()` in one place.

TonyGiorgio (changes requested):
Screenshot 2024-02-02 at 11 43 42 AM

An unknown 0 sat on chain payment appeared during some channel closes.

Screenshot 2024-02-02 at 11 47 15 AM

Multiple force close items are showing up when I only had one. 

Further, I cannot test this long enough without getting OOM issues.

TonyGiorgio:
Screenshot 2024-02-02 at 11 50 40 AM

This lightning payment is showing up between the older on chain force closes that are duplicated

benthecarman:
Fixed all the duplication errors which also seemed to fixed the crashing for me too, curious if it helps for you as well.

TonyGiorgio:
What if we did this and defaulted the activity list to the last ~30 items? 

My phone is starting to get very slow. And having the new social home screen basically being a "view all"  for the transaction list, I would rather it not be that much.

benthecarman:
Makes sense to me, can rebase

TonyGiorgio (approved):
Code looks good. If you just want to test it to make sure it works with current mutiny master, that would be great.

benthecarman:
Found a potential index out of bounds error when testing, fixed it and added tests. Otherwise everything else felt good

TonyGiorgio (approved):

benthecarman:
sorry last push, chatgpt told me another check to add just in case:

```
// handle start > end
if start > end {
	return vec![];
}
```

PR #1003, opened 2024-02-04
Add any missing channel capacity to reserve

benthecarman:
To avoid confusion like in https://github.com/MutinyWallet/mutiny-web/issues/863 lets just add any remaining capacity to the reserve

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -148,13 +148,20 @@ pub struct MutinyChannel {
 
 impl From<&ChannelDetails> for MutinyChannel {
     fn from(c: &ChannelDetails) -> Self {
+        let size = c.channel_value_satoshis;
+        let balance = c.next_outbound_htlc_limit_msat / 1_000;
+        let inbound = c.inbound_capacity_msat / 1_000;
+
+        // Don't calculate reserve, just make it what we didn't
+        // account for in balance and inbound
+        let reserve = size - (balance + inbound);
+
         MutinyChannel {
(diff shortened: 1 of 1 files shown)


mutiny-core/src/nodemanager.rs
```diff
@@ -142,13 +142,22 @@ pub struct MutinyChannel {
...
     fn from(c: &ChannelDetails) -> Self {
+        let size = c.channel_value_satoshis;
+        let balance = c.next_outbound_htlc_limit_msat / 1_000;
+        let inbound = c.inbound_capacity_msat / 1_000;
+        let mut reserve = ((c.outbound_capacity_msat - c.next_outbound_htlc_limit_msat) / 1_000)
+            + c.unspendable_punishment_reserve.unwrap_or(0);
+
+        // if there is any diff in the channel size and the other totals,
+        // just add it to the reserve to avoid confusion
+        reserve += size - (balance + inbound + reserve);
```
TonyGiorgio, line 153: I don't understand this. You're adding to reserve an amount that subtracts the reserve. I'd like to see some numbers / tests for these calculations.

TonyGiorgio (changes requested):

mutiny-core/src/nodemanager.rs
benthecarman: Okay simplified

TonyGiorgio (approved):
i have absolutely no idea if this is correct and there's no tests but yolo i guess

TonyGiorgio:
needs rebase

PR #1004, opened 2024-02-04
Simplify input_pairs signature

DanGould:
just a function signature simplification


DanGould:
noob move on my part. please ignore

benthecarman:
I think it would have worked, just needed to change to `iter_funding_utxos` 

I assume this is a slight performance increase we are no longer using `Box` so it doesn't need to be allocated to the heap

PR #1010, opened 2024-02-06
Make node manager locks RwLocks

### LINKED ISSUE #925: Find other places we can use async RwLock
There's probably other places in the codebase we can do this too, would be a good followup

_Originally posted by @benthecarman in https://github.com/MutinyWallet/mutiny-node/issues/924#issuecomment-1874681808_

benthecarman:
Closes #925

Pulled out from #1002 

This was breaking tests in the above PR because the syncing background thread would block us from making other calls (ie get balance). This fixes by putting it in a `RwLock` instead of in a generic `Mutex` as well as making us drop the lock after reading the nodes we are gonna sync instead of holding it during the entire sync duration

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -24,6 +24,7 @@ use crate::{
     storage::{MutinyStorage, DEVICE_ID_KEY, KEYCHAIN_STORE_KEY, NEED_FULL_SYNC_KEY},
 };
 use anyhow::anyhow;
+use async_lock::RwLock;
 use bdk::chain::{BlockId, ConfirmationTime};
 use bdk::{wallet::AddressIndex, FeeRate, LocalUtxo};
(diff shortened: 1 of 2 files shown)


mutiny-core/src/nodemanager.rs
```diff
@@ -1097,13 +1098,17 @@ impl<S: MutinyStorage> NodeManager<S> {
     /// This should be called before syncing the on-chain wallet
     /// to ensure that new on-chain transactions are picked up.
     async fn sync_ldk(&self) -> Result<(), MutinyError> {
-        let nodes = self.nodes.lock().await;
+        // get nodes hashmap, immediately drop lock because sync can take a while
+        let nodes = {
+            let nodes = self.nodes.read().await;
+            nodes.deref().clone()
+        };
```
benthecarman, line 1105: this is the main change that fixed the issue

PR #1013, opened 2024-02-06
Sync scorer data in the background

benthecarman:
Changes to sync scorer in the background rather than upfront, this made the startup ~500ms faster for me. This will still read the scorer on start up so we will still have a scorer up until it is synced.

#1016

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -1,3 +1,4 @@
+use crate::auth::MutinyAuthClient;
 use crate::event::HTLCStatus;
 use crate::labels::LabelStorage;
 use crate::logging::LOGGING_KEY;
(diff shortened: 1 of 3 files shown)


mutiny-core/src/nodemanager.rs
```diff
@@ -578,11 +577,6 @@ impl<S: MutinyStorage> NodeManager<S> {
     /// Creates a background process that will sync the wallet with the blockchain.
     /// This will also update the fee estimates every 10 minutes.
     pub fn start_sync(nm: Arc<NodeManager<S>>) {
-        // If we are stopped, don't sync
-        if nm.stop.load(Ordering::Relaxed) {
-            return;
-        }
-
```
benthecarman, line 585: removed this check because we just do it immediately once we enter the loop anyways

PR #1015, opened 2024-02-06
Start fetching auth JWT immediately

benthecarman:
The main change for this in `mutiny-wasm/lib.rs`. The idea is to immediately start fetching the JWT for auth, before we would only fetch it once we got our first VSS request, this saves us about ~200ms because otherwise we wait until we have read all of indexed db until we make this call.

Needed to modify the auth client to only allow one call to `retrieve_new_jwt` at a time, otherwise this would sometimes not complete by the time the first VSS call came in and we'd request multiple JWTs.

Otherwise just adds some timing logs and tiny optimizations to indexed db

#1016

### DIFF
--- a/mutiny-core/src/auth.rs
+++ b/mutiny-core/src/auth.rs
@@ -6,16 +6,17 @@ use crate::{
     networking::websocket::{SimpleWebSocket, WebSocketImpl},
     utils,
 };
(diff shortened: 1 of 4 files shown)


benthecarman:
Ran into something weird, don't merge this one until I confirm its not from this

PR #1021, opened 2024-02-09
feat: add cashu token melting (redeem) to mutiny-node

oleonardolima:
- add a new `melt_cashu_token` method to `mutiny-wasm`
- add a new `CashuMintUrlFailure` error variant too

PR #1026, opened 2024-02-12
add cashu token melting to mutiny-node

elnosh:
following up on @oleonardolima outline on https://github.com/MutinyWallet/mutiny-node/pull/1021.

### DIFF
--- a/mutiny-core/Cargo.toml
+++ b/mutiny-core/Cargo.toml
@@ -4,7 +4,10 @@ cargo-features = ["per-package-target"]
 name = "mutiny-core"
 version = "0.6.0-rc1"
 edition = "2021"
-authors = ["Tony Giorgio <tony@mutinywallet.com>", "benthecarman <ben@mutinywallet.com>"]
+authors = [
+    "Tony Giorgio <tony@mutinywallet.com>",
+    "benthecarman <ben@mutinywallet.com>",
+]
 description = "The core SDK for the mutiny node"
 license = "MIT"
 documentation = "https://docs.rs/mutiny-core"
@@ -12,13 +15,24 @@ homepage = "https://mutinywallet.com"
 repository = "https://github.com/mutinywallet/mutiny-node"
 
 [dependencies]
-lnurl-rs = { version = "0.4.0", default-features = false, features = ["async", "async-https"] }
+lnurl-rs = { version = "0.4.0", default-features = false, features = [
+    "async",
+    "async-https",
+] }
 
 cfg-if = "1.0.0"
 bip39 = { version = "2.0.0" }
-bitcoin = { version = "0.30.2", default-features = false, features = ["std", "serde", "secp-recovery", "rand"] }
+bitcoin = { version = "0.30.2", default-features = false, features = [
+    "std",
+    "serde",
+    "secp-recovery",
+    "rand",
+] }
 bdk = { version = "1.0.0-alpha.5" }
-bdk_esplora = { version = "0.7.0", default-features = false, features = ["std", "async-https"] }
+bdk_esplora = { version = "0.7.0", default-features = false, features = [
+    "std",
+    "async-https",
+] }
 bdk_chain = { version = "0.9.0", features = ["std"] }
 bdk-macros = "0.6.0"
 getrandom = { version = "0.2" }
@@ -27,19 +41,38 @@ serde = { version = "^1.0", features = ["derive"] }
 serde_json = { version = "^1.0" }
 uuid = { version = "1.1.2", features = ["v4"] }
 esplora-client = { version = "0.6", default-features = false }
-lightning = { version = "0.0.121", default-features = false, features = ["max_level_trace", "grind_signatures", "std"] }
+lightning = { version = "0.0.121", default-features = false, features = [
+    "max_level_trace",
+    "grind_signatures",
+    "std",
+] }
 lightning-invoice = { version = "0.29.0", features = ["serde"] }
 lightning-rapid-gossip-sync = { version = "0.0.121" }
 lightning-background-processor = { version = "0.0.121", features = ["futures"] }
-lightning-transaction-sync = { version = "0.0.121", default-features = false, features = ["esplora-async-https"] }
+lightning-transaction-sync = { version = "0.0.121", default-features = false, features = [
+    "esplora-async-https",
+] }
 lightning-liquidity = { git = "https://github.com/lightningdevkit/lightning-liquidity.git", rev = "478ccf9324e2650d200ea289a0ba8905afe420b6" }
 chrono = "0.4.22"
 futures-util = { version = "0.3", default-features = false }
-reqwest = { version = "0.11", default-features = false, features = ["multipart", "json"] }
+reqwest = { version = "0.11", default-features = false, features = [
+    "multipart",
+    "json",
+] }
 async-trait = "0.1.68"
 url = { version = "2.3.1", features = ["serde"] }
-nostr = { version = "0.28.0", default-features = false, features = ["nip04", "nip05", "nip47", "nip57"] }
-nostr-sdk = { version = "0.28.0", default-features = false, features = ["nip04", "nip05", "nip47", "nip57"] }
+nostr = { version = "0.28.0", default-features = false, features = [
+    "nip04",
+    "nip05",
+    "nip47",
+    "nip57",
+] }
+nostr-sdk = { version = "0.28.0", default-features = false, features = [
+    "nip04",
+    "nip05",
+    "nip47",
+    "nip57",
+] }
 cbc = { version = "0.1", features = ["alloc"] }
 aes = { version = "0.8" }
 jwt-compact = { version = "0.8.0-beta.1", features = ["es256k"] }
@@ -83,8 +116,20 @@ gloo-net = { version = "0.4.0" }
 instant = { version = "0.1", features = ["wasm-bindgen"] }
 getrandom = { version = "0.2", features = ["js"] }
 # add nip07 feature for wasm32
-nostr = { version = "0.28.0", default-features = false, features = ["nip04", "nip05", "nip07", "nip47", "nip57"] }
-nostr-sdk = { version = "0.28.0", default-features = false, features = ["nip04", "nip05", "nip07", "nip47", "nip57"] }
+nostr = { version = "0.28.0", default-features = false, features = [
+    "nip04",
+    "nip05",
+    "nip07",
+    "nip47",
+    "nip57",
+] }
+nostr-sdk = { version = "0.28.0", default-features = false, features = [
+    "nip04",
+    "nip05",
+    "nip07",
+    "nip47",
+    "nip57",
+] }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokio = { version = "1", features = ["rt"] }
--- a/mutiny-core/Cargo.toml
+++ b/mutiny-core/Cargo.toml
@@ -4,10 +4,7 @@ cargo-features = ["per-package-target"]
 name = "mutiny-core"
 version = "0.6.0-rc1"
 edition = "2021"
-authors = [
-    "Tony Giorgio <tony@mutinywallet.com>",
-    "benthecarman <ben@mutinywallet.com>",
-]
+authors = ["Tony Giorgio <tony@mutinywallet.com>", "benthecarman <ben@mutinywallet.com>"]
 description = "The core SDK for the mutiny node"
 license = "MIT"
 documentation = "https://docs.rs/mutiny-core"
@@ -15,24 +12,13 @@ homepage = "https://mutinywallet.com"
 repository = "https://github.com/mutinywallet/mutiny-node"
 
 [dependencies]
-lnurl-rs = { version = "0.4.0", default-features = false, features = [
-    "async",
-    "async-https",
-] }
+lnurl-rs = { version = "0.4.0", default-features = false, features = ["async", "async-https"] }
 
 cfg-if = "1.0.0"
 bip39 = { version = "2.0.0" }
-bitcoin = { version = "0.30.2", default-features = false, features = [
-    "std",
-    "serde",
-    "secp-recovery",
-    "rand",
-] }
+bitcoin = { version = "0.30.2", default-features = false, features = ["std", "serde", "secp-recovery", "rand"] }
 bdk = { version = "1.0.0-alpha.5" }
-bdk_esplora = { version = "0.7.0", default-features = false, features = [
-    "std",
-    "async-https",
-] }
+bdk_esplora = { version = "0.7.0", default-features = false, features = ["std", "async-https"] }
 bdk_chain = { version = "0.9.0", features = ["std"] }
 bdk-macros = "0.6.0"
 getrandom = { version = "0.2" }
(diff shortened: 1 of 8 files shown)


mutiny-core/src/error.rs
```diff
@@ -161,6 +162,10 @@ pub enum MutinyError {
     /// Payjoin configuration error
     #[error("Payjoin configuration failed.")]
     PayjoinConfigError,
+    #[error("Error calling Cashu mint")]
+    MokshaWalletError,
```
benthecarman, line 166: Can you add a doc string for this

mutiny-core/src/error.rs
```diff
@@ -239,6 +244,7 @@ impl PartialEq for MutinyError {
             (Self::NostrError, Self::NostrError) => true,
             (Self::IncorrectPassword, Self::IncorrectPassword) => true,
             (Self::SamePassword, Self::SamePassword) => true,
+            (Self::MokshaWalletError, Self::MokshaWalletError) => true,
```
benthecarman, line 247: need to add `EmptyMintURLError` here

mutiny-core/src/lib.rs
```diff
@@ -2189,6 +2197,46 @@ impl<S: MutinyStorage> MutinyWallet<S> {
     pub fn is_safe_mode(&self) -> bool {
         self.safe_mode
     }
+
+    /// Calls upon a Cashu mint and withdrawls/redeem/melts the token from it.
+    pub async fn melt_cashu_token(&self, token_v3: TokenV3) -> Result<bool, MutinyError> {
```
benthecarman, line 2202: I think it'd make more sense if this function returned `MutinyInvoice`, can just call `self.get_invoice(mutiny_invoice_str)` at the end to get it.

if it doesn't end up paid can just return an error like `MutinyError::RoutingFailed` or something

mutiny-core/src/lib.rs
```diff
@@ -2189,6 +2197,46 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+        let mint_client = CrossPlatformHttpClient::new();
+        let mut paid = false;
+
+        for token in token_v3.tokens {
+            let mint_url = match token.mint {
+                Some(url) => url,
+                None => return Err(MutinyError::EmptyMintURLError),
+            };
+
+            let invoice_amount = token.proofs.total_amount() * (1 - 3 / 100);
```
benthecarman, line 2212: Im not sure that `(1 - 3 / 100)` will actually do anything, this might just make end up as `1` because they are all `u64`s. Might need to do it like `let invoice_amount = token.proofs.total_amount() as f64 * (1.0 - 3.0 / 100.0);`

mutiny-wasm/Cargo.toml
```diff
@@ -43,6 +43,7 @@ urlencoding = "2.1.2"
 once_cell = "1.18.0"
 payjoin = { version = "0.13.0", features = ["send", "base64"] }
 fedimint-core = { git = "https://github.com/fedimint/fedimint", rev = "6a923ee10c3a578cd835044e3fdd94aa5123735a" }
+moksha-core = { git = "https://github.com/ngutech21/moksha" }
```
benthecarman, line 46: we should pin these to a specific commit so stuff doesn't break, can you add `rev = "0b9a32d216ba0782a4aa2cdb19a87589e16d9089"`

mutiny-wasm/src/error.rs
```diff
@@ -160,6 +161,11 @@ pub enum MutinyJsError {
     /// Payjoin configuration error
     #[error("Payjoin configuration failed.")]
     PayjoinConfigError,
+    /// Failed to call on the given Cashu mint URL
+    #[error("Failed to call on the given Cashu mint URL.")]
+    CashuMintUrlFailure,
+    #[error("Mint URL in token is empty")]
+    EmptyMintURLError,
```
benthecarman, line 168: can you add a doc string for this one

benthecarman (changes requested):

mutiny-core/src/lib.rs
elnosh: ok, I think a `Vec<MutinyInvoice>`? Since we are looping through tokens and could generate multiple invoices.

benthecarman: Ah, is there no way to aggregate them? Could we not reissue them into a single token then do one withdrawal

benthecarman: Oh interesting. Okay then yeah `Vec<MutinyInvoice>` makes sense to me

mutiny-core/src/lib.rs
```diff
@@ -2189,6 +2192,50 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+                .post_melt_quote_bolt11(
+                    &mint_url,
+                    mutiny_invoice_str.to_string(),
+                    CurrencyUnit::Sat,
+                )
+                .await?;
+
+            let post_melt_bolt11_response = mint_client
+                .post_melt_bolt11(&mint_url, token.proofs, melt_quote_res.quote, vec![])
+                .await?;
```
benthecarman, line 2231: looks like this has a `change` in the return. I assume this is from NUT-8 for overpaid fees?

Maybe we can save them to storage and try to claim them along side our next withdrawal?

benthecarman: Could be fine for a follow up if you just want to put a log for now

mutiny-core/src/lib.rs
```diff
@@ -2189,6 +2192,50 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+
+            if post_melt_bolt11_response.paid {
+                invoices.push(mutiny_invoice);
```
benthecarman, line 2234: should refetch the `mutiny_invoice` here so it has all the filled in metadata after it is paid.

should be `self.get_invoice`

benthecarman (commented):
really close, just a couple things

mutiny-core/src/lib.rs
elnosh: yeah it's for overpaid fees

oleonardolima: I agree, I think it could be even a new issue just so we don't forget to address it, wdyt @elnosh ?

oleonardolima (approved):
utACK

mutiny-core/src/lib.rs
```diff
@@ -2230,6 +2233,51 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+
+            let invoice_amount = token.proofs.total_amount() as f64 * (1.0 - 3.0 / 100.0);
+
+            let mut mutiny_invoice = self
+                .create_invoice(invoice_amount as u64, vec!["Cashu Token Melt".to_string()])
+                .await?;
+
+            let mutiny_invoice_str = mutiny_invoice
+                .clone()
+                .bolt11
```
benthecarman, line 2259: nit: if you do `.bolt11.clone()` instead of `.clone().bolt11` then it'll only clone the invoice instead of the whole struct.

mutiny-core/src/lib.rs
```diff
@@ -2230,6 +2233,51 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+        for token in token_v3.tokens {
+            let mint_url = match token.mint {
+                Some(url) => url,
+                None => return Err(MutinyError::EmptyMintURLError),
+            };
+
+            let invoice_amount = token.proofs.total_amount() as f64 * (1.0 - 3.0 / 100.0);
+
+            let mut mutiny_invoice = self
+                .create_invoice(invoice_amount as u64, vec!["Cashu Token Melt".to_string()])
```
benthecarman, line 2254: Can you make `"Cashu Token Melt"` a const at the top of this file, we have one for `SWAP` as well

mutiny-core/src/lib.rs
benthecarman: still need to do this @elnosh

mutiny-wasm/src/lib.rs
```diff
@@ -910,6 +911,18 @@ impl MutinyWallet {
...
 
+    /// Calls upon a Cash mint and melts the token from it.
+    #[wasm_bindgen]
+    pub async fn melt_cashu_token(
+        &self,
+        maybe_token: String,
+    ) -> Result<JsValue /* Vec<MutinyInvoice> */, MutinyJsError> {
+        let token = TokenV3::deserialize(maybe_token)?;
+        Ok(JsValue::from_serde(
+            &self.inner.melt_cashu_token(token).await?,
```
benthecarman, line 922: We need to convert this into mutiny-wasm's version of `MutinyInvoice`

Have to do something like this.

```rust
let result = self.inner.melt_cashu_token(token).await?;
let invoices: Vec<MutinyInvoice> = result.into_iter().map(|i| i.into()).collect();
Ok(JsValue::from_serde(&invoices)?)
```

benthecarman (approved):
lgtm

mutiny-core/src/lib.rs
```diff
@@ -2286,6 +2290,60 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+                .await?;
+            let amt: u64 = post_melt_bolt11_response
+                .change
+                .iter()
+                .map(|s| s.amount)
+                .sum();
+            log_debug!(
+                self.logger,
+                "will need to add extra outputs for {amt} for overpaid fees"
+            );
```
TonyGiorgio, line 2338: I'm a little confused why this debug is here. Should this be a warning? And does it always occur?

TonyGiorgio (changes requested):

mutiny-core/src/lib.rs
```diff
@@ -2286,6 +2290,60 @@ impl<S: MutinyStorage> MutinyWallet<S> {
     pub fn is_safe_mode(&self) -> bool {
         self.safe_mode
     }
+
+    /// Calls upon a Cashu mint and withdrawls/redeem/melts the token from it.
+    pub async fn melt_cashu_token(
+        &self,
+        token_v3: TokenV3,
+    ) -> Result<Vec<MutinyInvoice>, MutinyError> {
+        let mint_client = CrossPlatformHttpClient::new();
```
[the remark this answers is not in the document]
elnosh: you are right. I think we can get rid of moksha-wallet and make the http calls to the mint without it.

mutiny-core/src/lib.rs
```diff
@@ -2286,6 +2290,60 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+        let mut invoices: Vec<MutinyInvoice> = Vec::with_capacity(token_v3.tokens.len());
+
+        for token in token_v3.tokens {
+            let mint_url = match token.mint {
+                Some(url) => url,
+                None => return Err(MutinyError::EmptyMintURLError),
+            };
+
+            let invoice_amount = token.proofs.total_amount() as f64 * (1.0 - 3.0 / 100.0);
```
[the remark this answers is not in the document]
elnosh: the amount for the invoice needs to be a bit lower than the actual amount of the cashu token to account for potential fees. Not sure on best way to calculate it but landed on that number after discussing with @benthecarman

mutiny-core/src/lib.rs
elnosh: @benthecarman commented to add that log that could be for a follow-up. Follow up could be to add outputs param to the `post_melt_bolt` call in order to get the overpaid fees if any. Or if we create an invoice for "right" amount, we could get rid of this.

mutiny-core/src/lib.rs
```diff
@@ -2290,6 +2300,118 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+            }
+        }
+    }
+
+    /// Calls upon a Cashu mint and redeems/melts the token.
+    pub async fn melt_cashu_token(
+        &self,
+        token_v3: TokenV3,
+    ) -> Result<Vec<MutinyInvoice>, MutinyError> {
+        let client = reqwest::Client::new();
```
benthecarman, line 2349: its kinda bad practice to init a new `reqwest` client for every request. It doesn't matter in the browser but for other build targets this will make us not do any of the nice connection pooling and stuff that reqwest would do for us.

if you make the `CashuHttpClient` that could have its own `reqwest::Client` so it doesn't to re-init every request.

mutiny-core/src/lib.rs
```diff
@@ -2290,6 +2300,118 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+                Some(url) => url,
+                None => return Err(MutinyError::EmptyMintURLError),
+            };
+
+            let total_proofs_amount = token.proofs.total_amount();
+            // create invoice for 1% less than proofs amount
+            let mut invoice_amount = total_proofs_amount as f64 * (1.0 - 1.0 / 100.0);
+            let mut mutiny_invoice: MutinyInvoice;
+            let mut mutiny_invoice_str: Bolt11Invoice;
+            let mut melt_quote_res: PostMeltQuoteBolt11Response;
```
benthecarman, line 2363: since you aren't using these outside the loop there's no need to initialize them out here with a `mut`.

elnosh: yeah I'm using them outside the loop

```
if post_melt_bolt11_response.paid {
    mutiny_invoice = self.get_invoice(&mutiny_invoice_str).await?;
    invoices.push(mutiny_invoice);
}
```

benthecarman:
looks like you have some rebase conflicts in the cargo.lock, I would just revert the changes in there and then compile again, let cargo fix it for you

mutiny-core/src/lib.rs
```diff
@@ -2289,6 +2299,76 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+                    request: mutiny_invoice_str.to_string(),
+                    unit: CurrencyUnit::Sat,
+                };
+
+                melt_quote_res = self
+                    .cashu_client
+                    .mint_post(
+                        &mint_url.join("/v1/melt/quote/bolt11")?,
+                        json!(quote_request),
+                    )
```
benthecarman, line 2343: can make this a dedicated function on the cashu client. This way you can just pass the mint url and a `PostMeltQuoteBolt11Request`. Will make things easier to use in the future and we could eventually mock it for tests too

mutiny-core/src/lib.rs
```diff
@@ -2289,6 +2299,76 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+            let melt_request = PostMeltBolt11Request {
+                quote: melt_quote_res.quote,
+                inputs: token.proofs,
+                outputs: vec![],
+            };
+
+            let post_melt_bolt11_response: PostMeltBolt11Response = self
+                .cashu_client
+                .mint_post(&mint_url.join("v1/melt/bolt11")?, json!(melt_request))
+                .await?;
```
benthecarman, line 2363: should do the same for this function

mutiny-core/src/lib.rs
```diff
@@ -2289,6 +2299,76 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+
+        for token in token_v3.tokens {
+            let mint_url = match token.mint {
+                Some(url) => url,
+                None => return Err(MutinyError::EmptyMintURLError),
+            };
+
+            let total_proofs_amount = token.proofs.total_amount();
+            // create invoice for 1% less than proofs amount
+            let mut invoice_amount = total_proofs_amount as f64 * (1.0 - 1.0 / 100.0);
```
benthecarman, line 2318: this would be more readable as
`let mut invoice_amount = total_proofs_amount as f64 * 0.99;`

TonyGiorgio (approved):
Looks really good! Nice work. 

The only question/comment I have on is around the multiple token thing. Is that common? If so, I wonder if all of the errors in the `for` loop should just result in a `continue` and if a new `PartialCashuError` could be introduced. If it's not very common, then I think it's fine to leave it as is.

benthecarman (approved):

mutiny-core/Cargo.toml
```diff
@@ -56,6 +56,7 @@ fedimint-mint-client = { git = "https://github.com/fedimint/fedimint", rev = "6a
 fedimint-ln-client = { git = "https://github.com/fedimint/fedimint", rev = "6a923ee10c3a578cd835044e3fdd94aa5123735a" }
 fedimint-bip39 = { git = "https://github.com/fedimint/fedimint", rev = "6a923ee10c3a578cd835044e3fdd94aa5123735a" }
 fedimint-ln-common = { git = "https://github.com/fedimint/fedimint", rev = "6a923ee10c3a578cd835044e3fdd94aa5123735a" }
+moksha-core = { git = "https://github.com/ngutech21/moksha", rev = "0b9a32d216ba0782a4aa2cdb19a87589e16d9089" }
```
benthecarman, line 59: I updated waila to the commit `18d99977965662d46ccec29fecdb0ce493745917`, can you do the same here so their in sync, then we should be good to go

benthecarman (changes requested):

benthecarman (approved):

PR #1029, opened 2024-02-15
Don't sync unconfirmed txids when syncing bdk

benthecarman:
This is what @notmandatory recommended we do in discord, this seems a little unsafe to me as we aren't getting the status of these txs anymore, but maybe its fine because it'd always be associated with our spks anyways.


TonyGiorgio:
> this seems a little unsafe to me as we aren't getting the status of these txs anymore

Is this when the transaction confirms? If there's any problems, a resync should fix it, right?

benthecarman:
> Is this when the transaction confirms? If there's any problems, a resync should fix it, right?

Well when we sync we only check unused spks, but since this address would have a transaction it would be used and we'd stop checking it

TonyGiorgio:
Made no difference, still 117 requests on my mainnet wallet.

TonyGiorgio:
Feel free to test this on mainnet too @benthecarman

TonyGiorgio:
wasn't needed

PR #1031, opened 2024-02-15
feature package workflow attempt

futurepaul:
the ideal here is that we get a tagged release on npm for feature branches without overwriting the real one. not sure if this is sufficient so we should discuss before dispatching the action


TonyGiorgio (changes requested):
Please add the instructions to the readme as well

PR #1034, opened 2024-02-15
Handle subscriptions created outside the app

### LINKED ISSUE #1033: Restore mutiny+ nwc

benthecarman:
Closes #1033

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -1507,6 +1507,10 @@ impl<S: MutinyStorage> MutinyWallet<S> {
             match self.storage.get_data::<u64>(SUBSCRIPTION_TIMESTAMP) {
                 Ok(Some(timestamp)) if timestamp > now => {
                     // if we have a timestamp and it is in the future, we are subscribed
+                    // make sure we have a NWC profile, this needs to be done in case
+                    // the subscription was created outside the app.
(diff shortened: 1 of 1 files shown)


mutiny-core/src/lib.rs
```diff
@@ -1507,6 +1507,10 @@ impl<S: MutinyStorage> MutinyWallet<S> {
             match self.storage.get_data::<u64>(SUBSCRIPTION_TIMESTAMP) {
                 Ok(Some(timestamp)) if timestamp > now => {
                     // if we have a timestamp and it is in the future, we are subscribed
+                    // make sure we have a NWC profile, this needs to be done in case
+                    // the subscription was created outside the app.
+                    self.ensure_mutiny_nwc_profile(subscription_client, true)
```
TonyGiorgio, line 1512: I've thought about this function, but isn't the problem going to be that there's two NWC connections for the same thing in the user's wallet?

benthecarman: `ensure_mutiny_nwc_profile` only creates a NWC if there isn't a profile already

TonyGiorgio: Right but the NWA flow of an external site is going to make a non mutiny+ profile

TonyGiorgio (approved):

PR #1032, opened 2024-02-15
Update bdk to alpha7

benthecarman:
https://github.com/bitcoindevkit/bdk/releases/tag/v1.0.0-alpha.6

PR #1035, opened 2024-02-16
Increase wasm memory limit to 3mb

benthecarman:
While debugging #1032 found how to increase our memory for wasm. This will 3x from the default stack size to 3mb.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -2,3 +2,10 @@
 target = "wasm32-unknown-unknown"
 # for fedimint to use tokio::task::Builder - https://github.com/fedimint/fedimint/issues/3951
 rustflags = ["--cfg", "tokio_unstable"]
+
+[target.wasm32-unknown-unknown]
+# Increase the stack size to 3MB, the default is 1MB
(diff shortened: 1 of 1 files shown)


TonyGiorgio:
Can we go much bigger?

benthecarman:
> Can we go much bigger?

probably, not sure how'd that effect things though

benthecarman:
tried 3gb and things stopped working so there is an upper limit

TonyGiorgio:
More research should be made into this parameter change, and commented in the file.

benthecarman:
Tried 30mb and things worked fine, imo it'd make sense to increase to 3mb and see how it behaves. 2mb increase is basically nothing on modern devices and we can see if the out of bounds issues go away. If not we can look into large increases to the limit

TonyGiorgio:
That sounds good but I want to know more before we change anything. Surely you found out about this somewhere.

benthecarman:
sorry thought i linked it

https://github.com/rustwasm/wasm-pack/issues/479

TonyGiorgio (approved):

PR #1037, opened 2024-02-17
wip(feat): reissue ecash notes from `OOBNotes``

oleonardolima:


mutiny-core/src/lib.rs
```diff
@@ -1308,6 +1310,22 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+        let maybe_federation_id = federation_ids
+            .iter()
+            .find(|id| id.to_prefix() == oob_notes.federation_id_prefix());
+
+        if let Some(fed_id) = maybe_federation_id {
+            let fedimint_client = federation_lock.get(&fed_id).ok_or(MutinyError::NotFound)?;
+            Ok(fedimint_client.reissue(oob_notes).await?)
+        } else {
+            return Err(MutinyError::NotFound);
+        }
```
TonyGiorgio, line 1326: With the new version of OOB notes, they support having the federation code inside of them, right? I think it would be cool to do it like cashu where it joins the federation just to redeem it on lightning. 

https://github.com/fedimint/fedimint/pull/4231

oleonardolima: yep, we can have that when we bump to v0.3 😁

TonyGiorgio: Ah, didn't realized it wasn't backported.

mutiny-core/src/lib.rs
```diff
@@ -1352,6 +1353,26 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+        let maybe_federation_id = federation_ids
+            .iter()
+            .find(|id| id.to_prefix() == oob_notes.federation_id_prefix());
+
+        if let Some(fed_id) = maybe_federation_id {
+            log_info!(self.logger, "found federation_id {:?}", fed_id);
+            let fedimint_client = federation_lock.get(fed_id).ok_or(MutinyError::NotFound)?;
+            log_info!(self.logger, "got fedimint client for federation_id {:?}", fed_id);
+            let reissue = fedimint_client.reissue(oob_notes).await?;
+            log_info!(self.logger, "successfully reissued for federation_id {:?}", fed_id);
```
benthecarman, line 1369: These don't need to all be info level

mutiny-core/src/lib.rs
```diff
@@ -112,6 +112,7 @@ use crate::utils::parse_profile_metadata;
 use mockall::{automock, predicate::*};
 
 const DEFAULT_PAYMENT_TIMEOUT: u64 = 30;
+const DEFAULT_REISSUE_TIMEOUT: u64 = 60;
```
benthecarman, line 115: Is 60 seconds really needed? I imagine this at most should take like 5?

mutiny-core/src/federation.rs
```diff
@@ -603,6 +603,29 @@ impl<S: MutinyStorage> FederationClient<S> {
...
+        // Get the `MintClientModule`
+        let mint_module = self.fedimint_client.get_first_module::<MintClientModule>();
+
+        // TODO: (@leonardo) Do we need any `extra_meta` ?
+        // Reissue `OOBNotes`
+        let operation_id = mint_module.reissue_external_notes(oob_notes, ()).await?;
+
+        // TODO: (@leonardo) re-think about the results and errors that we need/want
+        match process_reissue_outcome(&mint_module, operation_id, logger.clone()).await? {
+            ReissueExternalNotesState::Created | ReissueExternalNotesState::Failed(_) => {
```
benthecarman, line 618: Why return error on created

mutiny-core/src/federation.rs
```diff
@@ -603,6 +603,29 @@ impl<S: MutinyStorage> FederationClient<S> {
         }
     }
 
+    pub(crate) async fn reissue(&self, oob_notes: OOBNotes) -> Result<bool, MutinyError> {
+        let logger = Arc::clone(&self.logger);
+
+        // Get the `MintClientModule`
+        let mint_module = self.fedimint_client.get_first_module::<MintClientModule>();
+
+        // TODO: (@leonardo) Do we need any `extra_meta` ?
```
benthecarman, line 612: Don't think we need any Metadata, we don't use that anywhere else

mutiny-core/src/federation.rs
```diff
@@ -603,6 +603,29 @@ impl<S: MutinyStorage> FederationClient<S> {
...
+
+        // TODO: (@leonardo) re-think about the results and errors that we need/want
+        match process_reissue_outcome(&mint_module, operation_id, logger.clone()).await? {
+            ReissueExternalNotesState::Created | ReissueExternalNotesState::Failed(_) => {
+                log_info!(logger, "re-issuance of OOBNotes failed!");
+                Err(MutinyError::FedimintReissueFailed)
+            }
+            _ => {
+                log_info!(logger, "re-issuance of OOBNotes was successful!");
+                Ok(true)
```
benthecarman, line 624: This either returns an error or ok(true). Is there any other information we could return that'd be useful, otherwise should just return ()

mutiny-core/src/federation.rs
oleonardolima: I'm not sure how to proceed with this, but if the stream is already finished and the final result is still Created, it didn't get processed by the federation without a real guaranteeing it will.

mutiny-core/src/federation.rs
oleonardolima: No, we don't. I'll update it to ()

.editorconfig
benthecarman, line 1: think this is from your ide

oleonardolima: yep, missed that on .gitignore

mutiny-core/src/federation.rs
oleonardolima: @benthecarman What do you think we should do here with Created variant?
It's not supposed to have that after the whole stream is consumed, so still handle as failure/error ? 🤔

benthecarman: in `process_reissue_outcome` you only return the outcome if it is `Failed` or `Done` so you should just handle those two here, and leave the rest in the `_ => ` case where you can throw an error.

mutiny-core/src/lib.rs
```diff
@@ -119,6 +119,7 @@ use crate::utils::parse_profile_metadata;
 use mockall::{automock, predicate::*};
 
 const DEFAULT_PAYMENT_TIMEOUT: u64 = 30;
+const DEFAULT_REISSUE_TIMEOUT: u64 = 50;
```
benthecarman, line 122: this is only used in federation.rs, would be better to define in there. This should probably go back to 30 seconds too. you're also always casting to `i32`, can just define it as such

mutiny-core/src/federation.rs
```diff
@@ -866,6 +886,53 @@ where
...
+                    match outcome {
+                        ReissueExternalNotesState::Failed(_) | ReissueExternalNotesState::Done => {
+                            log_trace!(
+                                logger,
+                                "streamed outcome received is final {:?}, returning",
+                                outcome
+                            );
+                            return Ok(outcome);
+                        }
+                        _ => { /* ignore and continue */ }
```
benthecarman, line 927: would be helpful to log each state

benthecarman (changes requested):

mutiny-core/src/federation.rs
oleonardolima: wdyt about [eb59a8b](https://github.com/MutinyWallet/mutiny-node/pull/1037/commits/eb59a8b357aec7b89b7c7c487b2e2334185f0aa4)

mutiny-core/src/lib.rs
oleonardolima: nice, done!
but I kept it as u64 to follow the standard, or should I also change the other ones? 🤔

mutiny-core/src/federation.rs
oleonardolima: nice, I'll add it :)

PR #1040, opened 2024-02-19
Add public and private zaps

benthecarman:
This was _almost_ very simple, but sadly doing private zaps with a NIP-07 extension isn't very well supported. The way clients generate the random npub is deterministic from the user's nsec, but since we don't have the user's nsec, we cannot do this. Private zaps will still work, we can just randomly generate the key, the problem is if we see the event in the future, we won't be able to decrypt it ourselves.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -244,6 +244,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic-destructor"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4653a42bf04120a1d4e92452e006b4e3af4ab4afff8fb4af0f1bbb98418adf3e"
+dependencies = [
+ "tracing",
+]
+
+[[package]]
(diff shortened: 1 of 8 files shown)


TonyGiorgio:
> the problem is if we see the event in the future, we won't be able to decrypt it ourselves.

We'll still have some reference for it inside of mutiny, right? And for private zaps, the receiver will still see it was from us?

TonyGiorgio:
> if I do a private zap Ben -> Tony, if I open up in Damus I will see you were private zapped

That's fine IMO, I don't care that much if we're limited here. But will it still show up in Mutiny is more my question.

benthecarman:
> That's fine IMO, I don't care that much if we're limited here. But will it still show up in Mutiny is more my question.

yes

TonyGiorgio:
needs rebase

benthecarman:
New release of rust-nostr came out so was able to simplify

TonyGiorgio (approved):

PR #1041, opened 2024-02-20
Allow dm decrypting to fail

benthecarman:


futurepaul (approved):
this solved my problem and saved my life

PR #1043, opened 2024-02-20
Remove requirement for password for has_node_manager

benthecarman:
We use this to check if we have a wallet initialized, we don't need to decrypt to seed to check. This was breaking the new front end

### DIFF
--- a/mutiny-wasm/src/indexed_db.rs
+++ b/mutiny-wasm/src/indexed_db.rs
@@ -80,6 +80,31 @@ impl IndexedDbStorage {
         })
     }
 
+    pub(crate) async fn has_mnemonic() -> Result<bool, MutinyError> {
+        let indexed_db = Self::build_indexed_db_database().await?;
+        let tx = indexed_db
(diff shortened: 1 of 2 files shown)


futurepaul (approved):
worked for me thank you

mutiny-wasm/src/indexed_db.rs
```diff
@@ -80,6 +80,29 @@ impl IndexedDbStorage {
         })
     }
 
+    pub(crate) async fn has_mnemonic() -> Result<bool, MutinyError> {
+        let indexed_db = Self::build_indexed_db_database().await?;
+        let tx = indexed_db
+            .transaction(&[WALLET_OBJECT_STORE_NAME], TransactionMode::ReadWrite)
```
TonyGiorgio, line 86: Does it need to be `ReadWrite` if it's just reading? Same thing for the log checking too.

TonyGiorgio (changes requested):

mutiny-wasm/src/indexed_db.rs
benthecarman: fixed

TonyGiorgio (approved):

PR #1045, opened 2024-02-21
Expose waila through mutiny-wasm

benthecarman:
probably want @futurepaul to verify this first but I got it working in the front end pretty easily.


TonyGiorgio (approved):

oleonardolima (approved):
ACK :shipit:

PR #1047, opened 2024-02-22
specify payment size with buy request

johncantrell97:
Looks like during the LDK 121 and liquidity library upgrade PR this got missed.  The payment size was not being set with the BuyRequest being issued and so later in the flow when the invoice params are surfaced via the LSPS event the payment_size_msat was being returned as None.  Mutiny was then bailing on the flow because it requires a payment size to be set.

This PR stores the payment amount in the "JitChannelInfo" struct that is used to store the fee information between the get_info and buy requests so that we can pass it along with the buy request.


johncantrell97:
closing in favor of #1050

PR #1048, opened 2024-02-22
Support new NWC commands

### LINKED ISSUE #895: Improve NWC support (return response from send payment, get_balance, make_invoice/lookup_invoice)

benthecarman:
This adds support for `get_balance`, `get_info`, `make_invoice`, and `lookup_invoice`.

Currently missing the multi commands as they'd require a larger refactor that I'd rather do as a follow up. Also left out keysends because that'd require a refactor for budgeting and `list_transactions` because I am not sure how we'd want to implement that while preserving privacy.

Closes #895

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -78,6 +78,7 @@ use ::nostr::nips::nip57;
 #[cfg(target_arch = "wasm32")]
 use ::nostr::prelude::rand::rngs::OsRng;
 use ::nostr::prelude::ZapRequestData;
+use ::nostr::nips::nip47::Method;
 use ::nostr::{EventBuilder, EventId, JsonUtil, Kind};
 #[cfg(target_arch = "wasm32")]
 use ::nostr::{Keys, Tag};
@@ -1653,6 +1654,7 @@ impl<S: MutinyStorage> MutinyWallet<S> {
                                 period: BudgetPeriod::Month,
                             }),
                             NwcProfileTag::Subscription,
+                            vec![Method::PayInvoice], // subscription only needs pay invoice
                         )
                         .await?
                         .nwc_uri
@@ -1662,6 +1664,7 @@ impl<S: MutinyStorage> MutinyWallet<S> {
                             ProfileType::Reserved(ReservedProfile::MutinySubscription),
                             SpendingConditions::RequireApproval,
                             NwcProfileTag::Subscription,
+                            vec![Method::PayInvoice], // subscription only needs pay invoice
                         )
                         .await?
                         .nwc_uri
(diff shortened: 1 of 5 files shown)


TonyGiorgio:
needs rebase

TonyGiorgio:
Failing tests

TonyGiorgio:
needs rebase

TonyGiorgio:
`This adds support for get_balance, get_info, make_invoice, and lookup_invoice.` 

These should probably be allowed by default

benthecarman:
> These should probably be allowed by default

Done

TonyGiorgio (approved):

rolznz:
Cool! How can I test this? do I need to create a new connection? I tried to create one and get "Unknown error occurred" - auto approve, 10000 sats, resets daily

benthecarman:
This hasn't been deployed yet, should be by the end of the month

rolznz:
I tried the Mutiny latest update (PWA version) and am still having issues doing any requests outside of pay_invoice. Is it in this release?

I would love to see Mutiny working with some of the NWC-powered apps like the BuzzPay PoS https://github.com/getAlby/pos

benthecarman:
If you make a new connection it should get the new commands.

rolznz:
I tried again today. `make_invoice` works, but `lookup_invoice` never seems to be updated so the PoS never gets a confirmation of the payment (it polls lookup_invoice). My mutiny wallet was open and I received the payment there.

I updated the PoS to be more relaxed and will show a confirmation rather than erroring if the wrong NWC permissions are given. You should be able to try it: https://github.com/getAlby/pos

benthecarman:
In the logs I am getting these errors after it is paid, seems like something on your end, that error message isn't from us
image

rolznz:
@benthecarman that error happens when we get no reply from the wallet within a timeout. Can you check the reply payload you are sending?

benthecarman:
hmm looks like `rust-nostr` is parsing for `bolt11` but the spec is `invoice` :man_facepalming: my bad

PR #1049, opened 2024-02-22
Only check voltage invoice is within 10 sats

### LINKED ISSUE #1046: Received unexpected invoice from LSP
```
af04 2024-02-22 14:06:24.228 ERROR [mutiny_core::node:1159] Received unexpected invoice from LSP: Some(195630815) when amount was 195630000
```

benthecarman:
Closes #1046

### DIFF
--- a/mutiny-core/src/node.rs
+++ b/mutiny-core/src/node.rs
@@ -1152,21 +1152,10 @@ impl<S: MutinyStorage> Node<S> {
                             })
                             .await?;
 
-                        if invoice.network() != self.network {
-                            return Err(MutinyError::IncorrectNetwork);
-                        }
-
-                        if lsp_invoice.payment_hash() != invoice.payment_hash()
-                            || lsp_invoice.recover_payee_pub_key() != lsp_pubkey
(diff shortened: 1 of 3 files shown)


mutiny-core/src/node.rs
```diff
@@ -1158,7 +1158,8 @@ impl<S: MutinyStorage> Node<S> {
 
                         if lsp_invoice.payment_hash() != invoice.payment_hash()
                             || lsp_invoice.recover_payee_pub_key() != lsp_pubkey
-                            || (lsp_invoice.amount_milli_satoshis() != Some(amount_sat * 1_000)
+                            || (lsp_invoice.amount_milli_satoshis().map(|s| s / 1_000)
```
TonyGiorgio, line 1161: is there not just an `amount_sats` ? What if it rounds the wrong way?

benthecarman: rust always rounds down, ldk only has a msats option

benthecarman: changed to just check within 10 sats

TonyGiorgio (changes requested):
I think this should be pulled out to a different method and tested

TonyGiorgio (approved):

PR #1053, opened 2024-02-26
Add federation code to list_federations

benthecarman:


futurepaul (approved):
did not test anything but looks like it would work

edit: it works

TonyGiorgio (approved):
I did not test this but it looks like it would work

PR #1056, opened 2024-03-07
Don't attempt payments before on chain syncing

### LINKED ISSUE #1055: Don't attempt payments before on chain syncing
I knew this would backfire. Don't allow for payments to happen before syncing occurs. Because it locks up HTLC's with a bad block height.

https://github.com/MutinyWallet/mutiny-web/issues/913

benthecarman:
Closes #1055

Adds an `AtomicBool` that we have synced this session. Not using the one in storage because we should check every new startup. 

Also noticed that we did not have all of our initial send logic for keysends so copied it over.

### DIFF
--- a/mutiny-core/src/nodemanager.rs
+++ b/mutiny-core/src/nodemanager.rs
@@ -510,6 +510,7 @@ impl<S: MutinyStorage> NodeManagerBuilder<S> {
             bitcoin_price_cache: Arc::new(Mutex::new(price_cache)),
             do_not_connect_peers: c.do_not_connect_peers,
             safe_mode: c.safe_mode,
+            has_done_initial_ldk_sync: Arc::new(AtomicBool::new(false)),
         };
 
         Ok(nm)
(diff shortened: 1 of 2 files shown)


mutiny-core/src/nodemanager.rs
```diff
@@ -1238,6 +1241,9 @@ impl<S: MutinyStorage> NodeManager<S> {
             return Err(e);
         }
 
+        // set has synced to true
+        self.has_done_initial_ldk_sync.swap(true, Ordering::SeqCst);
+
```
TonyGiorgio, line 1246: Payments before syncing LDK isn't the problem, the problem is BDK chain tip is not up to date. This needs to be moved down.

TonyGiorgio (changes requested):
Also needs to be reproduced and tested.

mutiny-core/src/nodemanager.rs
benthecarman: I'm confused, how would bdk effect ldk? They don't talk to each other

TonyGiorgio: how is the chain tip retrieved?

benthecarman: ldk syncs it

TonyGiorgio (approved):
I'm trusting that you reproduced and tested this first before merging.

PR #1059, opened 2024-03-08
Show in-flight fedimint sends

### LINKED ISSUE #1022: Fedimint sends are not showing up as pending in UI

alpeb-btc:
Fixes #1022

Right after the payment has initiated successfully, and before persisting the invoice, set its status to InFlight, so it doesn't get filtered out in the activity list.

image

### DIFF
--- a/mutiny-core/src/federation.rs
+++ b/mutiny-core/src/federation.rs
@@ -560,6 +560,7 @@ impl<S: MutinyStorage> FederationClient<S> {
         let mut stored_payment: MutinyInvoice = invoice.clone().into();
         stored_payment.inbound = inbound;
         stored_payment.labels = labels;
+        stored_payment.status = HTLCStatus::InFlight;
         let hash = stored_payment.payment_hash.into_32();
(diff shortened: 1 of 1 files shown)


benthecarman (approved):
thanks!

mutiny-core/src/federation.rs
```diff
@@ -560,6 +560,7 @@ impl<S: MutinyStorage> FederationClient<S> {
         let mut stored_payment: MutinyInvoice = invoice.clone().into();
         stored_payment.inbound = inbound;
         stored_payment.labels = labels;
+        stored_payment.status = HTLCStatus::InFlight;
```
TonyGiorgio, line 563: nice catch. Though we don't typically use `InFlight` much, just wondering if it was `Pending` before and was being filtered out for whatever reason, or really why `InFlight` is used at all.

benthecarman: `Pending` is what we use for invoices we create that are unpaid

PR #1060, opened 2024-03-08
Mint discoverability

benthecarman:
basic implementation of: https://github.com/nostr-protocol/nips/pull/1110

we should probably also add support for creating a recommendation event for federations we've joined

### DIFF
--- a/mutiny-core/src/nostr/mod.rs
+++ b/mutiny-core/src/nostr/mod.rs
@@ -26,9 +26,9 @@ use lnurl::lnurl::LnUrl;
 use nostr::nips::nip47::*;
 use nostr::{
     nips::nip04::{decrypt, encrypt},
-    SecretKey,
+    Alphabet, Event, EventBuilder, EventId, Filter, JsonUtil, Keys, Kind, Metadata, SecretKey,
+    SingleLetterTag, Tag, TagKind, Timestamp,
 };
-use nostr::{Event, EventBuilder, EventId, Filter, JsonUtil, Keys, Kind, Metadata, Tag, Timestamp};
 use nostr_sdk::{Client, NostrSigner, RelayPoolNotification};
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
@@ -1363,6 +1363,39 @@ impl<S: MutinyStorage> NostrManager<S> {
         Ok(event_id)
     }
 
+    /// Creates a recommendation event for a federation
+    pub async fn recommend_federation(
+        &self,
+        invite_code: &InviteCode,
+        review: Option<&str>,
+    ) -> Result<EventId, MutinyError> {
+        let kind = Kind::from(38000);
+
+        // properly tag the event as a federation with the federation id
+        let d_tag = Tag::Identifier(invite_code.federation_id().to_string());
+        let k_tag = Tag::Generic(
+            TagKind::SingleLetter(SingleLetterTag::lowercase(Alphabet::K)),
+            vec!["38173".to_string()],
+        );
+
+        // tag the federation invite code
+        let invite_code_tag = Tag::Generic(
+            TagKind::SingleLetter(SingleLetterTag::lowercase(Alphabet::U)),
+            vec![invite_code.to_string()],
+        );
+
+        // todo tag the federation announcement event, to do so we need to have the pubkey of the federation
+
+        let builder = EventBuilder::new(
+            kind,
+            review.unwrap_or_default(),
+            [d_tag, k_tag, invite_code_tag],
+        );
+
+        // send the event
+        Ok(self.client.send_event_builder(builder).await?)
+    }
+
     /// Queries our relays for federation announcements
     pub async fn discover_federations(&self) -> Result<Vec<NostrDiscoveredFedimint>, MutinyError> {
         // get contacts by npub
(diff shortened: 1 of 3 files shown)


TonyGiorgio:
needs rebase

benthecarman:
Implemented so now we check against primal's list of trusted users.

Made it so we can see recommended federations even if they haven't announced themselves on nostr

Added the ability to recommend federations.

mutiny-core/src/nostr/mod.rs
```diff
@@ -102,6 +108,25 @@ pub struct NostrManager<S: MutinyStorage> {
     pub primal_client: PrimalClient,
 }
 
+/// A fedimint we discovered on nostr
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct NostrDiscoveredFedimint {
```
TonyGiorgio, line 113: should this be in the nostr rust library?

mutiny-core/src/nostr/mod.rs
```diff
@@ -1172,6 +1197,242 @@ impl<S: MutinyStorage> NostrManager<S> {
...
 
+    /// Creates a recommendation event for a federation
+    pub async fn recommend_federation(
+        &self,
+        invite_code: &InviteCode,
+    ) -> Result<EventId, MutinyError> {
+        let kind = Kind::from(18173);
+        // first pull down our current recommendation event, we need to do this
+        // to make sure we don't overwrite any other recommendations
+        let filter = Filter::new().kind(kind).author(self.public_key).limit(1);
```
TonyGiorgio, line 1208: it's a single replaceable event with a list of other recommendations? I fucking hate nostr lists, come on.

mutiny-core/src/nostr/mod.rs
```diff
@@ -1172,6 +1197,242 @@ impl<S: MutinyStorage> NostrManager<S> {
...
+            .into_iter()
+            .find_map(|e| if e.kind == kind { Some(e) } else { None });
+
+        // tag the federation invite code
+        let invite_code_tag = Tag::Generic(
+            TagKind::U,
+            vec![invite_code.to_string(), "fedimint".to_string()],
+        );
+
+        // todo tag the federation announcement event, to do so we need to have the pubkey of the federation
```
TonyGiorgio, line 1225: this isn't really possible?

mutiny-core/src/nostr/mod.rs
```diff
@@ -1172,6 +1197,242 @@ impl<S: MutinyStorage> NostrManager<S> {
...
+    pub async fn discover_federations(&self) -> Result<Vec<NostrDiscoveredFedimint>, MutinyError> {
+        // get contacts by npub
+        let mut npubs: HashMap<nostr::PublicKey, Contact> = self
+            .storage
+            .get_contacts()?
+            .into_iter()
+            .filter_map(|(_, c)| c.npub.map(|npub| (npub, c)))
+            .collect();
+
+        const NUM_TRUSTED_USERS: u32 = 500;
```
TonyGiorgio, line 1264: what if this was higher to like 1k? 

Also, why `const` right here? Stick it towards the top.

mutiny-core/src/nostr/primal.rs
```diff
@@ -110,4 +111,62 @@ impl PrimalClient {
 
         Ok(messages)
     }
+
+    /// Returns a list of trusted users from primal with their trust rating
+    pub async fn get_trusted_users(&self, limit: u32) -> Result<Vec<TrustedUser>, MutinyError> {
+        // fixme doesn't work with mutiny caching service
```
TonyGiorgio, line 117: is this fixed yet?

TonyGiorgio (changes requested):

mutiny-core/src/nostr/primal.rs
benthecarman: got it working :)

mutiny-core/src/nostr/mod.rs
benthecarman: spec is updated to be a single event per recommendation

mutiny-core/src/nostr/mod.rs
benthecarman: Maybe eventually, when the spec is merged

mutiny-core/src/nostr/mod.rs
TonyGiorgio: I'm a little unsure how this is supposed to work and link together. Is it just by the `d_tag` being the same across reviews? Does the federation invite code even need to be here? Should we pull down the other federation members invite code's to put in the vec?

TonyGiorgio:
Also, is this up to date with the latest of the spec?

benthecarman:
> Also, is this up to date with the latest of the spec? 

Yeah

mutiny-core/src/nostr/mod.rs
benthecarman: >Is it just by the d_tag being the same across reviews?

Yeah the d tag is the federation id

> Does the federation invite code even need to be here?

Yeah, in the case the federation has not announced itself

> Should we pull down the other federation members invite code's to put in the vec?

We could... I figured since it is a recommendation, they recommend the code they are using, maybe that is the most reliable member in the federation.

TonyGiorgio (approved):

benthecarman:
rebased

TonyGiorgio (approved):

PR #1067, opened 2024-03-18
Blindauth first pass

TonyGiorgio:


TonyGiorgio:
This could use an intermittent review @benthecarman 

There will certainly need to have a few more methods in the blind auth client, but this structure should be complete as far as creating blind messages and stuff.

mutiny-core/src/blindauth.rs
```diff
@@ -0,0 +1,332 @@
...
+/// Child ID used to derive the spend key from a service plan's DerivableSecret
+const SPEND_KEY_CHILD_ID: ChildId = ChildId(0);
+
+/// Child ID used to derive the blinding key from a service plan's DerivableSecret
+const BLINDING_KEY_CHILD_ID: ChildId = ChildId(1);
+
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
+pub struct TokenStorage {
+    // (service_id, plan_id): number of times used
+    pub map: HashMap<(u32, u32), u32>,
```
benthecarman, line 37: should make a struct for this tuple, gonna be hard to reason about otherwise

mutiny-core/src/blindauth.rs
```diff
@@ -0,0 +1,332 @@
...
+
+    let nonce = fedimint_mint_client::Nonce(spend_key.public_key());
+    let blinding_key = BlindingKey(
+        child_secret
+            .child_key(BLINDING_KEY_CHILD_ID)
+            .to_bls12_381_key(),
+    );
+    let blinded_message = blind_message(nonce.to_message(), blinding_key);
+
+    let blinded_token = NotYetBlindedToken {
```
benthecarman, line 300: `NotYetBlindedToken` doesn't really make sense here, in the line above you are doing the blinding. Something like `UnsignedToken`  would make more sense

mutiny-core/src/blindauth.rs
```diff
@@ -0,0 +1,332 @@
...
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct NotYetBlindedToken {
+    pub counter: u32,
+    pub service_id: u32,
+    pub plan_id: u32,
+    pub blinded_message: BlindedMessage,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct BlindedToken {
```
benthecarman, line 64: Then can rename this to `SignedToken`

mutiny-core/src/blindauth.rs
```diff
@@ -0,0 +1,332 @@
...
+        let token_to_blind =
+            derive_blind_token(&self.secret, service_id, plan_id, next_counter).await?;
+        let token_req = RedeemServiceTokenRequest {
+            service_id,
+            plan_id,
+            blinded_message: token_to_blind.blinded_message,
+        };
+
+        // request a blinded signature
+        let token_resp = post_redeem_tokens(&self.auth_client, &self.base_url, token_req).await?;
```
benthecarman, line 182: we shouldn't use our `auth_client` here, otherwise the request will have the JWT token and all the blind token stuff will be for nothing

benthecarman (commented):
overall looking good

mutiny-core/src/blindauth.rs
TonyGiorgio: We have to here. This is to get a blind signature from the server. In order to know if we can issue a blind signature, we need to know if the user has not been given one before.

Perhaps the name is confusing? The actual redemption of the service token isn't done here but will be done with the individual service clients. In the hermes client, the unblinding part will take place and that's what needs to be done without the auth client.

benthecarman: Ah yeah I see, the naming confused me thinking you are redeeming the tokens but you are really just getting them signed.

TonyGiorgio: gotcha, yeah i'll fix the names

mutiny-core/src/hermes.rs
```diff
@@ -0,0 +1,315 @@
...
+        let client = Client::new(signer.clone());
+
+        let relays: Vec<String> = vec![
+            "wss://relay.primal.net".to_string(),
+            "wss://relay.damus.io".to_string(),
+            "wss://nostr.mutinywallet.com".to_string(),
+            "wss://relay.mutinywallet.com".to_string(),
+        ];
+        client
+            .add_relays(relays.clone())
```
benthecarman, line 93: don't need to clone relays here

mutiny-core/src/hermes.rs
```diff
@@ -0,0 +1,315 @@
...
+            .blind_auth
+            .get_unblinded_info_from_token(available_paid_token.clone());
+        let unblinded_sig = unblind_signature(blinding_key, available_paid_token.blind_sig);
+
+        // send the register request
+        let req = RegisterRequest {
+            name: Some(name),
+            pubkey: self.public_key.to_string(),
+            federation_id: federation_identity.federation_id,
+            federation_invite_code: federation_identity.invite_code.to_string(),
```
benthecarman, line 230: the `federation_id` is a part of the invite code so in theory we don't need to send it

mutiny-core/src/nostr/mod.rs
```diff
@@ -1285,6 +1262,31 @@ impl<S: MutinyStorage> NostrManager<S> {
...
+    let index = match index {
+        Some(index) => ChildNumber::from_hardened_idx(index)?,
+        None => ChildNumber::from_normal_idx(0)?,
+    };
+
+    let path = DerivationPath::from_str(&format!("m/44'/1237'/{account}'/{chain}/{index}"))?;
+    let key = xprivkey.derive_priv(context, &path)?;
+
+    // just converting to nostr secret key, unwrap is safe
+    let secret_key = SecretKey::from_slice(&key.private_key.secret_bytes()).unwrap();
```
benthecarman, line 1286: while you're here, I think we can remove this now

benthecarman (commented):
looking good so far

mutiny-core/src/nostr/mod.rs
TonyGiorgio: Remove what? We use this in multiple places

benthecarman: we dont need this line to convert types, this was from us and rust-nostr using different versions of rust-bitcoin

TonyGiorgio: okay changed

TonyGiorgio:
Need to check for tokens after subscribing to mutiny+

benthecarman (approved):
looks like a good enough base for now, won't nit pick you to death :)

benthecarman:
Made an issue from all the TODOs in here #1079

PR #1070, opened 2024-03-19
Follow and Unfollow nostr contacts

### LINKED ISSUE #1068: Follow and Unfollow People

benthecarman:
Closes #1068

This has the base functionality needed for following and unfollowing. Tried to do my best to prevent us from deleting a follow list. ~One thing with this is there is potential race conditions if you call follow/unfollow in rapid succession. Could add a mutex around adding follows but that feels a little overkill, maybe will have to bite the bullet though~

@futurepaul I imagine you'll want some extra utility around this. I was thinking of adding `is_followed` to `TagItem`. Is there anything else I could add that you can think of?

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -42,13 +42,13 @@ pub mod vss;
 mod test_utils;
 
 use crate::cashu::CashuHttpClient;
-use crate::federation::GatewayFees;
 pub use crate::gossip::{GOSSIP_SYNC_TIME_KEY, NETWORK_GRAPH_KEY, PROB_SCORER_KEY};
 pub use crate::keymanager::generate_seed;
 pub use crate::ldkstorage::{CHANNEL_CLOSURE_PREFIX, CHANNEL_MANAGER_KEY, MONITORS_PREFIX_KEY};
 use crate::storage::{
-    list_payment_info, IndexItem, MutinyStorage, DEVICE_ID_KEY, EXPECTED_NETWORK_KEY,
-    NEED_FULL_SYNC_KEY, ONCHAIN_PREFIX, PAYMENT_INBOUND_PREFIX_KEY, PAYMENT_OUTBOUND_PREFIX_KEY,
+    list_payment_info, persist_payment_info, update_nostr_contact_list, IndexItem, MutinyStorage,
+    DEVICE_ID_KEY, EXPECTED_NETWORK_KEY, NEED_FULL_SYNC_KEY, ONCHAIN_PREFIX,
+    PAYMENT_INBOUND_PREFIX_KEY, PAYMENT_OUTBOUND_PREFIX_KEY, SUBSCRIPTION_TIMESTAMP,
 };
 use crate::{auth::MutinyAuthClient, logging::MutinyLogger};
 use crate::{error::MutinyError, nostr::ReservedProfile};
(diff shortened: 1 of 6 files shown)


mutiny-core/src/lib.rs
```diff
@@ -1062,6 +1062,24 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+                                                        // sync in background so we don't block processing other events
+                                                        let self_clone = self_clone.clone();
+                                                        utils::spawn(async move {
+                                                            match self_clone.sync_nostr_contacts(self_clone.nostr.public_key).await {
+                                                                Err(e) => log_error!(self_clone.logger, "Failed to sync nostr: {e}"),
+                                                                Ok(_) => log_debug!(self_clone.logger, "Successfully synced nostr contacts"),
+                                                            }
+                                                        });
+                                                    }
+                                                    Ok(false) => log_debug!(logger, "Got older contact list, ignoring..."),
```
TonyGiorgio, line 1080: Is there no other quicker way to have the update reflected in the contacts list than doing a full resync?

benthecarman: I guess we could diff our current list and sync the individual profiles.

TonyGiorgio: What does the sync actually do?

benthecarman: Pulls down the their profile and sets in our storage. So we have their name, pgp, lightning addr, etc.

TonyGiorgio: Gotcha. 

Couldn't we just do that for once for one profile? Instead of the whole list or trying to do a diff? It should just be adding/removing one at a time.

benthecarman: Yeah, this listener is for if a change is made outside the app.

TonyGiorgio: Oh, damn, didn't expand the top. Nevermind, that is fine I think.

TonyGiorgio (approved):
I'll defer to @futurepaul on this

mutiny-wasm/src/lib.rs
```diff
@@ -1756,6 +1756,36 @@ impl MutinyWallet {
...
+            .into_iter()
+            .map(|n| n.to_bech32().expect("bech32"))
+            .collect();
+        Ok(JsValue::from_serde(&npubs)?)
+    }
+
+    /// Follows the npub on nostr if we're not already following
+    ///
+    /// Returns true if we're now following, false if we were already following
+    pub async fn follow_npub(&self, npub: String) -> Result<bool, MutinyJsError> {
```
TonyGiorgio, line 1774: I don't really understand the purpose of these bools. Doesn't seem useful, just confusing.

benthecarman: Felt it was better to return something rather than `()`

TonyGiorgio: I don't think it matters. It either follows them or errors. Nothing else really matters.

benthecarman:
Added `is_followed` to `TagItem` and added a `get_follows_sorted` to mutiny wasm for @futurepaul

futurepaul (approved):
Sorry for the goose chase but I'm not actually using `get_follows_sorted` now. Follow / unfollow / is_followed all work great!

PR #1075, opened 2024-03-21
Fix LSP connection failures during maintance

### LINKED ISSUE #1065: Error: Failed to make a request to the LSP.
This should not freeze up on startup when the LSP is down.

benthecarman:
Fixes #1065

First commit adds a bunch of logging for us

Second commit fixes an issue i discovered while debugging. If we changed the LSP env variable then it'd set the lsp in the user's config to `null`. This would cause problems if we ever changed the default LSP as then it would break people's old wallets.

Third commit fixes the root issue, was a bug from #1018 where all the optimizations I did weren't being fully utilized because we weren't using the saved LSP config but preferring the one from env variables which did not have the connection info so we'd have to refetch it.

For simulating the LSP being down I added another domain for the lsp `signet-lsp.mutinynet.com` and would change the proxy pass port when I wanted it to be down

### DIFF
--- a/mutiny-core/src/lsp/mod.rs
+++ b/mutiny-core/src/lsp/mod.rs
@@ -124,9 +124,12 @@ pub enum AnyLsp<S: MutinyStorage> {
 }
 
 impl<S: MutinyStorage> AnyLsp<S> {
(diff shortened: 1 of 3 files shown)


TonyGiorgio (approved):
This makes sense why I've seen some logs recently that there was no LSP information available for wallets that were existing.

PR #1077, opened 2024-03-23
Use lightning-net-tokio for non-wasm builds

benthecarman:
was helping @elnosh with getting networking working for mutiny-server 

tried a bunch of things and gave up, just using `lightning-net-tokio` instead of rolling our own network impl for server context


benthecarman:
Don't merge yet, waiting for @elnosh to say everything is working for him with this

elnosh:
Thanks for the help here. It fixed some of the problems I was having but still running into some intermittent issues while starting the server so I wouldn't merge this yet. I'll keep testing this

benthecarman:
Was still getting issues so just switched to using `lightning-net-tokio` when in non-wasm

elnosh:
I've been trying now with your changes using `lightning-net-tokio` and so far no issues for me.

TonyGiorgio (approved):
Seems contained to just the server code, so ack from me if it works.

PR #1078, opened 2024-03-24
Use gloo-timers for windowless sleep

benthecarman:
This should be safe, same sleep code that fedimint is using.

It uses `setTimeout` so I believe this can be used in a web worker :rocket: 

https://crates.io/crates/gloo-timers


TonyGiorgio (approved):

TonyGiorgio:
needs rebase

PR #1084, opened 2024-03-26
(wip) mutiny server

elnosh:
this is a wip of mutiny-server

There are other endpoints I will need to add but I have basic ones working right now (open channel, create invoice, pay invoice, send to address)

I would appreciate any reviews of what I've done so far :)


mutiny-server/src/routes.rs
```diff
@@ -0,0 +1,249 @@
...
+    };
+
+    Ok(Json(json!(invoice)))
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PayInvoiceRequest {
+    amount_sat: Option<u64>,
+    invoice: String
```
benthecarman, line 169: you can just make this `invoice: Bolt11Invoice` so you don't need to parse it in the function

mutiny-server/src/sled.rs
```diff
@@ -0,0 +1,169 @@
...
+    fn cipher(&self) -> Option<Cipher> {
+        self.cipher.clone()
+    }
+
+    fn vss_client(&self) -> Option<Arc<MutinyVssClient>> {
+        None
+    }
+
+    fn activity_index(&self) -> Arc<RwLock<BTreeSet<IndexItem>>> {
+        Arc::new(RwLock::new(BTreeSet::new()))
```
benthecarman, line 68: will need to properly do this

mutiny-server/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+        .route("/newaddress", get(routes::new_address))
+        .route("/sendtoaddress", post(routes::send_to_address))
+        .route("/openchannel", post(routes::open_channel))
+        .route("/createinvoice", post(routes::create_invoice))
+        .route("/payinvoice", post(routes::pay_invoice))
+        .route("/balance", get(routes::get_balance))
+        .fallback(fallback)
+        .layer(Extension(state.clone()));
+
+    axum::serve(listener, server_router).await.unwrap();
```
benthecarman, line 78: should start the server in a thread, so we can actually wait for the shutdown hook

mutiny-server/src/extractor.rs
```diff
@@ -0,0 +1,33 @@
+use async_trait::async_trait;
+use axum::extract::{rejection::FormRejection, Request, FromRequest};
+use axum::http::StatusCode;
+use axum::Json;
+use serde_json::{json, Value};
+
+pub struct Form<T>(pub T);
```
benthecarman, line 7: why forms? is that what phoenixd is using?

mutiny-core/src/lib.rs
```diff
@@ -704,6 +704,38 @@ pub struct MutinyWalletConfig {
...
+            primal_url: None,
+            blind_auth_url: None,
+            hermes_url: None,
+            do_not_connect_peers: false,
+            skip_device_lock: false,
+            safe_mode: false,
+            skip_hodl_invoices: true,
+        }
+    }
+}
```
benthecarman, line 737: you should just be able to use `MutinyWalletConfigBuilder` in `main.rs` instead of needing to create this

benthecarman (commented):
looking good so far!

mutiny-server/src/extractor.rs
elnosh: yeah, that's what they are doing

elnosh:
Added rest of endpoints. Now just need to fix this cargo doc failing

mutiny-server/src/routes.rs
```diff
@@ -0,0 +1,526 @@
...
+
+    let fees = match invoice.fees_paid {
+        Some(fee) => fee,
+        None => {
+            return Err((
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Json(json!({"error": "unable to pay invoice"})),
+            ))
+        }
+    };
```
benthecarman, line 292: You do this style a bunch, it is cleaner to do `let fees = invoice.fees_paid.ok_or((StatusCode::INTERNAL_SERVER_ERROR, Json(json!({"error": "unable to pay invoice"})))?`

mutiny-server/src/routes.rs
```diff
@@ -0,0 +1,526 @@
...
+        .node_manager
+        .list_channels()
+        .await
+        .map_err(|e| handle_mutiny_err(e))?;
+    let channels = channels
+        .into_iter()
+        .map(|channel| {
+            let state = match channel.is_usable {
+                true => "usable",
+                false => "unusable",
```
benthecarman, line 462: Can you make this into an `enum`?

benthecarman (commented):
Mostly looks good to me just some style things. 

I think we will need to properly figure out the issue you had with the docs CI. When I try to run `cargo check` in the root of `mutiny-node` now I get issues. Probably will need to figure out correct ways to handle the per-package-target stuff

mutiny-server/src/routes.rs
```diff
@@ -305,11 +305,7 @@ pub async fn get_incoming_payments(
 
     let mut payments = Vec::with_capacity(invoices.len());
     for invoice in invoices.into_iter() {
-        let pr = match invoice.bolt11.clone() {
-            Some(bolt_11) => Some(bolt_11.to_string()),
-            None => None,
-        };
-
+        let pr = invoice.bolt11.clone().map(|bolt11| bolt11.to_string());
```
benthecarman, line 308: you don't need this clone, can do `as_ref()` instead

mutiny-server/src/routes.rs
```diff
@@ -343,12 +339,9 @@ pub async fn get_payment(
...
         .get_invoice_by_hash(&hash)
         .await
-        .map_err(|e| handle_mutiny_err(e))?;
+        .map_err(handle_mutiny_err)?;
 
-    let pr = match invoice.bolt11.clone() {
-        Some(bolt11) => Some(bolt11.to_string()),
-        None => None,
-    };
+    let pr = invoice.bolt11.clone().map(|bolt11| bolt11.to_string());
```
benthecarman, line 344: same here

mutiny-server/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "mutiny-server"
+version = "0.3.7"
```
benthecarman, line 3: can you bump to 0.6.4

mutiny-server/src/main.rs
```diff
@@ -0,0 +1,103 @@
...
+use std::time::Duration;
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    pretty_env_logger::try_init()?;
+    let mut shutdown = Shutdown::new().unwrap();
+    let config: Config = Config::parse();
+
+    let network = config.network();
+    let storage = SledStorage::new(&config.db_file, config.clone().password)?;
```
benthecarman, line 28: can instead do `config.password.clone()` here, that way you don't need to clone the entire config

mutiny-server/src/routes.rs
```diff
@@ -0,0 +1,486 @@
...
+        .await
+        .map_err(handle_mutiny_err)?;
+
+    let channels = channels
+        .into_iter()
+        .map(|channel| {
+            let state = match channel.is_usable {
+                true => ChannelState::Usable,
+                false => ChannelState::Unusable,
+            };
```
benthecarman, line 436: instead of all this, can just have `ChannelState` impl `Serialize`, maybe just need to add the serde flag for all lowercase

benthecarman (commented):
couple nits

Can you also add some commands to the just file? Make sure the clippy commands will run clippy for mutiny-server. Also would be good to add a somehting like `just run-signet` that will run mutiny server on signet with all the required params

TonyGiorgio:
Would like to figure out how to get the workspaces to work correctly. I can look into that more. Skimmed the structure and looks pretty sound. Will play with it more soon and can merge whenever the base looks good, can then can keep iterating on the commands. Nice work so far!

TonyGiorgio:
Actually, I think we should use rocksdb instead of sleddb.

PR #1085, opened 2024-03-26
Fixes for blind tokens

benthecarman:
- We couldn't ever write the tokens to storage because it was trying to use `ServicePlanIndex` as a key for the json object. Json object's keys need to be strings so I manually implemented serde for it doing it just as `x-y`
- Added implementation for reading tokens from VSS


TonyGiorgio (approved):
I still think it was better to leave it as (u32, u32) but I guess this works too.

PR #1089, opened 2024-03-26
Better handling for marking DM'd invoices as paid

### LINKED ISSUE #1087: Invoices paid in DMs won't clear the "Pending (1)" notification.
Screenshot 2024-03-26 at 4 45 51 PM

Screenshot 2024-03-26 at 4 45 46 PM

benthecarman:
Closes #1087

First commit is pretty straight forward, makes it so in our startup check instead of removing only expired invoices, we also check for paid ones

Second commit I don't love but not sure the best way to do this otherwise. This attempts to remove any payment we make from our pending nwc. I am not really sure a better way to handle this besides making the frontend instead call `approve_invoice` instead of `pay_invoice` for invoices in our dm's but that doesn't seem the best either.

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -1331,6 +1331,17 @@ impl<S: MutinyStorage> MutinyWallet<S> {
                         .await;
                     match payment_result {
                         Ok(r) => {
+                            // spawn a task to remove the pending invoice if it exists
+                            let nostr_clone = self.nostr.clone();
+                            let payment_hash = *inv.payment_hash();
+                            let logger = self.logger.clone();
+                            utils::spawn(async move {
+                                if let Err(e) =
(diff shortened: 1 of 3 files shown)


TonyGiorgio:
> making the frontend instead call approve_invoice instead of pay_invoice

Actually, that doesn't seem that paid of an idea. This works too I guess but this is going to be more of a performance requirement on every sent payment. There's going to be more sent payments than invoices in DMs.

benthecarman:
> Actually, that doesn't seem that paid of an idea

that's what i tried to do first, the problem is that the front end takes you to the send screen so we'd need to keep the state of which function to call throughout. Also we don't have a guarantee that it is actually in the pending nwc list, the user could reject it in the list but then in DMs they could pay it.

TonyGiorgio:
Yeah this is probably fine at least for now. Odds are there's not really any pending requests anyways for very long.

TonyGiorgio (approved):

Issue #1079, opened 2024-03-25
Blind Auth Follow Ups


benthecarman:
- [x] Check for new tokens after mutiny+ subscription
- [x] Proper DM listening handling
- [ ] save our lightning address in storage

follow up: 
Add change federation functionality

futurepaul:
is there a way I can get a better error for if a name is already taken?

<img width="250" alt="Screenshot 2024-03-26 at 4 32 01 PM" src="https://github.com/MutinyWallet/mutiny-node/assets/543668/6fa8fed5-79c0-45df-b557-caf3f6a5e3f6">

TonyGiorgio:
The check_name API call should be called first, and it returns a boolean value.

futurepaul:
> The check_name API call should be called first, and it returns a boolean value.

ah nice

TonyGiorgio:
> save our lightning address in storage

I've been thinking about how to do this. There's the naive approach of storing to local storage what the LN address should be. But what I'm thinking is that it can be a query to the hermes server. 

We already know what the deterministic nostr pubkey will be for the user, so it can create a signed message from that nostr pubkey to something like `/check-address` which returns the federation id and the name for the user. That's the source of truth afterall. Anything we did locally would need to have some sort of cross check against the server. 

Then from there, we can have federation changes happen the same way. We can see if the user's defined federation is different than the one stored on hermes. If so, then it creates a signed nostr message and then makes a request to `/change-federation` with the invite code it wants to change to. 

Lastly (for a follow up some other day), to change names it can follow the same federation change flow, but for `/change-name` and it maybe just gets an invoice back that it has to pay to change the name?

benthecarman:
That makes sense to me. I think we should try to do some local caching so we don't need to hit the server every time we boot up

TonyGiorgio:
What kind of caching do you think would be best here? 

We can do something like a daily cache? Weekly? 

Store it in local storage but not needed to be remotely backed up? If it's missing in local storage then it fetches remotely. 

My main worry is from switching devices and not having the latest state if one of the devices changes federations or does the name change.

benthecarman:
If we just store it in VSS then it should be fine

TonyGiorgio:
> If we just store it in VSS then it should be fine

I want to avoid storing in VSS just to store in VSS.

PR #1090, opened 2024-03-26
Claim hermes tokens

benthecarman:
Part of #1079

### DIFF
--- a/mutiny-core/src/federation.rs
+++ b/mutiny-core/src/federation.rs
@@ -14,7 +14,7 @@ use crate::{
 use async_lock::RwLock;
 use async_trait::async_trait;
 use bip39::Mnemonic;
-use bitcoin::secp256k1::ThirtyTwoByteHash;
+use bitcoin::secp256k1::{SecretKey, ThirtyTwoByteHash};
 use bitcoin::{
     bip32::{ChildNumber, DerivationPath, ExtendedPrivKey},
     hashes::sha256,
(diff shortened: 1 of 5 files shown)


mutiny-core/src/federation.rs
```diff
@@ -636,13 +695,7 @@ impl<S: MutinyStorage> FederationClient<S> {
...
 
     pub async fn recover_backup(&self) -> Result<(), MutinyError> {
-        self.fedimint_client
-            .restore_from_backup(None)
-            .await
-            .map_err(|e| {
-                log_error!(self.logger, "Could not restore from backup: {}", e);
-                e
-            })?;
+        // todo this now a part of init
```
TonyGiorgio, line 698: What do you mean by this? Init should not be trying to call the restore functions.

mutiny-core/src/storage.rs
```diff
@@ -549,15 +550,26 @@ pub trait MutinyStorage: Clone + Sized + Send + Sync + 'static {
...
+        let key = if is_hermes {
+            LAST_HERMES_SYNC_TIME_KEY
+        } else {
+            LAST_DM_SYNC_TIME_KEY
+        };
+        self.get_data(key)
     }
 
-    fn set_dm_sync_time(&self, time: u64) -> Result<(), MutinyError> {
+    fn set_dm_sync_time(&self, time: u64, is_hermes: bool) -> Result<(), MutinyError> {
```
TonyGiorgio, line 562: it would be better to just make this an alternative key anyways. If we do more things with alternative nostr key DMs, which I imagine we will soon, then this boolean won't make very much sense.

mutiny-core/src/lib.rs
```diff
@@ -1066,7 +1070,13 @@ impl<S: MutinyStorage> MutinyWalletBuilder<S> {
...
 
         // start the hermes background process
-        mw.start_hermes()?;
+        // get profile key if we have it, we need this to decrypt private zaps
+        let profile_key = match &mw.nostr.primary_key {
+            NostrSigner::Keys(keys) => Some(keys.clone()),
+            #[cfg(target_arch = "wasm32")]
+            NostrSigner::NIP07(_) => None,
+        };
+        mw.start_hermes(profile_key)?;
```
TonyGiorgio, line 1079: We shouldn't be using the user's nostr key for this.

TonyGiorgio (changes requested):

mutiny-core/src/lib.rs
benthecarman: It's for decrypting private zaps, not for the actual dms

mutiny-core/src/storage.rs
benthecarman: Yeah I figured we'd just switch to an enum, I can do that now I guess

mutiny-core/src/federation.rs
benthecarman: There's no longer the `restore_from_backup` function, instead you need to call it when building the client.

TonyGiorgio: Okay then it can be deleted then. Not used in mutiny web.

mutiny-core/src/storage.rs
TonyGiorgio: That's fine, no need to block on it.

TonyGiorgio (approved):

TonyGiorgio:
Is this good to merge? It's still in draft.

mutiny-core/src/hermes.rs
```diff
@@ -306,12 +316,180 @@ async fn register_name(
...
+            }
+            Err(e) => log_error!(logger, "Error claiming external receive: {e}"),
+        }
+    } else {
+        log_warn!(
+            logger,
+            "Received DM for unknown federation {}",
+            notification.federation_id
+        );
+        // todo should we save this to storage?
```
TonyGiorgio, line 461: Save what in storage? The failed DM?

TonyGiorgio (approved):
nit on a random comment, otherwise gtg

PR #1092, opened 2024-03-27
Assume mainnet for fedimint recommendations

benthecarman:
In the last PR I only switched it to assume mainnet for the federations, didn't do it for the recommendations.


TonyGiorgio:
Why do the recommendations need that at all? It should just be for the mint event?

benthecarman:
Because there is no guarantee there is a mint event

TonyGiorgio (approved):

PR #1098, opened 2024-04-01
Replace instant dependency with web-time

DanGould:
web-time offers a number of advantages over the instant crate

- Actively maintained
- Stable 1.x API
- Exact std::time  replacement
- Compatible with rustls pki-types web feature
- Re-exports std::time when not targeting wasm32-unknown-unknown without pulling in any dependencies

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2525,7 +2525,6 @@ dependencies = [
  "getrandom",
  "gloo-net",
  "hex-conservative",
- "instant",
  "itertools 0.11.0",
  "js-sys",
  "jwt-compact",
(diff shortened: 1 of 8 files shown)


benthecarman (changes requested):
i'd rather use `std::time` when we aren't on wasm32. no need to use this dep in those instances

DanGould:
This crate is just an alias for `std::time` when we aren't on wasm32-unknown-unknown. This dep will not be used in those instances thanks to rust's dead code elimination.

See: https://github.com/daxpedda/web-time/blob/main/src/lib.rs

benthecarman:
Yeah but then we need to trust that dep vs just doing less of 10 lines of code to handle it ourself

benthecarman (approved):
lgtm

DanGould:
Thanks for the prompt review

PR #1105, opened 2024-04-03
Check LNURL name

TonyGiorgio:
Part of #1079

### DIFF
--- a/mutiny-core/src/hermes.rs
+++ b/mutiny-core/src/hermes.rs
@@ -64,6 +64,8 @@ pub struct HermesClient<S: MutinyStorage> {
     pub(crate) federations: Arc<RwLock<HashMap<FederationId, Arc<FederationClient<S>>>>>,
     blind_auth: Arc<BlindAuthClient<S>>,
     base_url: String,
+    // bool represents whether or not it was successfully checked
+    current_address: Arc<RwLock<(Option<String>, bool)>>,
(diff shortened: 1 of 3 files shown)


mutiny-core/src/hermes.rs
```diff
@@ -120,25 +120,58 @@ impl<S: MutinyStorage> HermesClient<S> {
...
+                    &http_client_check_clone,
+                    &base_url_check_clone,
+                    public_key_check_clone,
+                )
+                .await
+                {
+                    Ok(o) => {
+                        let mut c = current_address_check_clone.write().await;
+                        log_info!(logger_check_clone, "checked lightning address: {o:?}");
+                        *c = o;
```
benthecarman, line 157: need to break here

mutiny-core/src/lib.rs
```diff
@@ -2774,6 +2774,14 @@ impl<S: MutinyStorage> MutinyWallet<S> {
         }
     }
 
+    pub async fn check_lnurl_name(&self) -> Result<Option<String>, MutinyError> {
+        if let Some(hermes_client) = self.hermes_client.clone() {
```
benthecarman, line 2778: can use `as_ref` instead of clone

benthecarman (approved):

futurepaul (approved):
works on my machine

PR #1125, opened 2024-04-09
Better setup for new profiles

benthecarman:


benthecarman:
todo expose to mutiny-web

TonyGiorgio (approved):
ACK for me but needs @futurepaul's approval.

futurepaul:
so I would just need to update the calls in `NewProfile.tsx` https://github.com/MutinyWallet/mutiny-web/blob/cdc7c880fa5fa4cdaf52bd7c0471cf8c817de2c5/src/routes/setup/NewProfile.tsx to use this? And pass in nothing in the handleSkip case? not sure what it simplifies from the frontend perspective otherwise

benthecarman:
> so I would just need to update the calls in NewProfile.tsx [MutinyWallet/mutiny-web@cdc7c88/src/routes/setup/NewProfile.tsx](https://github.com/MutinyWallet/mutiny-web/blob/cdc7c880fa5fa4cdaf52bd7c0471cf8c817de2c5/src/routes/setup/NewProfile.tsx) to use this? And pass in nothing in the handleSkip case? not sure what it simplifies from the frontend perspective otherwise

yeah pretty much, think you should still pass in the `DEFAULT DEFAULT_NOSTR_NAME` just for consistency

futurepaul (approved):
works for me https://github.com/MutinyWallet/mutiny-web/pull/1031

PR #1127, opened 2024-04-10
Implement deleting federation recommendations

benthecarman:
Uses [NIP-09](https://github.com/nostr-protocol/nips/blob/master/09.md) event deletions, this is what bitcoinmints.com uses.

Tbh not the biggest fan of event deletions but this works for now

### DIFF
--- a/mutiny-core/src/nostr/mod.rs
+++ b/mutiny-core/src/nostr/mod.rs
@@ -24,6 +24,7 @@ use lightning::{log_debug, log_error, log_info, log_warn};
 use lightning_invoice::Bolt11Invoice;
 use lnurl::lnurl::LnUrl;
 use nostr::nips::nip47::*;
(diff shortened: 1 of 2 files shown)


mutiny-core/src/nostr/mod.rs
```diff
@@ -1644,19 +1645,50 @@ impl<S: MutinyStorage> NostrManager<S> {
     }
 
     /// Checks if we have recommended the given federation
-    pub async fn has_recommended_federation(
+    async fn get_my_recommendation_events(
```
TonyGiorgio, line 1648: comment is wrong.

What's the purpose of this?

TonyGiorgio: Nevermind, I see. Not exposed, just to refactor.

TonyGiorgio (changes requested):
nit on the comment

TonyGiorgio (approved):

Issue #1093, opened 2024-03-27
Write tests for all the new stuff


benthecarman:
- [x] Fedimint discovery
- [x] Fedimint recommendations
- [x] follow/unfollow
- [x] creating/editing profile
- [x] changed nostr keys handling
- [x] blind auth
- [x] hermes registration
- [x] hermes dm listening  

We've been manually testing all these, would be good to get in proper tests so we don't break in the future

PR #1128, opened 2024-04-10
Tests for lots new nostr functionality

benthecarman:
For #1093

Required some minor changes for testing to make sure we don't broadcast events.

Did find one small bug where we weren't saving our contact list locally when calling `setup_new_profile`.

### DIFF
--- a/mutiny-core/src/nostr/mod.rs
+++ b/mutiny-core/src/nostr/mod.rs
@@ -2920,4 +2920,42 @@ mod test {
         assert!(ben.image_url.is_some());
         assert!(!ben.name.is_empty());
     }
+
+    #[tokio::test]
+    async fn test_create_recommendation_event() {
+        let mut nostr_manager = create_nostr_manager().await;
+        nostr_manager
+            .client
+            .expect_sign_event_builder()
+            .once()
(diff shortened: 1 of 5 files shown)


mutiny-core/src/nostr/mod.rs
```diff
@@ -473,15 +477,28 @@ impl<S: MutinyStorage> NostrManager<S> {
...
+                .nostr_keys
+                .read()
+                .await
+                .signer
+                .sign_event_builder(builder)
+                .await?;
+
+            // don't broadcast if we're testing
+            #[cfg(not(test))]
+            self.client.send_event(event.clone()).await?;
```
TonyGiorgio, line 490: I don't like these, there's no reason it should have errors in CI doing this.

Especially if you're already doing mock stuff anyways. Either let them broadcast or mock them out, but don't comment out the code for the sake of tests.

TonyGiorgio (changes requested):

mutiny-core/src/nostr/mod.rs
TonyGiorgio: still gotta remove them

TonyGiorgio (approved):

PR #1130, opened 2024-04-11
Hermes address change

TonyGiorgio:


mutiny-core/src/hermes.rs
```diff
@@ -333,8 +385,38 @@ impl<S: MutinyStorage> HermesClient<S> {
...
+    let signer = nostr_client.signer().await?;
+    let event_builder = EventBuilder::new(
+        NEW_FEDERATION_EVENT_KIND,
+        federation.invite_code.to_string(),
+        [],
+    );
+    let event = signer.sign_event_builder(event_builder).await?;
+
+    // send request
+    let url = Url::parse(&format!("{}/v1/change-name", base_url))
```
benthecarman, line 412: wrong url

mutiny-core/src/hermes.rs
```diff
@@ -148,26 +151,62 @@ impl<S: MutinyStorage> HermesClient<S> {
...
+                                o.name
+                            );
+                            *c = (o.name.clone(), true);
+                        }
+
+                        // check that federation is still the same if the user has a name
+                        if o.name.is_some() {
+                            if let Some(f) = first_federation {
+                                // user has a federation, set the current one as the address
+                                match change_federation_info(
```
benthecarman, line 186: This is going to be called on every start up, should only call `change_federation_info` if the federation ids don't match

TonyGiorgio: had that until a did a little refactoring and missed it. added

benthecarman (approved):
out with one todo, in with another

mutiny-core/src/hermes.rs
```diff
@@ -144,35 +148,86 @@ impl<S: MutinyStorage> HermesClient<S> {
...
+                            log_info!(
+                                logger_check_clone,
+                                "checked lightning address: {:?}",
+                                o.name
+                            );
+                            *c = (o.name.clone(), true);
+                        }
+
+                        // check that federation is still the same
+                        if let Some(f) = first_federation {
```
benthecarman, line 184: i think we still want to do these checks only if the user has a name

benthecarman: oh i see you check in `disable_zaps`

mutiny-core/src/lib.rs
```diff
@@ -2451,6 +2452,16 @@ impl<S: MutinyStorage> MutinyWallet<S> {
             return Err(MutinyError::NotFound);
         }
 
+        // remove the federation from hermes
+        if let Some(h) = self.hermes_client.clone() {
```
benthecarman, line 2456: nit: as_ref()

mutiny-core/src/hermes.rs
```diff
@@ -338,8 +419,77 @@ impl<S: MutinyStorage> HermesClient<S> {
...
+    base_url: &str,
+    nostr_client: Client,
+    current_address: Arc<RwLock<(Option<String>, bool)>>,
+    federation: FederationIdentity,
+    logger: &MutinyLogger,
+) -> Result<(), MutinyError> {
+    // make sure name is registered already
+    if current_address.read().await.0.is_none() {
+        return Ok(());
+    }
```
benthecarman, line 441: This and the one for `disable_zaps` has a minor race condition if the user manages to remove/switch federation before the check call has finished. Could do rechecking until `.1` is true

TonyGiorgio:
need to see if the user is enabled or disabled with zaps to know if we need to call disable-zaps on init if they do not have a federation. other wise it keeps calling it even if the server already knows that.

benthecarman (approved):

PR #1132, opened 2024-04-11
log when we prune a monitor

benthecarman:
May be the cause of https://github.com/MutinyWallet/mutiny-web/issues/1032, this would make it easier to debug


TonyGiorgio:
Wrong issue number?

TonyGiorgio (approved):

PR #1135, opened 2024-04-12
handle dms with an invoice along with message

elnosh:


benthecarman (changes requested):
otherwise lgtm

mutiny-core/src/nostr/mod.rs
```diff
@@ -1443,23 +1443,51 @@ impl<S: MutinyStorage> NostrManager<S> {
...
-                log_debug!(self.logger, "Not adding DM'd invoice: {msg}");
-                return Ok(());
-            }
-        };
+        // loop through dm to check for invoice
+        for word in decrypted.split_whitespace() {
+            let invoice_request_param = match bitcoin_waila::PaymentParams::from_str(word) {
+                // if not an invoice, go to next word in dm
+                Ok(param) => {
+                    match param {
```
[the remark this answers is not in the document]
elnosh: ahh nice, thanks

mutiny-core/src/nostr/mod.rs
```diff
@@ -1443,23 +1443,36 @@ impl<S: MutinyStorage> NostrManager<S> {
...
-                log_debug!(self.logger, "Not adding DM'd invoice: {msg}");
-                return Ok(());
-            }
-        };
+        // loop through dm to check for invoice
+        for word in decrypted.split_whitespace() {
+            let invoice_request_param = match bitcoin_waila::PaymentParams::from_str(word) {
+                // if not an invoice, go to next word in dm
+                Ok(param) => match param.invoice() {
+                    Some(invoice) => PayInvoiceRequestParams {
```
TonyGiorgio, line 1451: I really don't like the idea of trying to parse every single word as an invoice. I think you can just look for `lnc` as a word prefix first.

benthecarman: this wouldn't work with ones with the `lightning` prefix for bip21s. Maybe a better way is to filter out strings with a length less than 15

elnosh: I could ignore any string that doesn't start with an 'L' or 'b' ?

TonyGiorgio: that's still going to include a lot of words.

benthecarman (approved):

TonyGiorgio (approved):

PR #1142, opened 2024-04-16
Make NodeManager optional

### LINKED ISSUE #1071: Deadlock in `NodeManager::new_node`
Was debugging with @elnosh on something with the mutiny-server stuff. Seems like when we attempt to create 2 new nodes at the same time a deadlock occurred. Not sure if this is from the `async-lock` crate or something else, but since we only ever really use one node it is not a huge priority

benthecarman:
Was able to fix #1071 while i was at it

This puts the node manager is a RwLock is by default off, only until `new_node` is called will we have a lightning node and on-chain wallet. Otherwise it will use fedimint when there is no nodemanager or throw an error if neither are configured.

Still WIP as I would like to test more and this requires on-chain fedimint to exist first


benthecarman:
This is getting close, want to write tests for everything but in my manual testing things seem to be working well

mutiny-core/src/lib.rs
```diff
@@ -1363,6 +1377,85 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
+        // if we don't have a node manager, return
+        if lock.is_none() {
+            return Ok(());
+        }
+
+        let node_manager = lock.as_ref().expect("should have a node manager");
+
+        if !node_manager.get_balance().await?.is_zero() {
+            return Err(MutinyError::InvalidArgumentsError);
+        }
```
TonyGiorgio, line 1437: I think it should probably do more checks than just this. I feel like it should be as strict as possible. Stuff like no channels, no pending closes, perhaps as far as to say never having a channel. 

I can imagine what the purpose of `remove_node_manager` is, and I think we shouldn't ever take the node away from someone that has already used their node.

benthecarman: Yeah I agree there, it is scary. My hope is that we can get people who have been fedimint only to click this, also I think it'd good to have for people who add a node manager just because they are exploring the app, they can get rid of it as well.

Do note, this function also calls `archive_node` which does the same check that we do for pruning monitors, if any monitors are not prunable, it will fail to archive. I'll add a comment so it is more clear when reading this function

benthecarman:
note to self: need to make sure bringing back a node manager doesn't create a new node and instead brings back the old nodes

mutiny-core/src/lib.rs
TonyGiorgio: We could probably just call this immediately on everyone that has never had a node with channels.

benthecarman: > We could probably just call this immediately on everyone that has never had a node with channels.

Added that and made the checks stronger so we should always unarchive the first node when coming back rather than always creating a new one.

Will start working on unit tests next.

TonyGiorgio:
needs rebase

TonyGiorgio:
Errors when receiving on chain when only a federation is added

Screenshot 2024-06-06 at 8 01 02 PM

TonyGiorgio (changes requested):
needs further testing, on chain receives don't work at all

benthecarman:
Went through full testing guide, everything is working as of now. Going to add a few more things to it that will now apply with this

benthecarman:
discussed in meeting: don't default remove empty node managers

PR #1151, opened 2024-04-19
Maybe net fixes

benthecarman:
Noticed that our sleep timer when using `lightning_net_tokio` differed from ldk-sample so I fixed that.

This lead me to think maybe our stop checker for causing issues, tried removing that and it _seems_ to help


TonyGiorgio:
Nack

PR #1153, opened 2024-04-22
liquidity alpha3

johncantrell97:
This has the final set of serialization breaking changes for LSPS0, LSPS1, and LSPS2 as the specs are now finalized and no more breaking changes will be allowed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2301,9 +2301,9 @@ dependencies = [
 
 [[package]]
 name = "lightning-liquidity"
-version = "0.1.0-alpha.2"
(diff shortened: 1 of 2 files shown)


benthecarman (approved):

benthecarman:
is this safe to update, will it break compat for you?

johncantrell97:
yeah we'd love to upgrade in-sync with you but would prefer it to be "on us to upgrade" to fix the compatibility so deploy whenever you want. a heads up would be nice though :)

benthecarman:
Okay I'll let you guys know when we do

PR #1155, opened 2024-04-23
nwc fixups

johncantrell97:
A couple small fixes for NWC:

- Return error event for events sent to inactive profiles. Even though you mentioned it's not something that users should run into since you delete profiles instead of archive, the logic to skip events is still there.  Would be best to respond so the user who made the request can get some feedback about what is going on.  At a minimum should log the error.

- If create_invoice fails it would be nice to respond with an error to the user requesting the invoice.

### DIFF
--- a/mutiny-core/src/nostr/nwc.rs
+++ b/mutiny-core/src/nostr/nwc.rs
@@ -421,10 +421,7 @@ impl NostrWalletConnect {
         let mut needs_save = false;
         let mut needs_delete = false;
         let mut result = None;
-        if self.profile.active()
-            && event.kind == Kind::WalletConnectRequest
-            && event.pubkey == client_pubkey
-        {
+        if event.kind == Kind::WalletConnectRequest && event.pubkey == client_pubkey {
             let server_key = self.server_key.secret_key()?;
 
             let decrypted = decrypt(server_key, &client_pubkey, &event.content)?;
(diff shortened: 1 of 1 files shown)


mutiny-core/src/nostr/mod.rs
```diff
@@ -292,8 +292,9 @@ impl<S: MutinyStorage, P: PrimalApi, C: NostrClient> NostrManager<S, P, C> {
         let time_stamp = match self.storage.get_nwc_sync_time()? {
             None => {
                 let now = Timestamp::now();
-                self.storage.set_nwc_sync_time(now.as_u64())?;
-                now
+                let since_last_check = Timestamp::from(now.as_u64() - 20);
```
benthecarman, line 295: This can cause us to double reply to events. Would be better if we only changed the timestamp of the new filter. 

Also you add comments for this, otherwise will be confusing

johncantrell97: Yeah, I originally extracted it as a constant but was annoying because sleep() needs an i32 for some reason (how do you sleep for negative amount of time?!).

Oh, maybe I misunderstand this then, I thought the idea was that you have _never_ synced before at all, so you could not get double events.

johncantrell97: Ok yeah, I see. This same get_filters() is called during the regular check for new filters as well. 

Isn't it still true though that if you've never synced then this is safe?

It does mean though for _new_ filters added they would have the last_sync + 1 timestamp when really you'd want the buffered timestamp.

I don't really see an _easy_ way to detect new ones.  I guess I could pass in Option<Vec<Filter>> of last_filters and compare the lists similar to how you do it in `utils::compare_filters_vec`

johncantrell97: i just removed this commit for now, more complicated then I originally thought.

benthecarman: yeah i was thinking some in memory map of new ones but is a bit overly complicated

benthecarman (approved):

Issue #1138, opened 2024-04-12
Set `last_updated` when fedimint payments complete


benthecarman:
Currently when federation payments finish we update the status but not the `last_updated` field. This can cause for some activity items to appear out of order. Especially can look weird in DMs. One potentially hard thing here is fedimint payments can complete while the app is not open so just calling `utils::now()` for the time could cause more confusion than the current version, not sure if we're able to get the timestamp of events from fedimint.

PR #1157, opened 2024-04-24
Better fedimint update streams

TonyGiorgio:
This focuses more on streams than continuous polling. Quite a few TODOs and cleanups needed but it functionally works. Should be able to finish tomorrow.

Fixes https://github.com/MutinyWallet/mutiny-node/issues/1138

### DIFF
--- a/mutiny-core/src/federation.rs
+++ b/mutiny-core/src/federation.rs
@@ -1,4 +1,4 @@
-use crate::utils::{convert_from_fedimint_invoice, convert_to_fedimint_invoice, spawn};
+use crate::utils::{convert_from_fedimint_invoice, convert_to_fedimint_invoice, now, spawn};
 use crate::{
     error::{MutinyError, MutinyStorageError},
     event::PaymentInfo,
@@ -730,7 +730,8 @@ fn maybe_update_after_checking_fedimint<S: MutinyStorage>(
             log_debug!(logger, "Saving updated payment");
             let hash = updated_invoice.payment_hash.into_32();
             let inbound = updated_invoice.inbound;
-            let payment_info = PaymentInfo::from(updated_invoice);
+            let mut payment_info = PaymentInfo::from(updated_invoice);
+            payment_info.last_update = now().as_secs();
             persist_payment_info(&storage, &hash, &payment_info, inbound)?;
         }
         HTLCStatus::Pending | HTLCStatus::InFlight => (),
--- a/mutiny-core/src/federation.rs
+++ b/mutiny-core/src/federation.rs
@@ -24,7 +24,6 @@ use bitcoin::{
 use core::fmt;
 use fedimint_bip39::Bip39RootSecretStrategy;
 use fedimint_client::{
-    db::ChronologicalOperationLogKey,
     derivable_secret::DerivableSecret,
     oplog::{OperationLogEntry, UpdateStreamOrOutcome},
     secret::{get_default_client_secret, RootSecretStrategy},
@@ -55,7 +54,7 @@ use fedimint_ln_common::lightning_invoice::{Bolt11InvoiceDescription, Descriptio
 use fedimint_ln_common::{LightningCommonInit, LightningGateway};
 use fedimint_mint_client::MintClientInit;
 use fedimint_wallet_client::{WalletClientInit, WalletClientModule};
-use futures::future::{self};
+use futures::{select, FutureExt};
 use futures_util::{pin_mut, StreamExt};
 use hex_conservative::{DisplayHex, FromHex};
 use lightning::{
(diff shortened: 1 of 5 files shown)


mutiny-core/src/federation.rs
```diff
@@ -587,50 +632,89 @@ impl<S: MutinyStorage> FederationClient<S> {
...
-            .await?;
+        self.maybe_update_after_checking_fedimint(inv.clone())?;
 
         match inv.status {
             HTLCStatus::Succeeded => Ok(inv),
             HTLCStatus::Failed => Err(MutinyError::RoutingFailed),
-            HTLCStatus::Pending => Err(MutinyError::PaymentTimeout),
-            HTLCStatus::InFlight => Err(MutinyError::PaymentTimeout),
+            _ => {
+                // keep streaming after timeout happens
```
TonyGiorgio, line 679: i should call subscribe_operation here instead of copying and pasting

mutiny-core/src/federation.rs
```diff
@@ -917,13 +1068,25 @@ where
...
+                }
+                    }
+                    _ = delay_fut => {
+                    if timeout.is_none() {
+                        // TODO check sleep time to determine break
+                        log_debug!(
+                            logger,
+                            "Still waiting for stream to finish {}",
+                            invoice.payment_hash
+                        );
```
TonyGiorgio, line 1080: need to add the stop channel to `federation` and check it here. Also this log is too noisy.

TonyGiorgio:
Two minor things but otherwise ready for a first pass.

mutiny-core/src/federation.rs
```diff
@@ -317,15 +321,122 @@ impl<S: MutinyStorage> FederationClient<S> {
...
             fedimint_client,
             fedimint_storage,
             storage,
             logger,
             invite_code: federation_code,
             gateway,
-        })
+        };
+
+        federation_client.process_previous_operations().await?;
```
benthecarman, line 335: can do we do this in a thread so we don't block startup on this?

mutiny-core/src/federation.rs
```diff
@@ -794,6 +897,46 @@ pub(crate) fn mnemonic_from_xpriv(xpriv: ExtendedPrivKey) -> Result<Mnemonic, Mu
...
+        LightningOperationMetaVariant::Receive { invoice, .. } => {
+            let invoice = convert_from_fedimint_invoice(&invoice);
+            if invoice.payment_hash().into_32() == hash {
+                match lightning_module.subscribe_ln_receive(operation_id).await {
+                    Ok(o) => Some(
+                        process_outcome(o, process_receive_state, invoice, true, None, logger)
+                            .await,
+                    ),
+                    Err(_) => Some(invoice.into()),
+                }
```
benthecarman, line 931: Why return `Some(invoice)` on `Err` here?

mutiny-core/src/federation.rs
```diff
@@ -707,6 +791,25 @@ impl<S: MutinyStorage> FederationClient<S> {
...
+            updated_invoice.status,
+            HTLCStatus::Succeeded | HTLCStatus::Failed
+        ) {
+            log_debug!(logger, "Saving updated payment");
+            let hash = updated_invoice.payment_hash.into_32();
+            let inbound = updated_invoice.inbound;
+            let payment_info = PaymentInfo::from(updated_invoice);
+            persist_payment_info(&storage, &hash, &payment_info, inbound)?;
+        },
+    )
```
benthecarman, line 810: this `Ok()` around everything is a weird pattern.

Also doing `if matches!` can just be a `match` statement

mutiny-core/src/federation.rs
```diff
@@ -902,13 +1045,21 @@ where
...
+                    sleep(1_000 as i32)
+                };
+
+                let mut stream_fut = Box::pin(s.next()).fuse();
+                let delay_fut = Box::pin(timeout_future).fuse();
+                pin_mut!(delay_fut);
+
+                select! {
+                    outcome_option = stream_fut => {
                 if let Some(outcome) = outcome_option {
```
benthecarman, line 1063: formatting looks weird here

benthecarman (commented):
In `get_invoice_by_hash` we have it look up in the fedimint operation logs as well. Do we still need that? Now it should be subscribed for every invoice so we can just trust storage.

benthecarman:
would be good to fix #1138 here

mutiny-core/src/federation.rs
TonyGiorgio: So we just get the latest state of the invoice in general, wherever it ended up as.

benthecarman: Can you add a comment then

benthecarman (approved):

PR #1158, opened 2024-04-24
Hermes Tests

benthecarman:
Did a small refactor so testing the hermes claiming is simpler, just moving some code into its own function.

Otherwise adds a bunch of tests for handling zaps from hermes. Did find a niche bug with private zaps and using a NIP-07 extension, it would mark it as public when we couldn't decrypt instead of marking as Anon.

Wanted to add registration tests but might be easier to wait for when we support free addresses as the paid tests would make it 1000x more complex

Part of #1093

### DIFF
--- a/mutiny-core/src/blindauth.rs
+++ b/mutiny-core/src/blindauth.rs
@@ -407,9 +407,54 @@ fn create_blind_auth_secret(
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


TonyGiorgio (approved):

TonyGiorgio:
Needs rebase

PR #1159, opened 2024-04-25
Better convert fedimint errors

benthecarman:
Fedimint uses anyhow a bunch everywhere which is why we get some `Unknown Error`s. I added `MissingInvoiceAmount` which happens for zero amount invoices


mutiny-wasm/src/error.rs
```diff
@@ -230,11 +230,11 @@ impl From<MutinyError> for MutinyJsError {
             MutinyError::FederationConnectionFailed => MutinyJsError::FederationConnectionFailed,
             MutinyError::Other(e) => {
                 error!("Got unhandled error: {e}");
-                // FIXME: For some unknown reason, InsufficientBalance is being returned as `Other`
-                if e.to_string().starts_with("Insufficient balance") {
-                    MutinyJsError::InsufficientBalance
-                } else {
-                    MutinyJsError::UnknownError
+                // handle fedimint anyhow errors
```
TonyGiorgio, line 233: should be in mutiny core

TonyGiorgio (changes requested):

TonyGiorgio (approved):

PR #1164, opened 2024-04-30
Return better NWC error code for payment failures

benthecarman:


mutiny-core/src/nostr/nwc.rs
```diff
@@ -1038,10 +1038,16 @@ impl NostrWalletConnect {
                                     }
                                 }
 
+                                let code = if e == MutinyError::InsufficientBalance {
+                                    ErrorCode::InsufficientBalance
+                                } else {
+                                    ErrorCode::PaymentFailed
```
TonyGiorgio, line 1044: payment timeout shouldn't give a failure

TonyGiorgio (changes requested):

mutiny-core/src/nostr/nwc.rs
benthecarman: added

mutiny-core/src/nostr/nwc.rs
```diff
@@ -999,6 +999,11 @@ impl NostrWalletConnect {
                                             nostr_manager.logger,
                                             "Payment timeout, not removing payment from budget"
                                         );
+
+                                        nostr_manager.save_nwc_profile(self.clone())?;
+
+                                        // don't save to pending list, we already paid it
+                                        return Ok(None);
```
TonyGiorgio, line 1006: Payments that timeout, but then fail, will not be included in the list for them to retry. That's a huge miss.

TonyGiorgio (changes requested):

mutiny-core/src/nostr/nwc.rs
benthecarman: good catch

TonyGiorgio (approved):

PR #1161, opened 2024-04-28
Anchors

benthecarman:
todos:

- [ ] fix lnd crate to be usable
- [ ] fix cargo workspace stuff
- [ ] more test cases

PR #1166, opened 2024-05-06
Enable anchor channels and scid aliases

benthecarman:
pulled from #1161 

Added my second commit that makes it more usable in regtest and better to use externally


TonyGiorgio (approved):
is this ready?

benthecarman:
Yeah

PR #1170, opened 2024-05-07
Allow swapping between federations

TonyGiorgio:
- [x] Specify a from federation
- [x] To federation
- [x] Change the sweep fee API as well to account for from / to


benthecarman:
lgtm so far, something we could try to do is prefer a gateway if it is in both federations, would be a good follow up

TonyGiorgio:
> lgtm so far, something we could try to do is prefer a gateway if it is in both federations, would be a good follow up

I'm not dealing with all the complicated gateway switching logic.

futurepaul (approved):
working great for me. you can try it out here: https://github.com/MutinyWallet/mutiny-web/pull/1129

PR #1171, opened 2024-05-08
Add function to resync hermes dms

### LINKED ISSUE #1163: 3 payments lost
A strange thing happened today on my Mutiny address 3 payments that I sent correctly from another wallet of mine did not arrive.
What can I check?
Screenshot_20240430-175333
[mutiny-logs.txt](https://github.com/MutinyWallet/mutiny-web/files/15167687/mutiny-logs.txt)

benthecarman:
Closes #1163

### DIFF
--- a/mutiny-core/src/hermes.rs
+++ b/mutiny-core/src/hermes.rs
@@ -30,7 +30,7 @@ use crate::event::{HTLCStatus, MillisatAmount, PaymentInfo};
 use crate::federation::FedimintClient;
 use crate::labels::LabelStorage;
 use crate::nostr::RELAYS;
-use crate::storage::persist_payment_info;
+use crate::storage::{persist_payment_info, LAST_HERMES_SYNC_TIME_KEY};
 use crate::{
     blindauth::{BlindAuthClient, SignedToken},
     error::MutinyError,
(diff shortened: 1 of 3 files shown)


TonyGiorgio:
Should we just have this in the sync chain part? or another button in the admin page? 

I would also like to know why it's happening in the first place, this is a bandaid fix.

benthecarman:
> Should we just have this in the sync chain part? or another button in the admin page?

i think its better as a separate function here, for the front end we can do whatever

> I would also like to know why it's happening in the first place, this is a bandaid fix.

I am not 100% sure, my guess is that relays are giving us events ordered latest first, so if the app is closed before they are all done, we may miss some events. Not really sure how to best fix this however, there isn't a way to specificy a sort order in a nostr filter, only real fix would be to ask the hermes server

TonyGiorgio:
The "don't parse DMs before this date" logic is the problem here. We need to remove it.

benthecarman:
> The "don't parse DMs before this date" logic is the problem here. We need to remove it.

We need something, otherwise we have to do a full rescan everytime which slows down the wallet to a crawl.

Maybe saving the timestamp of when we last finished processing all events instead of individual events

TonyGiorgio:
> Maybe saving the timestamp of when we last finished processing all events instead of individual events

Sure, and then back date it 30 days so we know we aren't missing money

PR #1173, opened 2024-05-08
Disable mutiny plus profiles instead of delete

TonyGiorgio:
This partially solves this problem: https://github.com/MutinyWallet/mutiny-web/issues/1130

It at least lets you "disable" mutiny+ subscriptions. Nothing yet to make it enable though, but it at least doesn't charge people money that explicitly removed the old profile. 

Basically we need to disable, not full remove, because we ensure that a profile exists for subscribers.

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -2098,6 +2098,8 @@ impl<S: MutinyStorage> MutinyWallet<S> {
             self.ensure_mutiny_nwc_profile(subscription_client, autopay)
                 .await?;
 
+            // FIXME: switch the subscription from disabled to enabled if it was disabled
+
             self.check_blind_tokens();
 
             Ok(())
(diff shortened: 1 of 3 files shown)


mutiny-core/src/lib.rs
```diff
@@ -2117,47 +2119,39 @@ impl<S: MutinyStorage> MutinyWallet<S> {
...
-                if profile.tag != NwcProfileTag::Subscription {
-                    let mut nwc = profile.clone();
-                    nwc.tag = NwcProfileTag::Subscription;
-                    self.nostr.edit_nwc_profile(nwc)?;
-                }
+            if let Some(nwc) = nwc {
+                // only should have to submit the NWC if never created locally before
+                subscription_client.submit_nwc(nwc).await?;
             }
         }
```
benthecarman, line 2162: why delete this?

mutiny-core/src/lib.rs
```diff
@@ -2098,6 +2098,8 @@ impl<S: MutinyStorage> MutinyWallet<S> {
             self.ensure_mutiny_nwc_profile(subscription_client, autopay)
                 .await?;
 
+            // FIXME: switch the subscription from disabled to enabled if it was disabled
```
benthecarman, line 2101: need to do this

mutiny-core/src/nostr/mod.rs
```diff
@@ -1569,9 +1582,15 @@ impl<S: MutinyStorage, P: PrimalApi, C: NostrClient> NostrManager<S, P, C> {
     }
 
     pub fn delete_nwc_profile(&self, index: u32) -> Result<(), MutinyError> {
-        let mut vec = self.nwc.write().unwrap();
+        log_info!(self.logger, "Deleting nwc profile: {index}");
+
+        // TODO don't delete mutiny+ profile
```
benthecarman, line 1587: looks like this todo is done

benthecarman (commented):
good catch,

mutiny-core/src/lib.rs
TonyGiorgio: Because it's not needed. Why keep it?

mutiny-core/src/lib.rs
TonyGiorgio: We don't _need_ to right now.

TonyGiorgio: https://github.com/MutinyWallet/mutiny-node/issues/1174

mutiny-core/src/lib.rs
benthecarman: We use this in the front end to mark it as a subscription so we know to not display the nwc connection string

TonyGiorgio: It should already be marked correctly and that's not how the frontend handles it. https://github.com/MutinyWallet/mutiny-web/blob/a4981cb8cf0d8afe6f7e5d540fac0cc6b9c858d4/src/routes/settings/Connections.tsx#L110-L118

benthecarman: ah, its for the open in client buttons

https://github.com/MutinyWallet/mutiny-web/blob/a4981cb8cf0d8afe6f7e5d540fac0cc6b9c858d4/src/routes/settings/Connections.tsx#L157

benthecarman: > It should already be marked correctly

ah okay, this was for fixing old profiles, that's fine then

benthecarman (approved):

Issue #1183, opened 2024-05-12
Error parsing federation config


TonyGiorgio:
```
Error parsing meta config: error decoding response body: invalid type: boolean `false`, expected a string at line 530 column 1
```

benthecarman:
We should probably just return the raw json to the front end instead of trying to parse every value ourselves

TonyGiorgio:
So the frontend can run into the same exact issues parsing configs as we do? No thanks.

benthecarman:
Javascript won't fail to parse the whole thing if one type is wrong

TonyGiorgio:
We're very capable of using a serializer that wouldn't do that.

PR #1188, opened 2024-05-13
Remove parsing for unnecessary federation bool

TonyGiorgio:
Fixes https://github.com/MutinyWallet/mutiny-node/issues/1183


benthecarman (approved):
added another commit that removes more values that we don't use

PR #1202, opened 2024-06-06
Fixes for olympus swaps

benthecarman:
Previously when doing a swap we'd get an extra channel sometimes because we'd call the fee api multiple time which would cause the olympus node to open the extra channel. To make sure we only call the fee api once, we need to generate the invoice when estimating fee and then when we execute the swap, use that invoice, this way we can guarantee there is only a single call to the api for a given swap. 

For testing use this: `https://mutinynet-lsps1.lnolymp.us`

I tested:

- fedimint -> lightning w/ no channels
- fedimint -> lightning w/ channel and enough inbound
- fedimint -> lightning w/ channel and not enough inbound
- fedimint -> fedimint

### DIFF
--- a/mutiny-core/src/lib.rs
+++ b/mutiny-core/src/lib.rs
@@ -1669,12 +1669,16 @@ impl<S: MutinyStorage> MutinyWallet<S> {
         })
     }
 
-    pub async fn sweep_federation_balance(
+    pub async fn sweep_federation_balance_to_invoice(
         &self,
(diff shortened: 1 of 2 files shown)


TonyGiorgio:
Does this still work with the voltage LSP?

benthecarman:
> Does this still work with the voltage LSP?

yes

TonyGiorgio (approved):

PR #1203, opened 2024-06-06
Claim zaps in batches

benthecarman:
Helps with #1163

Before we'd claim each zap individually and this could cause problems because relays would send the events with the newest first, so if we stopped claiming the zaps for whatever reason (user closes app), then we'd miss the remaining zaps because we'd set our sync timestamp to the first one we claimed. This fixes this problem by instead accumulating all the zaps until we get the End Of Stored Events message from the relays, and then claiming the zaps, this way we know we got everything from the relay.

### DIFF
--- a/mutiny-core/src/hermes.rs
+++ b/mutiny-core/src/hermes.rs
@@ -14,10 +14,10 @@ use bitcoin::{bip32::ExtendedPrivKey, secp256k1::Secp256k1};
 use fedimint_core::config::FederationId;
 use futures::{pin_mut, select, FutureExt};
 use lightning::util::logger::Logger;
-use lightning::{log_error, log_info, log_warn};
+use lightning::{log_debug, log_error, log_info, log_warn};
 use lightning_invoice::Bolt11Invoice;
 use nostr::secp256k1::SecretKey;
-use nostr::{nips::nip04::decrypt, Event, JsonUtil, Keys, Tag, ToBech32};
+use nostr::{nips::nip04::decrypt, Event, JsonUtil, Keys, RelayMessage, Tag, ToBech32};
 use nostr::{prelude::decrypt_received_private_zap_message, EventBuilder};
 use nostr::{Filter, Kind, Timestamp};
 use nostr_sdk::{Client, RelayPoolNotification};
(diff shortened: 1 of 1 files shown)


TonyGiorgio:
When does the dm sync timestamp saved?

benthecarman:
After an individual ecash notification is handled:

https://github.com/MutinyWallet/mutiny-node/blob/ed3731a98e4374c5f395c8f7fbad310359b3e94e/mutiny-core/src/hermes.rs#L764

TonyGiorgio:
> After an individual ecash notification is handled:

I don't know what that means. When in the lifetime of an ecash redemption does this even take place?

I guess instead of asking, I should specify that if we're going to move the redemption of the ecash to a batcahable thing in the end, then the timestamp should only be saved once the batched ecash completes fully.

benthecarman:
I see what you're saying, moved it to set the timestamp after the batch, that should be safer

mutiny-core/src/hermes.rs
```diff
@@ -282,8 +290,19 @@ impl<S: MutinyStorage> HermesClient<S> {
...
+                                                        // otherwise we should wait until we have received an EOSE so we can do the initial batch
+                                                        if has_received_eose {
+                                                            if let Err(e) = handle_ecash_notification(notification, event.created_at, &federations, &storage, &claim_key, profile_key.as_ref(), &logger).await {
+                                                                log_error!(logger, "Error handling ecash notification: {e}");
+                                                            }
+                                                            // save the last sync time after processing the notification
+                                                            if let Err(e) = storage.set_dm_sync_time(event.created_at.as_u64(), true) {
+                                                                log_error!(logger, "Error saving last sync time: {e}");
+                                                            }
```
TonyGiorgio, line 302: Only set sync time if it succeeds.

mutiny-core/src/hermes.rs
```diff
@@ -295,7 +314,26 @@ impl<S: MutinyStorage> HermesClient<S> {
...
+                                            for (notification, created_at) in batch_notifications.drain(..) {
+                                                if let Err(e) = handle_ecash_notification(notification, created_at, &federations, &storage, &claim_key, profile_key.as_ref(), &logger).await {
+                                                    log_error!(logger, "Error handling ecash notification: {e}");
+                                                }
+                                            }
+
+                                            // save the last sync time after the batch
+                                            if let Err(e) = storage.set_dm_sync_time(max_created_at, true) {
+                                                log_error!(logger, "Error saving last sync time: {e}");
+                                            }
```
TonyGiorgio, line 333: Only set sync time if it succeeds.

TonyGiorgio (approved):

PR #1208, opened 2024-06-07
Only send auth token to Zeus

benthecarman:
Also has CORS issues with the Zeus LSP, told evan he is working on enabling the header


benthecarman:
added a test for the api, should let us catch this in the future

TonyGiorgio (approved):

PR #1210, opened 2024-06-10
ci: update GH actions

storopoli:
- updates all GH actions/* to non-deprecated/unmaintained/outdated up-to-date versions
- Also moving to dtolnay/rust-toolchain from actions-rs/toolchain since it is pretty much abandoned.

In my fork, all test passes and emits no CI warnings from deprecated stuff

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,15 +8,13 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
         with:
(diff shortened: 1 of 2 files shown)


storopoli:
@TonyGiorgio what am I missing? Is this PR rejected because the codebase is only supposed to be touched by `MutinyWallet` org members? Or are there technical issues with the contribution?

TonyGiorgio:
> Is this PR rejected because the codebase is only supposed to be touched by MutinyWallet org members

Contributions are welcomed but not random unknown CI changes from random first time contributors that aren't actually fixing anything. No offense.

PR #1217, opened 2024-06-10
Make sure LSP url is a valid url when switching

benthecarman:


TonyGiorgio:
still doesn't work

Screenshot 2024-06-10 at 8 00 35 PM

benthecarman:
Main purpose of this wasn't to be able to get people out of this state but to prevent it from happening. Added some `trim()`s everywhere that should help get your wallet unstuck

TonyGiorgio:
I used a new wallet and still got into this state.

benthecarman:
Think this should prevent everywhere, otherwise plz lmk how you reproduce

TonyGiorgio (approved):

PR #1221, opened 2024-06-14
Skip hermes check if we don't have a ln address

benthecarman:
Our sentry was getting filled up with logs of people who were getting a log because they don't have a lightning address.

image

### DIFF
--- a/mutiny-core/src/hermes.rs
+++ b/mutiny-core/src/hermes.rs
@@ -180,6 +180,11 @@ impl<S: MutinyStorage> HermesClient<S> {
                                 o.name
                             );
                             *c = (o.name.clone(), true);
+
+                            // if we don't have a lightning address, no need to continue
+                            if o.name.is_none() {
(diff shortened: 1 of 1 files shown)


mutiny-core/src/hermes.rs
```diff
@@ -180,6 +180,11 @@ impl<S: MutinyStorage> HermesClient<S> {
                                 o.name
                             );
                             *c = (o.name.clone(), true);
+
+                            // if we don't have a lightning address, no need to continue
+                            if o.name.is_none() {
+                                break;
+                            }
```
TonyGiorgio, line 187: Something seems strange to me about this code. I think this works, but would it make more sense to have `Ok(Some(u))` and `Ok(None)`? It would probably need to do the same `current_address_check_clone` stuff but make be more logical to seperate when the user has a name registered and not.

TonyGiorgio (approved):

PR #1223, opened 2024-06-17
Set min anchor channel fee to 1 sat/vbyte

benthecarman:
With us moving to the Zeus LSP that is lnd based, we could get force closes because lnd sets a max anchor channel fee, to make sure our min fee is never above lnd's max, we just set our min fee to 1 sat/byte to be safe.


TonyGiorgio (approved):
I will never understand all of these fee settings and why we keep having to continuously change them.

benthecarman:
Recent changes are just because anchor channels are new to us

PR #1224, opened 2024-06-17
Return channel open errors to user

### LINKED ISSUE #1220: Max channel size error messages don't show up
Zeus LSP has a max size of 1.5m sats but if the user tries to open a channel to the LSP for something bigger than that, the proper error message is not returned

benthecarman:
Closes #1220

Now exposes the channel open error message to the user

image


TonyGiorgio (approved):
awesome. i'm guessing this would work for channel amount too low errors too?

benthecarman:
Yeah I assume so

PR #1228, opened 2024-06-19
Set preimage for fedimint invoices

### LINKED ISSUE #1227: I don't think lookup_invoice is working
Tested out a new NWC app and it wasn't detecthing the invoice response. It just uses lookup_invoice and I believe we already support that. Not sure what's up.

benthecarman:
Fixes #1227 

Was talking to @gudnuf and he was getting responses, but the preimage was not set after it was paid. It seems like for fedimints we never set the preimage for invoices we create, first commit will make sure we save it to the invoice. Needed to add the second commit so we don't overwrite the preimage on the payment actually being completed, before we'd just recreate the MutinyInvoice instead of editing the one we had persisted. I believe the 2nd commit might help fix some issues where the same invoice would show up multiple times in the activity as well

### DIFF
--- a/mutiny-core/src/federation.rs
+++ b/mutiny-core/src/federation.rs
@@ -507,7 +507,7 @@ impl<S: MutinyStorage> FederationClient<S> {
 
         let desc = Description::new(String::new()).expect("empty string is valid");
         let gateway = self.gateway.read().await;
-        let (id, invoice, _) = lightning_module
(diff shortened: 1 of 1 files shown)


benthecarman:
this doesn't actually fix, looks like we overwrite when paid, still testing

TonyGiorgio (approved):

PR #1234, opened 2024-07-03
Add function to return the current LSP

benthecarman:
This is to fix where old wallets will display on the server screen that its using the olympus LSP even though internally it is using the voltage LSP.

Needed to make the lsp module pub to be able to use it in mutiny-wasm.

needs https://github.com/MutinyWallet/mutiny-web/pull/1253 for testing


futurepaul (approved):
fixed the problem for my mainnet wallet

PR #1235, opened 2024-07-03
Exponential backoff for some retries

benthecarman:
Does exponential back off for getting our hermes address and syncing nostr


TonyGiorgio:
Quitting @'ing me on mutiny node reviews.

futurepaul (approved):
looks simple enough to me

feels like this would blow up if count got really high but the timeout will be too long before that to ever happen

PR #1236, opened 2024-07-09
fix: `estimate_ln_fee` should be in sats

storopoli:
`estimate_ln_fee` should always return amount in sats not msats. This is not what happens when we get the amount from the Bolt11 invoice.


benthecarman:
Adding a test would be great, not sure how viable it is without a large refactor because it seems like it'd require a fedimint client

benthecarman (approved):

PR #1239, opened 2024-07-12
Fedimint resync

benthecarman:
test with https://github.com/MutinyWallet/mutiny-web/pull/1263


futurepaul:
It seems to me that the get progress fn is only returning to the frontend once with data, (it returns with 0 a few times before that): https://github.com/MutinyWallet/mutiny-web/pull/1263/commits/86e0dd0f78889f57d29e6f42c29026037e988078#diff-add9b9b8a7e6365d4aeb8afcd6eb4b7d6f7f2807d0d739266140818ff767d502R379

TonyGiorgio (approved):

